PR #53, opened 2025-01-29
Add pre-commit hook for rustfmt

stefanv:
@Theodlz we may also want to run `pre-commit run -a` (or `cargo fmt --all`) and then commit that, potentially with a `.git-blame-ignore-revs` file.


stefanv:
Merged manually.

Issue #55, opened 2025-01-30
Implement Dummy Kafka Producer


Theodlz:
Right now, the way we emulate reading from Kafka for testing is to have a fake consumer that instead just reads from file and ingests the data in Redis/Valkey.

Instead, it would be much more elegant if we stuck with a real kafka consumer, but faked the producer. That way we only have one consumer that is unchanged whether we are testing the app or running in production, and for testing one can only edit the config file to have it point to the local kafka instance that we are running with docker compose, which could be populated by a kafka producer script.

**To close the issue, we need to see a new binary that allows us to push alerts (downloaded from an archive) to a kafka topic hosted in the kafka instance spawned by docker-compose, so that the alert consumer binary can read alerts from it.**

PR #58, opened 2025-02-01
new Kafka producer + consumer limitation

Theodlz:
This PR addresses issue: https://github.com/boom-astro/boom/issues/55.

Here we:
- [x] add a kafka producer function and associated binary, which is able to grab ZTF alert packets from the public archive, to push them to the local kafka cluster that is running in docker.
- [x] edit the kafka consumer to actually read from kafka, wrap the previous's binary code in a function co-located with the producer's code, and then called in an associated kafka consumer binary.
- [x] fix some inconsistencies with the alert packet redis queue name, which for now is hardcoded - which is fine and can be addressed in another PR - written differently in multiple places in the codebase.
- [x] add/update some instructions in the README.md
- [x] Add optional limit to the number of alert packets pushed to redis by the kafka consumer, defaults to 1000.
- [x] add some tests, where applicable / useful
- [x] move to using clap for command line arguments in the kafka producer binary


jesaerys (commented):
@Theodlz still looking this over. Overall it makes sense, and I see that the build is green. I'm working on running it myself according to the updated README, and I want to spend a little more time tracing through the rust stuff, if that's ok.

src/bin/kafka_producer.rs
```diff
@@ -0,0 +1,34 @@
...
+            info!("Using date from argument: {}", arg);
+            date = arg.to_string();
+        } else {
+            error!("Invalid date format: {}", arg);
+            return Ok(());
+        }
+    }
+    if args.len() > 2 {
+        limit = args[2].parse::<i64>().unwrap_or(0);
+    }
```
jesaerys, line 29: We'll definitely want to move toward using `clap` for these kinds of things. I tried passing `-h` to this when I was playing around earlier and I was surprised when it didn't work.

src/kafka.rs
```diff
@@ -0,0 +1,216 @@
...
+    let client = redis::Client::open("redis://localhost:6379".to_string()).unwrap();
+    let mut con = client.get_multiplexed_async_connection().await.unwrap();
+    let queue_name = "ZTF_alerts_packets_queue";
+
+    let mut total = 0;
+
+    info!("Reading from kafka topic and pushing to the queue");
+    // start timer
+    let start = std::time::Instant::now();
+    // poll one message at a time
```
jesaerys, line 104: I feel like there must be a more elegant way to set this up, maybe using channels or something like that. We can explore this at some point.

src/bin/kafka_producer.rs
stefanv: See #65

jesaerys (approved):
I was able to run this locally and see the messages in kafka from the new producer, and everything else apparently ran the same as before. From a system test standpoint, it all seems to be working as advertised 👍 

My instinct is that `download_alerts_from_archive` and `produce_from_archive` belong somewhere with the tests. My only other comments are just some observations about the code, but nothing that should block merging.

jesaerys:
btw, I notice this isn't formally linked to #55. Do we want issues to automatically close when a linked PRs is merged, or do we want to do things manually?

Theodlz:
> I was able to run this locally and see the messages in kafka from the new producer, and everything else apparently ran the same as before. From a system test standpoint, it all seems to be working as advertised 👍
> 
> My instinct is that `download_alerts_from_archive` and `produce_from_archive` belong somewhere with the tests. My only other comments are just some observations about the code, but nothing that should block merging.

Though I would agree with that, it also feels weird to me that it would be in the tests folder but then called from something in the `src/bin` dir, no?

jesaerys:
Yeah you're right. I'll mull it over. It seems like there should be some way to organize things so that the producer is clearly a thing on the side used for testing and not part of the main functionality of boom. Unless of course there's a longer term plan for boom using a producer in production?

Theodlz:
> Yeah you're right. I'll mull it over. It seems like there should be some way to organize things so that the producer is clearly a thing on the side used for testing and not part of the main functionality of boom. Unless of course there's a longer term plan for boom using a producer in production?

Actually, there is! And not so long term. We are hoping to have the filter worker push the results to a kafka topic, which means that at some point some of this code will be generalized and reimported in other workers. That said, downloading from the ZTF archive for example, that's purely a testing thing.

Theodlz:
@stefanv @jesaerys if you have a chance to take a very quick look at the changes I just made before merging, that would be fantastic!

src/bin/kafka_producer.rs
```diff
@@ -11,24 +24,24 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     let mut date = chrono::Utc::now().format("%Y%m%d").to_string();
     let mut limit = 0;
-    if args.len() > 1 {
-        let arg = &args[1];
-        if arg.len() == 8 {
-            info!("Using date from argument: {}", arg);
-            date = arg.to_string();
+
+    if let Some(d) = args.date {
+        if d.len() == 8 {
```
jesaerys, line 32: Not critical for this PR, but I feel like there's a way to have clap parse the given date arg as a `chrono::DateTime`, and then this error case would be handled automatically by clap.

src/kafka.rs
```diff
@@ -65,12 +65,18 @@ pub async fn consume_alerts(
...
 ) -> Result<(), Box<dyn std::error::Error>> {
     let group_id = match group_id {
         Some(id) => id,
         None => uuid::Uuid::new_v4().to_string(),
     };
 
+    let queue_name = match queue_name {
+        Some(q) => q,
+        None => "ZTF_alerts_packets_queue".to_string(),
+    };
```
jesaerys, line 78: We should keep an eye on this kind of special casing with optional args. It's probably fine for now, at some point we may want to make `queue_name` required and call it with `"ZTF_alerts_packets_queue"` when we need to.

src/kafka.rs
```diff
@@ -159,16 +164,20 @@ pub async fn consume_alerts(
...
             error!("Error downloading alerts: {}", e);
-            return Ok(());
+            return Err(e);
         }
     };
 
-    let topic_name = format!("ztf_{}_programid1", &date);
+    let topic_name = match topic {
+        Some(t) => t,
+        None => format!("ztf_{}_programid1", &date),
```
jesaerys, line 179: Similar to my comment above about `queue_name`.

src/testing_util.rs
```diff
@@ -128,7 +128,7 @@ pub async fn insert_test_filter() {
       "fv": [
         {
           "fid": "v2e0fs",
-          "pipeline": "[{\"$project\": {\"cutoutScience\": 0, \"cutoutDifference\": 0, \"cutoutTemplate\": 0, \"publisher\": 0, \"schemavsn\": 0}}, {\"$lookup\": {\"from\": \"alerts_aux\", \"localField\": \"objectId\", \"foreignField\": \"_id\", \"as\": \"aux\"}}, {\"$project\": {\"objectId\": 1, \"candid\": 1, \"candidate\": 1, \"classifications\": 1, \"coordinates\": 1, \"prv_candidates\": {\"$arrayElemAt\": [\"$aux.prv_candidates\", 0]}, \"cross_matches\": {\"$arrayElemAt\": [\"$aux.cross_matches\", 0]}}}, {\"$match\": {\"candidate.drb\": {\"$gt\": 0.5}, \"candidate.ndethist\": {\"$gt\": 1.0}, \"candidate.magpsf\": {\"$lte\": 18.5}}}]",
+          "pipeline": "[{\"$match\": {\"candidate.drb\": {\"$gt\": 0.5}, \"candidate.ndethist\": {\"$gt\": 1.0}, \"candidate.magpsf\": {\"$lte\": 18.5}}}]",
```
jesaerys, line 131: This change isn't intuitive to me. Why should the test filter change in this PR?

jesaerys: Oops, I forgot to make it clear that this is my only comment I'd like to address before approval. My other comments above are more about raising awareness.

Issue #32, opened 2025-01-28
Filter Base Pipeline -  prv_candidates JD cut


Theodlz:
In `src/filter.rs`, we grab the user-defined filters which really are just MongoDB aggregation pipelines, and pre-prend them with the stages required to run their filter on specific alerts, as well as adding the data from the alert_aux catalog (prv_candidates and cross_matches).

At the moment we are limiting the prv_candidates data from going back further than 1 year, which is a good way to keep the filters from being too slow if they perform array operations. However, we should also limit the data going forward, to not have photometry that is more recent than the current alert.

While this likely won't be a problem if running it in real time, if we want to rerun filters on past data (which one day, we will want to offer as a feature), you would essentially "see in the future" when running on an alert that isn't the latest one. So, we just need a cut like `prv_candidate.jd <= current candidate.jd`

PR #59, opened 2025-02-02
Filter Base Pipeline - prv_candidates JD cut

Theodlz:
This PR edits the filtering function to limit the selected prv_candidates (lightcurve) datapoints to only the points up to the current alert being filters.

This is necessary so that if we want to run filters on past data, while running on a  alert for an object that has been detected again after that specific alert, we do not make the more recent datapoints available. That would be quite literally "cheating" as we would be looking into the future! A lot of filter have operations on lightcurves, and would yield incorrect/unexpected results if we were to run them again on older alerts for objects that have been detected again since these alerts, and that cut makes sure that the lightcurves doesn't go past their associated alert.

Addresses https://github.com/boom-astro/boom/issues/32


jesaerys (commented):
It seems like this should work. Is there an easy way to demonstrate that it does what we want?

jesaerys (approved):
In that case, looks good to me.

Issue #33, opened 2025-01-28
Alert Processing - append current alert to prv_candidates


Theodlz:
One issue we noticed in Kowalski is that aggregating the `prv_candidates` of alerts doesn't give you full lightcurves. Indeed, since the `prv_candidates` contains PAST detections up to but EXCLUDING the current alert, the current alert's data does not get added to the array.

So here are 2 things that happen:
- if we detect an alert for the last time, since we do not have a following alert that would include it, it never gets added to the lightcurve
- if we detect an object again but more than 30 days later, the previous datapoint isn't part of the 30-day rolling window of the `prv_candidates` and will be missing from the lightcurve

So, during the alert processing, we want to include the last datapoint (the alert itself) in the `prv_candidates` before inserting it in the database.

PR #60, opened 2025-02-02
Alert processing - add current alert to prv_candidates

Theodlz:
See https://github.com/boom-astro/boom/issues/33


jesaerys (approved):
Looks like it does exactly what it needs to based on #33. Error handling is already done, so I think this is good to go.

Theodlz:
Actually I noticed while looking at it that this wasn't quite right, only the candidate should be pushed to the array (not the whole alert). 

I'm writing a big big refactor to support LSST alerts and this is fixed there, so closing this PR

Issue #57, opened 2025-02-01
Implement named cmd line arguments when calling binaries


Theodlz:
Currently, our different binaries in `src/bin` that take cmd line arguments do so positionally. However that creates some issues. Say I have 3 arguments and the last 2 are optional, how do I only pass the third without it being read as the second? Also, it's not great from a UX perspective, named arguments avoid a lot of confusion.

**This issue is addressed once we have a better cmd line argument system, that can take a `--help` command as input, and can then take named arguments in the `--key=value` style or similar, with optional defaults.**

*Hint: I believe that a number of crates exist to make clis in rust. Maybe those - or more lightweight solutions - could help here.*

stefanv:
uv and pixi both use clap.

Theodlz:
Yep! I had this one in mind

PR #65, opened 2025-02-06
Cli args as keywords

stefanv:
A first attempt at addressing #57

### DIFF
--- a/src/bin/alert_worker.rs
+++ b/src/bin/alert_worker.rs
@@ -1,9 +1,18 @@
 use boom::{alert, conf, types::ztf_alert_schema, worker_util};
+use clap::Parser;
 use redis::AsyncCommands;
-use std::env;
 use std::sync::{Arc, Mutex};
 use tracing::{error, info, warn, Level};
 use tracing_subscriber::FmtSubscriber;
+
+#[derive(Parser)]
+struct Cli {
+    #[arg(required = true, help = "Name of stream to ingest")]
+    stream: Option<String>,
+
+    #[arg(long, value_name = "FILE", help = "Path to the configuration file")]
+    config: Option<String>,
+}
 
 #[tokio::main]
 async fn main() {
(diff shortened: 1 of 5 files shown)


stefanv:
Not sure what the correct protocol is wrt updating dependencies, so made that a separate commit.

src/bin/alert_worker.rs
```diff
@@ -1,10 +1,19 @@
...
 use redis::AsyncCommands;
-use std::env;
 use std::sync::{Arc, Mutex};
 use tracing::{error, info, warn, Level};
 use tracing_subscriber::FmtSubscriber;
 
+#[derive(Parser)]
+struct Cli {
+    #[arg(required = true, help = "Name of stream to ingest")]
+    stream: Option<String>,
```
Theodlz, line 11: making we should make the stream an enum? That way we can error out if the value provided isn't correct

src/bin/alert_worker.rs
```diff
@@ -1,10 +1,19 @@
...
+
+    #[arg(long, value_name = "FILE", help = "Path to the configuration file")]
```
Theodlz, line 13: what does `value_name` do?

src/bin/alert_worker.rs
```diff
@@ -13,23 +22,15 @@ async fn main() {
...
-        error!("Usage: alert_worker <stream_name> <config_file>, where config_file is optional");
-        return;
-    }
+    let args = Cli::parse();
 
     let interrupt_flag = Arc::new(Mutex::new(false));
     worker_util::sig_int_handler(Arc::clone(&interrupt_flag)).await;
 
-    let stream_name = &args[1];
+    let stream_name = &args.stream.unwrap();
```
Theodlz, line 30: maybe we should read those before creating that interrupt flag? That way if the values are incorrect we can catch that earlier

src/conf.rs
```diff
@@ -1,10 +1,18 @@
...
 use config::File;
+use std::path::Path;
+use tracing::error;
 
 use crate::types;
 
 pub fn load_config(filepath: &str) -> Result<Config, ConfigError> {
+    let path = Path::new(filepath);
+
+    if !path.exists() {
```
Theodlz, line 12: nice, thanks!

Theodlz (commented):
minor comments

src/bin/alert_worker.rs
stefanv: What would the values of that enum be?

src/bin/alert_worker.rs
stefanv: That's for when you type `--help`, so it shows `--config <FILE>`.

src/bin/alert_worker.rs
stefanv: I moved these up.

src/bin/alert_worker.rs
Theodlz: the instruments that we support? So for now just `ZTF`? Later will be ZTF + LSST + ...

src/bin/alert_worker.rs
Theodlz: got it, thanks!

Theodlz (approved):
LGTM, I think we can merge this as is!

PR #75, opened 2025-02-27
GeoJSON index on aux table + db utils

Theodlz:
This PR adds:
- a `coordinates.radec_geojson` field on the alert_aux entries to allow direct object-level queries. One issue we noticed in Kowalski is that folks might want to simply query for objects (often even just their names) without having to query for alerts (where you have multiple entries for the same object). This will also speed up the survey <-> survey cone searches we'll have to do for ZTF + Rubin operations, so we can query a table of N objects rather than a table of N * M alerts.
- We take advantage of this PR to properly add code to create the 2dsphere indexes we need to query the tables with cone searches, which was missing so far. For that we add a new `db.rs` file with a method to create indexes given a collection, and an index details' as a bson document

TODO: think about the error handling :) Here we do not care if creating an index fails, so the current error handling which simply logs errors should be fine already.

### DIFF
--- a/src/alert.rs
+++ b/src/alert.rs
@@ -75,6 +75,12 @@ pub async fn process_alert(
             "fp_hists": fp_hist_doc,
             "created_at": jd_timestamp,
             "updated_at": jd_timestamp,
+            "coordinates": {
+                "radec_geojson": {
+                    "type": "Point",
+                    "coordinates": [ra - 180.0, dec],
+                },
+            },
         };
         doc.insert(
             "cross_matches",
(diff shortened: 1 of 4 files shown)


Theodlz:
Now that the PR from Jake has been merged, I'll resolve conflicts on this one and will use the same type of error handling before asking for a review again

jesaerys:
Let me know if you need a hand with anything

Theodlz:
thanks a lot for your contribution to the error handling @jesaerys! Merging

PR #78, opened 2025-03-02
handle index creation errors

jesaerys:
@Theodlz let me know whether this is helpful.

It looks like `db::create_index` might be a general-purpose thing, so this is the kind of function where it's good to wrap the 3rd party error (`mongodb::error::Error`) in our own type, so I made a simple tuple struct named `CreateIndexError`. In `alert_worker`, there are five different places where `create_index` is called, so I _could_ have added five new variants to the `AlertWorkerError` enum corresponding to those failure cases. However, I'm not sure we really care whether it's one index vs another that failed to be created, so I'm simply wrapping the `db::CreateIndexError` in `AlertWorkerError::CreateIndexError`. We can always change this in the future if we need more granularity.

You'll notice I got rid of the call to `error!` in `db.rs`. The error event is still logged, but it happens automatically via tracing's `instrument` proc macro. We don't have to do it this way, I'm fine logging events the way you had it, but the idea is that we just decorate functions and methods with `instrument` and that makes a lot of this logging stuff a little easier. It's also really useful for collecting context along the way, e.g., the collection name. This is what I had in mind for the rest of the codebase, so I thought I'd see what you think here.

### DIFF
--- a/src/alert_worker.rs
+++ b/src/alert_worker.rs
@@ -1,7 +1,7 @@
 pub use crate::conf;
 use crate::{
     alert,
-    db::create_index,
+    db::{create_index, CreateIndexError},
     types::ztf_alert_schema,
     worker_util::{self, WorkerCmd},
 };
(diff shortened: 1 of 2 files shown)


Theodlz (approved):
Excellent! Thanks a lot @jesaerys, LGTM

PR #81, opened 2025-03-11
LSST Alerts: add additional data to alerts

Theodlz:
This PR adds magnitude data using the flux data found in the alerts, to the candidate, prv_candidates, and fp_hists. This includes:
- magpsf (magnitude of the psf flux)
- sigmapsf (magnitude error of the psf flux err)
- diffmaglim (upper limit at 5 sigma, like ZTF does)
- isdiffpos (boolean flag telling us if it's a positive or negative subtraction, like ZTF)
- snr (just added to the forced phot as it's already provided for the diaSource/candidate)

TODOs:
- [x] add diffmaglim to the non-detections array
- [x] add the non detection array to the database in the `process_alert` method
- [x] ~~validate some of the calculations. The upper limits seem to be brighter than the detections which shouldn't be possible, right?~~ -> this naturally happens when the SNR of the detection is less than 5, so the 5-sigma diffmaglim is brighter. Something to discuss.

Alternatively, their might be a way to do this directly when the avro is read by serde. If we can do that, that would be ideal. Not sure if it's worth it performance wise, but worth looking into before merging anything


src/alert/lsst.rs
```diff
@@ -462,6 +480,64 @@ pub struct LsstAlert {
...
 
+fn flux2mag(flux: f32, flux_err: f32) -> (f32, f32, f32) {
+    // convert from nJy to Jy
+    let flux = flux.abs() * 1e-9;
+    let flux_err = flux_err * 1e-9;
+
+    // convert from Jy to AB mag
+    let mag = -2.5 * (flux).log10() + 8.9;
+    let sigma = 1.0857362047581294 * (flux_err / flux);
+    let diffmaglim = -2.5 * (5.0 * flux_err).log10() + 8.9;
```
jesaerys, line 491: This looks like a duplication of logic with the `fluxerr2diffmaglim` function below. I would either replace this with a call to `fluxerr2diffmaglim` or (my preference) remove this line and having this function return just `(mag, sigma)` and call `fluxerr2diffmaglim` separately when you want a value for `diffmaglim`.

src/alert/lsst.rs
```diff
@@ -695,15 +774,36 @@ impl AlertWorker for LsstAlertWorker {
...
             .into_iter()
-            .map(|x| mongify(&x))
-            .collect::<Vec<_>>();
+            .map(|mut x| {
+                if let Err(e) = x.add_mag_data() {
+                    return Err(e);
+                }
+                Ok(mongify(&x))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
```
jesaerys, line 783: Nice use of `collect` to convert a vec of results into a result of a vec. It's a handy trick! We can also write this as,

```rust
            .map(|mut x| {
                x.add_mag_data()?;
                Ok(mongify(&x))
            })
            .collect::<Result<Vec<_>, AlertError>>()?;
```

And similarly for `fp_hist_doc` and `non_detections_doc`.

src/bin/kafka_consumer.rs
```diff
@@ -15,6 +15,8 @@ struct Cli {
     processes: Option<usize>,
     #[arg(help = "Clear the queue of alerts already consumed from Kafka and pushed to Redis")]
     clear: Option<bool>,
+    #[arg(help = "Set a maximum number of alerts to hold in redis, default is 15000")]
+    max_in_queue: Option<usize>,
```
jesaerys, line 19: I'm curious, did you run into an issue that inspired adding this, or is it just out of caution?

tests/data/alerts/lsst/0.avro
jesaerys, line 1: Is this fake data from some feed?

tests/test_lsst.rs
```diff
@@ -0,0 +1,79 @@
...
+    let bytes_content = std::fs::read(file_name).unwrap();
+    let mut alert = alert_worker
+        .alert_from_avro_bytes(&bytes_content)
+        .await
+        .unwrap();
+
+    assert_eq!(alert.candid, 25409136044802067);
+    assert_eq!(alert.candidate.object_id.unwrap(), 25401295582003262);
+
+    assert!((alert.candidate.ra - 149.802106).abs() < 1e-6);
```
jesaerys, line 19: This work, but if you're interested, there are various crates available that provide assertion macros to compare floats. Totally up to you.

jesaerys (commented):
I've been thinking about this more after our meeting on Tuesday, and I don't think serde can create new fields based on existing fields. Actually, this is for good reason: the point of serde is to convert between serialized and Rust representations of data, and adding arbitrary fields on the Rust side would break correspondence with the serialized data. This is effectively the approach you're taking with the new optional fields that are initially None and later mutated to Some. This certainly works, and I'm ok going forward with it, but I do want to throw out a suggestion...

I suggest leaving `DiaSource` as close to the actual schema as possible and introducing a new wrapper type with the derived fields. Something like this:

```rust
struct DiaSourceAugmented {
    dia_source: DiaSource,
    magpsf: f32,
    sigmapsf: f32,
    diffmaglim: f32,
    isdiffpos: bool,
}

impl TryFrom<DiaSource> for DiaSourceAugmented {
    type Error = AlertError;
    fn try_from(dia_source: DiaSource) -> Result<Self, Self::Error> {
        todo!()
    }
}
```

The advantage of this design is that it makes incorrect states impossible to represent in the program. E.g., logically it doesn't make sense to have a `DiaSource` with Some for `magpsf` and None for `sigmapsf`, yet the current code makes that a possibility. Let me know what you think.

src/alert/base.rs
```diff
@@ -54,6 +54,10 @@ pub enum AlertError {
     EmptyAlertError,
     #[error("failed to read avro")]
     AvroReadError(#[source] apache_avro::Error),
+    #[error("missing flux psf")]
+    MissingFluxPSF,
+    #[error("missing flux psf error")]
+    MissingFluxPSFError,
```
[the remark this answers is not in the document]
Theodlz: It's used here: https://github.com/boom-astro/boom/pull/81/files#diff-6b5ab6fa337436e2693f005c5ff7258c9e50ed926e6874596b79f8afc9bf2c81R507-R508

src/alert/lsst.rs
```diff
@@ -225,6 +225,14 @@ pub struct DiaSource {
     /// Template injection in the 3x3 region around the centroid.
     #[serde(rename = "pixelFlags_injected_templateCenter")]
     pub pixel_flags_injected_template_center: Option<bool>,
+
+    // these fields are computed later, from the psf_flux and science_flux (the latter is missing in the current LSST schema)
+    pub magpsf: Option<f32>,
+    pub sigmapsf: Option<f32>,
+    pub diffmaglim: Option<f32>,
+    pub isdiffpos: Option<bool>,
```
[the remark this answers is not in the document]
Theodlz: I might have just missed the `is_negative`, if we already have that then we don't need that extra field, but let's rename it so it's consistent with ZTF. I'll look at the schema description.

src/alert/lsst.rs
```diff
@@ -462,6 +480,64 @@ pub struct LsstAlert {
...
 }
 
+fn flux2mag(flux: f32, flux_err: f32) -> (f32, f32, f32) {
+    // convert from nJy to Jy
+    let flux = flux.abs() * 1e-9;
+    let flux_err = flux_err * 1e-9;
+
+    // convert from Jy to AB mag
+    let mag = -2.5 * (flux).log10() + 8.9;
+    let sigma = 1.0857362047581294 * (flux_err / flux);
```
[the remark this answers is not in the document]
Theodlz: yes of course, anyway this isn't python so it gets compiled and it doesn't matter. Sorry just a habit of precomputing things to avoid overhead :)

src/alert/lsst.rs
```diff
@@ -462,6 +480,64 @@ pub struct LsstAlert {
...
+        let (magpsf, sigmapsf, diffmaglim) = flux2mag(
+            self.psf_flux.ok_or(AlertError::MissingFluxPSF)?.abs(),
+            self.psf_flux_err.ok_or(AlertError::MissingFluxPSF)?,
+        );
+        self.magpsf = Some(magpsf);
+        self.sigmapsf = Some(sigmapsf);
+        self.diffmaglim = Some(diffmaglim);
+        self.isdiffpos = Some(self.psf_flux.unwrap() > 0.0);
+        self.snr = Some(self.psf_flux.unwrap() / self.psf_flux_err.unwrap());
+        Ok(())
```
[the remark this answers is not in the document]
Theodlz: Yes makes sense, let me change that

src/alert/lsst.rs
Theodlz: I like your suggestion better! Let me change that

src/bin/kafka_consumer.rs
Theodlz: Not really, just wanted to make it a parameter we can change, it was hardcoded before.

tests/data/alerts/lsst/0.avro
Theodlz: I extracted it from the kafka stream from LSST's real time test the other day. Given that it's simulated (hence fake) data and that their schema is public, this should be fine.

src/alert/lsst.rs
Theodlz: removed in the latest commit

src/alert/lsst.rs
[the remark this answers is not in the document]
Theodlz: done

src/alert/lsst.rs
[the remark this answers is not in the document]
Theodlz: ok I removed the is_negative (which definition is unclear based on the schema) in favor for our isdiffpos which is the ZTF standard too

Theodlz:
@jesaerys I addressed most if not all of your comments for your PR I believe. I will let you take another look. I think I might be able to deduplicate some code in the somewhat redundant custom deserializers, but this works very well and I end up with nicely formatted documents in Mongo thanks to the flatten serde attribute.

src/alert/lsst.rs
```diff
@@ -413,9 +435,25 @@ pub struct DiaNondetectionLimit {
...
-    // these fields are computed later, from the dia_noise
-    pub diffmaglim: Option<f32>,
+#[derive(Debug, PartialEq, Clone, serde::Deserialize, serde::Serialize)]
+pub struct NonDetection {
+    #[serde(flatten)]
+    pub dia_nondetection_limit: DiaNondetectionLimit,
+    pub diffmaglim: f32,
+}
+
+impl TryFrom<DiaNondetectionLimit> for NonDetection {
```
jesaerys, line 447: This one can just be `impl From` because the conversion never results in an error.

src/alert/lsst.rs
```diff
@@ -445,28 +483,62 @@ pub struct DiaForcedSource {
...
+        let (magpsf, sigmapsf) = flux2mag(flux.abs(), flux_err);
+        let diffmaglim = fluxerr2diffmaglim(flux_err);
+
+        Ok(ForcedPhot {
+            dia_forced_source,
+            magpsf: Some(magpsf),
+            sigmapsf: Some(sigmapsf),
+            diffmaglim: Some(diffmaglim),
+            isdiffpos: Some(flux > 0.0),
+            snr: Some(flux.abs() / flux_err),
```
jesaerys, line 520: Do you still want these fields to be `Option`s?

jesaerys (commented):
I think this looks great. Just a couple specific comments.

Is there ever a situation where we'd want to go backwards, i.e., _create_ serialized `DiaSource` etc. from an `LsstAlert`? If so, then we'll also want to add `serialize_with = ...` serde attributes to some of the fields in `LsstAlert`.

src/alert/lsst.rs
Theodlz: No you're right, these don't need to be options, I'll change that

Theodlz:
> I think this looks great. Just a couple specific comments.
> 
> Is there ever a situation where we'd want to go backwards, i.e., _create_ serialized `DiaSource` etc. from an `LsstAlert`? If so, then we'll also want to add `serialize_with = ...` serde attributes to some of the fields in `LsstAlert`.

Hmmmm I don't think so, we really just need to go from Avro -> rust struct -> mongodb document. The final mongodb document might look different enough (other microservices will add more data to it) from the initial struct, and we definitely never will need to go back to the original DiaSource. So, I'd say we don't need to worry about that.

src/alert/lsst.rs
Theodlz: done

src/alert/lsst.rs
Theodlz: done

jesaerys (approved):

PR #86, opened 2025-03-21
Additional conversions (magap, jd, ...) to LSST schema

Theodlz:
Here we:
- add the magnitude data for the aperture photometry
- convert all MJD fields from LSST to JD
- remove the provided SNR, replace it with the one we calculate from the reported flux and fluxerr (psf)
- remove some bad pixel flags that are redundant
- remove the "snaps" based photometry (whatever it might be lol)
- add the "3 closest objects" data to the `DiaObject` (will use it in the future).
- convert the ZTF `fid` (filter/band id) field to `band` as a string, similar to LSST
- ZTF's `ssnamenr`, avoid values such as "null" (as a str), making sure it is None/null when it should be.
- removed a number of ZTF alert fields (magdiff, classtar, magfromlim, aimagerat, bimagerat, sumrat, magapbig, sigmagapbig, jdendhist, tooflag, objectidps1, objectidps2, objectidps3, nmtchps, rfid, jdstartref, jdendref, nframesref, rbversion, nmatches, clrcoeff, clrcount, zpclrcount, clrmed, clrrms, neargaia, neargaiabright, maggaia, maggaiabright, drbversion) that are redundant, unused, or computed by us elsewhere.

### DIFF
--- a/src/alert/ztf.rs
+++ b/src/alert/ztf.rs
@@ -38,8 +38,6 @@ pub struct PrvCandidate {
     pub nid: Option<i32>,
     pub rcid: Option<i32>,
     pub field: Option<i32>,
-    pub xpos: Option<f32>,
-    pub ypos: Option<f32>,
     pub ra: Option<f64>,
     pub dec: Option<f64>,
     pub magpsf: Option<f32>,
--- a/src/alert/ztf.rs
+++ b/src/alert/ztf.rs
@@ -233,18 +233,8 @@ where
     D: Deserializer<'de>,
 {
     // if the value is null, "null", "", return None
-    let value: serde_json::Value = serde::Deserialize::deserialize(deserializer)?;
-    match value {
-        serde_json::Value::Null => Ok(None),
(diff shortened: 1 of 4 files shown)


src/alert/ztf.rs
```diff
@@ -254,6 +214,39 @@ where
...
+        serde_json::Value::Null => Ok(None),
+        serde_json::Value::String(s) => {
+            if s.is_empty() || s.eq_ignore_ascii_case("null") {
+                Ok(None)
+            } else {
+                Ok(Some(s))
+            }
+        }
+        _ => Ok(None),
+    }
```
jesaerys, line 247: I haven't tested it, but I offer this alternative:

```rust
    let value: Option<String> = serde::Deserialize::deserialize(deserializer)?;
    Ok(value.filter(|s| !s.is_empty() && !s.eq_ignore_ascii_case("null")))
```

Theodlz: added!

Issue #89, opened 2025-03-24
LSST & ZTF Forced photometry: flux -> mag, detections and non detections


Theodlz:
In the LSST case we already convert the flux data of the forced phot to mag, but do not account for the possibility that some will be detections, and some will be non detections. For ZTF, we don't convert the data to mag space yet.

We want to make fields like the magnitude and magnitude error optional in the LSST case when we have a non detections, and we need to implement the conversion logic for ZTF.

Theodlz:
I'm on it, a few things need discussing so I'll bring that up tomorrow

Theodlz:
Namely, well forced photometry just like the alerts comes with positive and negative detections (+- flux from the reference), and for transients we just care about the positive stuff (that's what we do in prod for ZTF in Kowalski), but maybe worth doing things well here and if there is a known obj with a star-like PSF in the reference we can have a proper absolute lightcurve, but then it isn't in the same reference frame as the alert so yeah not sure what to do for `isdiffpos` = false kind of objects.

PR #90, opened 2025-03-25
ZTF & LSST: Forced photometry flux-space to mag-space

Theodlz:
This PR addresses issue #89

### DIFF
--- a/src/alert/lsst.rs
+++ b/src/alert/lsst.rs
@@ -8,7 +8,7 @@ use crate::{
     alert::base::{AlertError, AlertWorker, AlertWorkerError, SchemaRegistryError},
     conf,
     utils::{
-        conversions::{flux2mag, fluxerr2diffmaglim},
+        conversions::{flux2mag, fluxerr2diffmaglim, SNT, ZP_AB},
         db::{cutout2bsonbinary, get_coordinates, mongify},
         spatial::xmatch,
     },
@@ -193,10 +193,10 @@ impl TryFrom<DiaSource> for Candidate {
             .ok_or(AlertError::MissingFluxAperture)?
             * 1e-9;
 
-        let (magpsf, sigmapsf) = flux2mag(psf_flux.abs(), psf_flux_err, 8.9);
-        let diffmaglim = fluxerr2diffmaglim(psf_flux_err, 8.9);
+        let (magpsf, sigmapsf) = flux2mag(psf_flux.abs(), psf_flux_err, ZP_AB);
+        let diffmaglim = fluxerr2diffmaglim(psf_flux_err, ZP_AB);
 
-        let (magap, sigmagap) = flux2mag(ap_flux.abs(), ap_flux_err, 8.9);
+        let (magap, sigmagap) = flux2mag(ap_flux.abs(), ap_flux_err, ZP_AB);
 
         Ok(Candidate {
             dia_source,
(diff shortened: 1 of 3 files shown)


Theodlz:
@jesaerys btw this one can be reviewed

src/alert/lsst.rs
```diff
@@ -466,35 +466,47 @@ pub struct DiaForcedSource {
...
+            Some(psf_flux) if (psf_flux / psf_flux_err) > 3.0 => {
+                let isdiffpos = psf_flux > 0.0;
+                let psf_flux = psf_flux * 1e-9;
+                let (magpsf, sigmapsf) = flux2mag(psf_flux, psf_flux_err, 8.9);
+                (
+                    Some(magpsf),
+                    Some(sigmapsf),
+                    Some(isdiffpos),
+                    Some(psf_flux / psf_flux_err),
+                )
```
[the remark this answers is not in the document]
Theodlz: Nice catch :)

src/alert/lsst.rs
```diff
@@ -466,35 +466,51 @@ pub struct DiaForcedSource {
...
             * 1e-9;
 
-        let (magpsf, sigmapsf) = flux2mag(psf_flux.abs(), psf_flux_err);
-        let diffmaglim = fluxerr2diffmaglim(psf_flux_err);
+        // for now, we only consider positive detections (flux positive) as detections
+        // may revisit this later
+        let (magpsf, sigmapsf, isdiffpos, snr) = match dia_forced_source.psf_flux {
+            Some(psf_flux) => {
+                let psf_flux = psf_flux * 1e-9;
+                if (psf_flux / psf_flux_err) > 3.0 {
```
jesaerys, line 489: This is false for `psf_flux < 0`, which I think is what you want, but I just want to make sure.

Also, I wonder if the 3.0 needs to be a `const`, just to help describe what this value is and to help prevent mistakes. This applies to ZTF, too. And I assume we want both surveys to use the same SNR threshold, so the `const` should be accessible to both modules.

src/alert/lsst.rs
```diff
@@ -193,10 +193,10 @@ impl TryFrom<DiaSource> for Candidate {
             .ok_or(AlertError::MissingFluxAperture)?
             * 1e-9;
 
-        let (magpsf, sigmapsf) = flux2mag(psf_flux.abs(), psf_flux_err);
-        let diffmaglim = fluxerr2diffmaglim(psf_flux_err);
+        let (magpsf, sigmapsf) = flux2mag(psf_flux.abs(), psf_flux_err, 8.9);
```
jesaerys, line 196: If the LSST zero point is always 8.9, I would make it a `const` this module.

src/alert/ztf.rs
```diff
@@ -106,6 +107,53 @@ pub struct FpHist {
...
+
+        let diffmaglim = fluxerr2diffmaglim(psf_flux_err, magzpsci);
+
+        Ok(ForcedPhot {
+            fp_hist,
+            magpsf,
+            sigmapsf,
+            diffmaglim,
+            isdiffpos,
+            snr: snr,
```
jesaerys, line 152: `snr: snr` can be just `snr`

jesaerys (approved):
Looks good. My only comments are to consider turning a couple of the special float values we use into `const`s. Those aren't blockers, though, so I'm approving.

src/alert/lsst.rs
Theodlz: yes that's what I wanted, for now let's only consider positive incursions as detections which is what we care about for transients. At a later date we might revisit this for stellar folks, but yeah just making an absolute value would make a negative incursion look like its brightening which is a problem.

Theodlz: and I like the idea of the 3 being a const! then we can have it somewhere at the top of the file (or in another enum-dedicated file) with a comment. Could call it SNT, for (Signal to Noise Threshold)

Theodlz:
@jesaerys I just addressed your comments

src/alert/lsst.rs
Theodlz: done

src/alert/ztf.rs
Theodlz: done

src/alert/lsst.rs
Theodlz: done

jesaerys (approved):
:shipit:

Issue #87, opened 2025-03-24
ZTF: split detections and non detections (like for LSST)?


Theodlz:
The LSST alerts do not provide one array of past detections & non detections, but instead provides 2 distinct arrays. The first one contains previous `diaSource` entries, an the latter provides non detections (visit id, time, noise).

For ZTF, we get one array with both, it makes the AVRO schema more complicated (which we can't do anything about), and to me it seems like all the filters that do make use of non detections + detections seem to split these in 2. So, I wonder if we should/could split those in 2 arrays, like LSST does. It might be a nice way to make the schema(s) from ZTF & LSST match even better.

Might be worth chatting with folks that have filters, to see what they think is more convenient.

jesaerys:
I'm not sure if this is possible with serde attributes. I suspect we'll need another intermediate struct to do this. And now I'm wondering if this might be where the deserialize-to-alert-in-one-shot-with-serde approach starts to fall apart...

Theodlz:
@jesaerys I think it's fine - and likely true - that we can't do everything in one shot in serde, but the more we can implement there the less likely we are to forget/erase some processing steps later on. Here it can be an exception for ZTF where we split the `prv_candidates` array in 2 based on if we have a valid magpsf value or not, when we are processing the alert packet.

jesaerys:
imo our best defense against forgetting or erasing important processing steps in the future is not to lean on serde, but to have clear descriptions of what our system is supposed to do through a combination of good type design and tests. If we can say that a deserialized avro alert from a survey should/will look like type X, and that we want to push documents serialized from type Y into the database, then that unlocks two things: first, it's super easy to test that our system can handle inputs as expected and produce correct outputs. Second, we're free to implement the conversion from X into Y however we want—a small function for simple cases, or maybe an entire module for more complex ones. If we treat X and Y as "anchors" for our system, and we have tests that hold us to that promise, then we're unlikely to accidentally remove important logic in something like a `impl From<X> for Y` or whatever. Using serde to go directly from avro to Y has been working so far, but I suspect it's going to start hurting soon.

The thing that makes this really tough for me is how to achieve a nice X and Y type design when we're working with schema objects that have SO MANY FIELDS. My feeling now is there's no way around just biting the bullet and explicitly writing out all of the fields, but maybe there's some way macros can help us. I don't know yet.

> then that unlocks two things

[...]

PR #91, opened 2025-03-25
 ZTF: split prv_candidates dets/non dets + various bugfixes and unit tests improvements

Theodlz:
Addresses https://github.com/boom-astro/boom/issues/87

In the same PR, I also added a few changes fixing issues I noticed while processing ZTF data:
- issues with the ML worker, looking for the images which we moved to another collection earlier
- issues with the ML worker, trying to push empty vecs to redis
- simplified the output of the ML worker to have one queue of (candid, programid) rather than having N queues per programid
- adapted the filter worker to use this simpler queue system, which prevents having to read from multiple queues. That was problematic as reading from a queue in batches doesn't let us some some kind of timeout so we just get struck reading one until data arrives, and can't read data from other queues so easily.
- use 15 partitions with the kafka producer docker image, to better mimick the setup of IPAC's kafka cluster for ZTF.
- add a dummy filter that is a bit more realistic and complex than the one we currently use for testing, in `data/filters`. This one can be used along with the new binary to add filters to the DB.
- improved the unit testing for the alerts. For both LSST and ZTF we test the process_alert method and verify that the resulting mongodb documents are correct.


Theodlz:
it's not finished (I want to add some proper error handling here and there, and in the new binary), but feel free to take a look if you'd like @jesaerys. Git says there are a lot of new lines but that's just the JSON file I added, I might re-upload it with lfs before merging this

Theodlz:
oh actually the current test just checks the deserialization which isn't impacted by this. I'm adding another test where we actually call the method to process the alert, then query it from the DB and check that a few key things are valid.

Theodlz:
@jesaerys I just added some unit tests that verify that we have what we expect in mongo. I'll open a ticket dedicated to this so we can expand on those tests and make them more robust.

jesaerys (approved):
I don't fully understand the filter side of the codebase yet (haven't spent much time there so far), but the changes look reasonable. The changes for #87 for sure look good.

Theodlz:
> I don't fully understand the filter side of the codebase yet (haven't spent much time there so far), but the changes look reasonable. The changes for #87 for sure look good.

I will open a PR about the filtering today or tomorrow, that might be a good time to talk about it over zoom and on github

PR #99, opened 2025-04-08
ML Worker - ZTF support with ACAI & BTSBOT

Theodlz:
This PR replaces the "fake ml worker" with an actual working one. For that we reorganized the codebase to follow the same pattern as the other workers:
- a `base` containing traits and methods that can take the survey specific implementations as parameters, and offer reusable methods
- survey specific implementations of said traits, here only for ZTF since we do not have models to run on LSST yet

We also have a `models` directory that has a `base` containing methods relevant for all models, and then a file per model we implement. For now we support the `ACAI` models and `BTSbot`, so we can be a drop-in replacement for Kowalski in the near future.


src/alert/ztf.rs
Theodlz, line 1: for context, here I re-added some fields that turned out to be used by btsbot. Anyways before we go to production we'll do a full review of what to keep vs what to throw

src/ml/models/base.rs
```diff
@@ -0,0 +1,48 @@
...
+};
+
+use crate::utils::fits::prepare_triplet;
+use mongodb::bson::Document;
+
+pub fn load_model(path: &str) -> Session {
+    let builder = Session::builder().unwrap();
+
+    builder
+        .with_execution_providers([CUDAExecutionProvider::default().build()])
```
Theodlz, line 14: I'll add Apple's CoreML here before merging

src/utils/fits.rs
```diff
@@ -1,5 +1,5 @@
-use flate2::read::GzDecoder;
-use std::io::Read;
+use zune_inflate::DeflateDecoder;
```
Theodlz, line 1: I should add a comment about this. It's basically an implementation of gzip/deflate that is faster than flate2. Plus it comes with a few handy options such as skipping checksum confirmation (ZTF should be sending us valid images anyways), and also we can provide it size hints to avoid it from reallocating memory as the decoded data grows larger (we can give it the max size in bytes we would get for a fits file if the image is at 63x63).

src/utils/fits.rs
```diff
@@ -85,8 +88,8 @@ pub fn buffer_to_image(buffer: &[u8]) -> Result<Vec<f32>, Box<dyn std::error::Er
     // so if NAXIS1 is not NAXIS_STANDARD, we need to add NAXIS_STANDARD - NAXIS1 zeros to the start and end of each row
     // and if NAXIS2 is not NAXIS_STANDARD, we need to add NAXIS_STANDARD - NAXIS2 zeros to the start and end of the vector
 
-    let offset1 = (NAXIS_STANDARD - naxis1) / 2; // Assuming naxis1 and naxis2 are both usize
-    let offset2 = (NAXIS_STANDARD - naxis2) / 2;
+    let offset1 = ((NAXIS_STANDARD - naxis1) as f32 / 2.0).ceil() as usize;
```
Theodlz, line 91: I had to change the code here, because what we had would behave incorrectly in some cases when the NAXIS wasn't an even number, or just the max size - 1. That deals with those errors by first conversion to f32 before doing the division, and then using ceil.

Theodlz:
Ok so I need to make some changes to remove the redis + receiver logic out of the individual workers and to the function that takes the worker type as input, just like we do for the alert worker, and also filter worker as of  #97. Will ask for a review once that and a few minor items are addressed.

src/ml/models/base.rs
Theodlz: done

src/utils/fits.rs
Theodlz: done

Theodlz:
> Ok so I need to make some changes to remove the redis + receiver logic out of the individual workers and to the function that takes the worker type as input, just like we do for the alert worker, and also filter worker as of #97. Will ask for a review once that and a few minor items are addressed.

All of this + proper error handling has been added. Unfortunately I do need #97 to be reviewed, edited, and merged to main before this one can be reviewed. I need some of the code in that other PR to be merged before this one can work end to end (the ML worker -> filter worker isn't working at the moment). That said, the code itself can be reviewed already. I will work on adding unit tests soon.

BTW adding the "coreml" to Cargo.toml creates some issues. I wonder if the Cargo.toml can be written so that some features are used based on the architecture we are compiling the app for/on.

Theodlz:
I am still working on unit tests, but please review this PR whenever you can @jesaerys. You'll see that over than (~800 lines) of the changed lines shown by github are just changes to the `Cargo.lock`, so it's actually not that big of a PR. A lot of moving things around to reorganize, and then the "real" new addition is the `ml/models` folder and it's content.

src/alert/base.rs
```diff
@@ -284,13 +284,15 @@ pub async fn run_alert_worker<T: AlertWorker>(
...
-                    warn!(error = %error, "Error processing alert, requeueing");
-                    con.lpush::<&str, Vec<u8>, isize>(&input_queue_name, avro_bytes.clone())
-                        .await
-                        .map_err(AlertWorkerError::PushAlertError)?;
-                    con.lrem::<&str, Vec<u8>, isize>(&temp_queue_name, 1, avro_bytes)
-                        .await
-                        .map_err(AlertWorkerError::RemoveAlertError)?;
+                    warn!(error = %error, "Error processing alert, skipping");
+                    // TODO: Handle alerts that we could not parse from avro
+                    // so we don't re-push them to the queue
```
jesaerys, line 289: I'm curious if you've seen this happen with real data.

jesaerys (approved):
lgtm

src/alert/base.rs
Theodlz: yep, a couple of failed aren't parse successfully. Could be some edge case not handled by the custom avro parsing logic we have for ZTF. Something for me to investigate, planning on doing so tomorrow actually.

jesaerys: That's really interesting. I'm eager to hear what you find out.

Issue #102, opened 2025-04-11
Test data reusability


Theodlz:
Our codebase including our tests is async. The issue is that for a number of tests we want to process avro records with the workers, and that means pushing that data to the DB, do things with it, and remove that data. Since these tests are async when 2 tests that use the same avro run, they will conflict and fail.

We want to have reusable test fixtures where the candid and objectId's are randomized. For that we could read the avro as a `Value` and edit it and write to avro again, or look for the bytes that match the candid and objectId we know are contained in the original avro and replace them with randomized values.

Theodlz:
@jesaerys I ran into that problem in my 2 PRs that are open right now and it's a blocker to add more tests. From our discussion earlier this week I think you were looking into it for your own issue you are working on, any updates?

I will likely work on this myself too as I need something up and running locally anyways.

Theodlz:
Ok I figured out something, I wanted to challenge myself a bit and see if I could do that on the raw avro data which worked so that's cool. I basically replace all occurences of the candid and objectId by new and randomized values so there is no overlap between tests. We can use the same approach to replace the ra/dec coordinates or other things. what we can't easily do is add/delete data but anyways if we need alerts with specific data in them, that's when we'll likely want to use some specific avro packets (of real objects for which we have predictable results) and just replace candid/objectId.

jesaerys:
Ok, this is good to know.

PR #103, opened 2025-04-11
Testing utils - Alert Randomizer

Theodlz:
This PR is a quick attempt at addressing #102. Here are some notes;
- avro uses variable length + zigzag encoding for its long data type (equivalent of i64 in rust). So unlike String values we can't just convert to bytes and look for those in the bytes of an avro file, we need to encode it first.
- for the same reason, when randomizing new candid values (and object id for lsst) the new values need to be in a similar range as the original values, as they would otherwise result in a different number of bytes and that creates an error when we try to replace existing bytes with the new ones.
- this could easily be solved by changing how we replace the existing bytes. Instead of just replacing them we could split the bytes input in 2 around the existing value, and append/prepend around the new one. However since in most cases I can think about we just want to randomize these and not set our own values, I don't think that would be a huge blocker. That said the ability to replace values with new ones will be essential if we want to edit things such as ra, dec, ... and other fields where there are reasons for us to create alerts with specific positions to test spatial queries.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@ data/alerts/ztf/20*.tar.gz
 __pycache__
 /env
 .DS_Store
+/experimental
 *~
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,7 +33,7 @@ dependencies = [
  "getrandom 0.2.15",
  "once_cell",
  "version_check",
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
@@ -141,7 +141,7 @@ dependencies = [
  "log",
  "num-bigint",
  "quad-rand",
- "rand",
+ "rand 0.8.5",
  "regex-lite",
  "serde",
  "serde_bytes",
@@ -299,6 +299,7 @@ dependencies = [
  "flate2",
  "futures",
  "mongodb",
+ "rand 0.9.0",
  "rdkafka",
  "redis",
  "reqwest",
@@ -326,7 +327,7 @@ dependencies = [
  "indexmap 2.7.1",
  "js-sys",
  "once_cell",
- "rand",
+ "rand 0.8.5",
  "serde",
  "serde_bytes",
  "serde_json",
@@ -1150,7 +1151,7 @@ dependencies = [
  "idna",
  "ipnet",
  "once_cell",
- "rand",
+ "rand 0.8.5",
  "thiserror 1.0.69",
  "tinyvec",
  "tokio",
@@ -1171,7 +1172,7 @@ dependencies = [
  "lru-cache",
  "once_cell",
  "parking_lot",
- "rand",
+ "rand 0.8.5",
  "resolv-conf",
  "smallvec",
  "thiserror 1.0.69",
@@ -1808,7 +1809,7 @@ dependencies = [
  "once_cell",
  "pbkdf2",
  "percent-encoding",
- "rand",
+ "rand 0.8.5",
  "rustc_version_runtime",
  "rustls 0.21.12",
  "rustls-pemfile 1.0.4",
@@ -2163,7 +2164,7 @@ version = "0.2.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77957b295656769bb8ad2b6a6b09d897d94f05c41b069aede1fcdaa675eaea04"
 dependencies = [
- "zerocopy",
+ "zerocopy 0.7.35",
 ]
 
 [[package]]
@@ -2218,8 +2219,19 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34af8d1a0e25924bc5b7c43c079c942339d8f0a8b57c39049bef581b46327404"
 dependencies = [
  "libc",
- "rand_chacha",
- "rand_core",
+ "rand_chacha 0.3.1",
+ "rand_core 0.6.4",
+]
+
+[[package]]
+name = "rand"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3779b94aeb87e8bd4e834cee3650289ee9e0d5677f976ecdb6d219e5f4f6cd94"
+dependencies = [
+ "rand_chacha 0.9.0",
+ "rand_core 0.9.3",
+ "zerocopy 0.8.24",
 ]
 
 [[package]]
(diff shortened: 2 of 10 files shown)


Theodlz:
@jesaerys perhaps you need something like this for your PR, + setting custom ra/dec as suggested in the description.

jesaerys:
Oh wow, I'll definitely take a look.

Theodlz:
@jesaerys I am taking a quick look at the float situation right now for ra, dec btw, will only spend 15 minutes or so on it so I'll commit whatever I have.

Theodlz:
ok now insensible to the original vs new length of the data in bytes, i'll take a quick look at the ra dec and I'll stop adding to this one

Theodlz:
I didn't commit everything but I think I've got the ra/dec figured out too, so I'll push that after a meeting

Theodlz:
ok @jesaerys I added support for randomizing or modifying ra/dec directly in the avro. I suggest looking at how I use that in the tests (there I use default() that randomized candid, objectId, ra, dec, but you can call new() and then set each value individually too).

jesaerys:
Wow, this is surprisingly complicated. It's really impressive that you've figured out how to encode things, though I'm curious if it's possible to get spurious matches when looking for `original_*_bytes` in the payload. Here's a contrived example: if ra and dec in the payload were the same value and you wanted to replace dec = 30 with dec = 40, then I think you'd end up with both ra and dec = 40 (though I could be misunderstanding the approach here). What I had in mind is to load our fixture file as an avro `Value`, munge _that_, and serialize back to avro bytes.

I think I'm now appreciating that we don't technically have pure avro data that can be easily serialized/deserialized with a schema, but rather avro bytes that have been decorated with magic bytes and whatever else. It might be nice to have a function to take us from raw bytes to avro bytes (for each survey, of course) as well as a function to go from avro bytes back to raw bytes suitable for testing `process_alert` and whatnot.

Theodlz:
@jesaerys I don't understand what you mean. there are no avro bytes vs raw bytes, it's just bytes. The reason why I took that approach (editing the value in the file itself) is simply because I wanted to give that a go, but I agree that it's not ideal. Ideally we could load it with the apache avro crate as a `Value`, edit it, and use a `writer` to re-convert it to an avro record (that's in bytes). But, I failed to see how I can edit the `Value` we get when reading the avro with the apache avro crate. If we can do that, then that's definitely the way to go. This PR should not be considered the only way to solve that problem, if there is a better way than we should do that.

When it comes to having ra=dec, or other additional occurences of a value: since we start from a known avro file and even use the original values (hardcoded) to replace with new values, well that input never changes and we know we don't run into those issues with those, so I don't think it's a real issue for our use case.

Theodlz:
I mean another approach is to change how we do things entirely, and serialize the avro to structs that match the schema exactly, and create new structs for those that match the structure and schema we want for further processing. However changing all of that just to randomize some values seems very overkill for me.

jesaerys:
re: ra=dec thing, point taken. It's not like we're dealing with unknown data, so my concern is a non-issue.

By "avro bytes" vs "raw bytes" I mean we can't just take an alert from the stream and hand it to the `apache_avro::from_avro_datum` to get a `Value` from it. Or maybe we can? I'm just reading through `alert_from_avro_bytes` (both ZTF and LSST) and part of the logic there is to skip past magic bytes so we can pass the remaining bytes to `from_avro_datum`. That's necessary because we can't just pass the entire array of bytes to `from_avro_datum`, right? If that's right, then it's like the bytes we get from the stream isn't strictly avro, but a blob with avro embedded.

Theodlz:
> > seems very overkill for me
> 
> That's funny because that to me that seems easier than the approach here :) I guess it comes down to different people having strengths in different areas. At the end of the day, I think this PR is really cool. It's just not how I would have approached it, but that's ok.

I mean it means changing quite literally how we do everything in the `src/alert` just to make fake data... I don't think that's justified just for unit tests convenience.

I am starting to figure out how to deal directly with the `Value`, so if I can figure that out we can do that instead. I'm able to read the avro, edit the `objectId` and save it back. The main issue is that the resulting avro file isn't really like the one from ztf's pipeline, and that is a big issue. We would run into the same problem even if we went the serde way so that would mean a lot of work for nothing.

Theodlz:
Ok I figured out how to do this with avro `Values`, not too hard in the end (was just badly documented as usual with rust if you ask me).

Theodlz:
Yeah ok I've got it fully figured out, not that hard at all. Will rewrite my PR fully in favor of using the avro crate which is cleaner and safer

Theodlz:
@jesaerys the real difficulty with the serde or `Value` way is LSST. The writer from that crate adds the schema and such before the data (like for ZTF) except well the LSST alerts don't need. LSST uses the schema registry and so alerts just have 5 bytes that include a magic byte (0) and the schema id so you can retrieve it from an API and cache it. Issue here is that we can - apparently - only create an avro that has that schema. Until we figure that out, the current approach from that PR is the only one that's viable.

Theodlz:
Ok I just got back to this, I found a simple solution. we find the start idx of the data (like for ZTF when reading and skipping the schema), only keep that and prepend it with the expected 5 bytes. About to push the new version of that PR.

Theodlz:
@jesaerys you can review again if you'd like. This PR desperately needs error handling, but other than that I think this new version is pretty good.

jesaerys:
The magic byte and schema ID thing are new to me, so I just spent some time looking into it. They aren't formally part of the avro spec, which only describes how to serialize data and schemas. The magic bytes and schema IDs are effectively an "extra" part that's tacked onto the actual avro-encoded blob. Even so, apparently this is fairly standard practice (as we see with ZTF and LSST). So, this is what I meant by "raw" vs "avro" bytes because I was naive. I learned something new today.

Theodlz:
Yes that's pretty much what it is. Although I think the best way to frame it isn't that the schema id is just slapped onto the avro data, but instead that it replaces the avro metadata (schema, codec,...) that usually prepends the actual data. So instead of having the schema in every file, you only have a pointer to the schema so you can download it once. That savings in terms of network usage, computing time, and above all energy usage are in my opinion justification enough for it to become the avro standard for single record packets.

Most feature complete avro packages in other languages support it, it looks like it's the rust one that is lagging behind, which is why I ended up implementing my own.

Theodlz:
@jesaerys anything else needed for this one to be merged?

Theodlz:
BTW I am asking because I would really like to wrap this one up first so I can make use of it in the ML and filtering PRs that are open, which would help clean things up especially for the filtering related one.

Theodlz:
> @jesaerys you can review again if you'd like. This PR desperately needs error handling, but other than that I think this new version is pretty good.

Oh and about that: Since this is only running for unique tests, I actually don't need nor want any fancy error handling. `unwrap`s are best in my opinion here, since they make the test fail (as expected) and provide pretty detailed error tracking.

jesaerys:
> Oh and about that: Since this is only running for unique tests, I actually don't need nor want any fancy error handling. `unwrap`s are best in my opinion here, since they make the test fail (as expected) and provide pretty detailed error tracking.

I completely agree. Failing a test because an unwrap panicked means something in the test didn't happen the way we expected, which is what tests are for.

Issue #101, opened 2025-04-11
ZTF Archive Kafka Producer - adding persistence


Theodlz:
<s>The `src/bin/kafka_producer.rs` we use to download a night of ZTF data from the archive and push it to a local kafka cluster is very slow. Technically we could skip that entirely and have a kafka consumer that can read the avro's from disk instead of kafka, but since we care about the performance when reading from kafka it does make sense to emulate that like we are doing.</s>

<s>That said, it's really slow. For some reason it takes a very long time to push alerts to the kafka cluster running in docker, and also that container doesn't persist the data we write to its topic. </s>

UPDATE: Performance has been addressed in #104. Now we still want to add persistence to the kafka docker image so that the topics that contain the alerts that come out of the user-defined (and later community defined) filters aren't just lost if the machine turns off or docker crashes.

Theodlz:
performance issues have been addressed, but leaving this issue open until persistence is added too

petebachant:
Seems to be working on my machine to enable a Docker volume for Kafka. Also related to @jesaerys' idea to add an option to the producer to skip producing if the topic already exists with the expected number of messages.

jesaerys:
> Seems to be working on my machine to enable a Docker volume for Kafka. Also related to [@jesaerys](https://github.com/jesaerys)' idea to add an option to the producer to skip producing if the topic already exists with the expected number of messages.

@petebachant wait, you figured it out already? Do you has code?

Theodlz:
I just assigned both of you on that issue. @petebachant  you already took care of the docker compose volume stuff (I assumed it worked fine right?). @jesaerys can you confirm that you are looking into the logic required for the producer to not produce again if the topic is populated already (with the right alert count, and otherwise deleted + recreates it)?

jesaerys:
@Theodlz yep, on it.

petebachant:
> I just assigned both of you on that issue. [@petebachant](https://github.com/petebachant) you already took care of the docker compose volume stuff (I assumed it worked fine right?). [@jesaerys](https://github.com/jesaerys) can you confirm that you are looking into the logic required for the producer to not produce again if the topic is populated already (with the right alert count, and otherwise deleted + recreates it)?

It's failing on my personal laptop. Need to check again on my Caltech machine. I assume this is something that will run once per volume and then fail therafter. Added some recommendations I found in #209.

PR #104, opened 2025-04-14
Kafka "produce_from_archive": performance improvements

Theodlz:
In this PR, we tweak some of the parameters of the `librdkafka` producer to improve the performance when pushing messages to a topic. Before these changes producing speeds would be ~around 5.5s for 1000 messages, now we're around ~250ms for the same volume, which is ~20x faster. In another PR we should consider making similar config changes to the producer used to push alerts that pass filters to kafka (introduces in PR #97), but due to message size and volume it is only critical for the `produce_from_archive` method at the moment.

Solves part of issue #101 (persistence is missing still, should be added in another PR).

BTW kafka optimizing isn't an exact science, it requires one to do a number of trials to finetune numbers. I am sure that speed could be even better, but this is already so much faster than it used to be and more than enough for this purpose. I was inspired by a number of threads online (https://medium.com/@charchitpatidar/optimizing-kafka-producer-for-high-throughput-f8e808b06bcb, https://stackoverflow.com/questions/66045267/kafka-setting-high-linger-ms-and-batch-size-not-helping) and the official `librdkafka` documentation (https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md).

### DIFF
--- a/src/kafka/ztf.rs
+++ b/src/kafka/ztf.rs
@@ -189,6 +189,15 @@ pub async fn produce_from_archive(
     let producer: FutureProducer = ClientConfig::new()
         .set("bootstrap.servers", "localhost:9092")
         .set("message.timeout.ms", "5000")
+        // it's best to increase batch.size if the cluster
+        // is running on another machine. Locally, lower means less
+        // latency, since we are not limited by network speed anyways
+        .set("batch.size", "16384")
+        .set("linger.ms", "5")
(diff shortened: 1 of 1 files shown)


jesaerys (approved):
> Before these changes producing speeds would be ~around 5.5s for 1000 messages, now we're around ~250ms for the same volume, which is ~20x faster.

Astounding. Nice work!

PR #107, opened 2025-04-17
Kafka producer bugfix: avro files only

Theodlz:
I noticed that if I had a non-avro file in the same dir as the alerts (say the original .tar.gz file from the ZTF archive), the producer tried to load it in memory which creates massive memory issues.

This PR simply makes sure that we skip non `.avro` files


jesaerys (approved):
I'm glad it wasn't the kafka producer tweaks that caused the memory issue :)

PR #108, opened 2025-04-17
Alert Randomizer: randomize only specified fields

Theodlz:
Currently we can use `.default()` or `.new()` on the alert randomizer classes. `default` randomizes the candid/objectId/ra/dec, whereas `new()` does not randomize any fields, and let's you set them to specific values.
However if one uses `new()` and only specified some fields but not all 4, the rest gets randomized. But, actually there are use cases where we do want to randomize some fields but keep the others at their original value.

In this PR, we change the behavior of the `.get()` method so if any of candid/objectId/ra/dec don't have specified values in the randomizer struct, then they stay at their original value. Also, we add `.rand_<field_name>()` methods so after calling `.new()` one can specify which fields should be randomized, and the rest can either be set to a new value or (by default) left to its original value.

For context this was needed as for the ML models the ra/dec can be part of the features used, in which case we don't want to randomize those as it would keep changing the scores.

### DIFF
--- a/src/utils/testing.rs
+++ b/src/utils/testing.rs
@@ -226,6 +226,10 @@ pub trait AlertRandomizerTrait {
     fn candid(self, candid: i64) -> Self;
     fn ra(self, ra: f64) -> Self;
     fn dec(self, dec: f64) -> Self;
+    fn rand_object_id(self) -> Self;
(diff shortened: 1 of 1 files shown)


jesaerys (approved):
I think this is a good way to do it.

Issue #114, opened 2025-04-22
ZTF alerts processing: aux ingestion concurrency issues


Theodlz:
If process A and process B are processing 2 alerts for the same objectId, which doesn't have an entry in the alert aux table yet, we might run into concurrency issue if both processes try to insert a new aux entry at the same time (as process A might do so a fraction of a second earlier than B, which yields an "duplication error" for B).

We can simply handle duplication errors on insert, in which case we try to update instead.

Issue #112, opened 2025-04-22
ZTF alerts: failed parsing


Theodlz:
Some alerts can't parsed for some reason, something to look into. It's a small fraction.

PR #115, opened 2025-04-23
ZTF processing: various bugfixes.

Theodlz:
This PR addresses issues #112  and #114.

For #112, the solution was that while some old avro spec stack overflow I read claimed that the block size detail was contained in a fixed 4 bytes, it was actually made of 2 variable length zigzag encoded long values (so avro's way of storing a simple i64 value, where it only uses the number of bytes required to encode the value => small value = less bytes):
- the first tells us how many records are in the next data block, which we can also double check is equal to 1 for our purpose.
- the second tells us the total length in bytes of the records in that block, useless for us.
So to fix that issue we replace the 4 bytes read by a proper decoding of these variable length values. We could quite frankly just read the 1 byte for the number of records but the performance impact is so minimal that I'd rather play it safe and read that data properly and not take shortcuts. Might revisit that bit later.

For #114 it is well described in the issue, but we can run in concurrency issues if 2 workers try to insert an entry in the aux table for the same object id. In which case we just need to handle mongo's duplication error and defer to an update. In that one I refactored the insert/update of the aux table into their own methods. That's just to keep the main logic clean, and when I add the same insert => update failsafe for LSST (in a future PR) I'll do the same and turn that into methods required by the AlertWorker trait.

PS: One alert that failed parsing was just genuinely problematic (as in the avro isn't quite correct). I tried to open it with an online avro reader (the excellent https://konbert.com/viewer/avro for those who want to check, it's a great website to check tabular data without writing any code) and it failed as well. At a later date I'll give a go at reading it with `fastavro` in Python, to see if somehow that library is able to read it.

### DIFF
--- a/src/alert/ztf.rs
+++ b/src/alert/ztf.rs
@@ -16,6 +16,43 @@ use serde_with::{serde_as, skip_serializing_none};
 use std::io::Read;
 use tracing::{error, trace};
 
+fn decode_variable<R: Read>(reader: &mut R) -> Result<u64, SchemaRegistryError> {
+    let mut i = 0u64;
+    let mut buf = [0u8; 1];
+
+    let mut j = 0;
+    loop {
+        if j > 9 {
+            return Err(SchemaRegistryError::IntegerOverflow);
+        }
+        reader
+            .read_exact(&mut buf[..])
+            .map_err(SchemaRegistryError::CursorError)?;
+
+        i |= (u64::from(buf[0] & 0x7F)) << (j * 7);
+        if (buf[0] >> 7) == 0 {
+            break;
+        } else {
+            j += 1;
+        }
+    }
+
+    Ok(i)
+}
+
+pub fn zag_i64<R: Read>(reader: &mut R) -> Result<i64, SchemaRegistryError> {
+    let z = decode_variable(reader)?;
+    if z & 0x1 == 0 {
+        Ok((z >> 1) as i64)
+    } else {
+        Ok(!(z >> 1) as i64)
+    }
+}
+
+fn decode_long<R: Read>(reader: &mut R) -> Result<i64, SchemaRegistryError> {
+    Ok(zag_i64(reader)?)
+}
+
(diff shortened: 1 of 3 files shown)


src/alert/ztf.rs
```diff
@@ -593,51 +709,42 @@ impl AlertWorker for ZtfAlertWorker {
...
-                "coordinates": {
-                    "radec_geojson": {
-                        "type": "Point",
-                        "coordinates": [ra - 180.0, dec],
-                    },
-                },
-            };
-            self.alert_aux_collection
-                .insert_one(alert_aux_doc)
+            match self
```
jesaerys, line 712: This is a good place for `if let Err(AlertError::AlertAuxExists) = self.insert_aux(...) {self.update_aux(...)}`.

jesaerys (approved):
I think this looks good, just one minor suggestion. I really appreciate the detail in the description.

src/alert/ztf.rs
Theodlz: but then what happens if it's a different error? Since I don't unwrap the other errors never yield right? That was why I used a match

jesaerys: Right, so you could do this:

```rust
            let result = self
                .insert_aux(
                    &object_id,
                    ra,
                    dec,
                    &prv_candidates_doc,
                    &prv_nondetections_doc,
                    &fp_hist_doc,
                    now,
                )
                .await;
            if let Err(AlertError::AlertAuxExists) = result {
                self.update_aux(
                    &object_id,
                    &prv_candidates_doc,
                    &prv_nondetections_doc,
                    &fp_hist_doc,
                    now,
                )
                .await?;
            } else {
                result?
            }
```

Using `if-let` saves a few lines and an explicit `return`. What you wrote is good, too.

jesaerys (approved):
LGTM. I don't really have any comments on the new changes.

src/alert/ztf.rs
Theodlz: Ah I see, that syntax makes sense yes. let me change that now

PR #116, opened 2025-04-25
Alert stream crossmatching

### LINKED ISSUE #83: Xmatch: new objects only vs every alert
We need to start xmatching ZTF alerts with the LSST_alert_aux collection. Unlike xmatches with static catalog which never change and therefore do not need to run more than once for a given position = object (based on object id), crossmatches between surveys need to happen for every single alert.

We could make that configurable using the existing xmatch config system, but basically need to split the xmatch catalogs in 2: static (only for new objects) and dynamic (for every alert). What we should really take advantage of is the fact that we know that LSST can see only up to dec = 33.5 deg. So, we should only bother crossmatchings ZTF alerts where their dec is less 33.50 deg. This will be a nice speedup to avoid querying the DB where we wont have any data anyways.

One thing to figure out is what we include in the projection (what fields from LSST_alerts_aux we keep). Ideally, we only want to keep the name so that the filters can do a lookup to that table and grab the lightcurves. So, perhaps it would be a bit more elegant/clean if we were to split the `cross_matches` (used for static crossmatches) from the `survey cross matches`. Perhaps we could just have an `aliases` dictionary, where we have the survey name as the key, and a list of objectids as the values.

jesaerys:
Closes #83 

`ZtfAlertWorker` has been updated so that whenever it processes a ZTF alert it checks for nearby alerts in the LSST aux collection.

Highlights:

* This survey crossmatching is essentially just a `nearSphere` geojson query where we take object_id of the nearest match, if there is one, and stash it in the ZTF aux table.
* The geojson query is supported by a new geospatial index on ra and dec.
* This PR also comes with some refactorings in the `AlertWorker` trait so that inserting into and updating the aux table are done via new `insert_aux` and `update_aux` methods. (Courtesy of @Theodlz in #117)
* A collection of test cases


jesaerys:
This is *almost* working. The new xmatch routine is putting something in the ztf aux collection, but it's not the expected ID in my unit test:
```
assertion `left == right` failed: matches: ["7035012124027342176"]
  left: ["7035012124027342176"]
 right: ["4354095285634829837"]
```
I'm trying to understand what's going on here.

jesaerys:
@Theodlz fyi

Theodlz:
thanks for the PR @jesaerys, I'll take a look at it tonight. I see that the tests are failing because its inserting an object (with a different ID) and the position is identical to another. It looks like you made your 2dsphere index unique which isn't needed in my opinion, so removing that would prevent the failure.

jesaerys:
Ok, @Theodlz this is finally ready to go.

jesaerys:
...for review, that is.

Theodlz:
@jesaerys thanks! Let me take a look tonight

src/alert/lsst.rs
```diff
@@ -885,6 +897,7 @@ impl AlertWorker for LsstAlertWorker {
                     &prv_candidates_doc,
                     &prv_nondetections_doc,
                     &fp_hist_doc,
+                    &Document::new(),
```
Theodlz, line 900: might just want to make it an option so you don't need to create a new document when you wouldn't use it anyways?

src/alert/base.rs
```diff
@@ -217,17 +223,19 @@ pub trait AlertWorker {
         object_id: impl Into<Self::ObjectId> + Send,
         ra: f64,
         dec: f64,
-        prv_candidates_doc: &Vec<mongodb::bson::Document>,
-        prv_nondetections_doc: &Vec<mongodb::bson::Document>,
-        fp_hist_doc: &Vec<mongodb::bson::Document>,
+        prv_candidates_doc: &Vec<Document>,
+        prv_nondetections_doc: &Vec<Document>,
+        fp_hist_doc: &Vec<Document>,
+        survey_matches: &Document,
```
Theodlz, line 229: might just want to make it an option if it's not relevant for all surveys, maybe better than having to create an empty document as I saw in `lsst.rs`?

Theodlz (approved):
LGTM 🚀 !

src/alert/lsst.rs
jesaerys: Yeah that occurred to me, I wanted to see how you reacted first. I wasn't sure if crossmatching between surveys will be normal/common enough that we'd always be providing `Some(doc)` in practice. If survey crossmatching truly is going to be an optional or extra thing, then my preference to make it `Option<Document>`.

jesaerys: And this is what I struggle with about nosql, everything is optional all of the time forcing me to think and make decisions :p

Theodlz: yeah I know hahaha

So really the situation is that the priority is ZTF being crossmatched with LSST, but long term we're hoping to do the opposite, if the performance is good enough that we can crossmatch 10M LSST alerts every night with ZTF, which is unclear with a MongoDB based implementation.

So let's just keep it as is for now.

jesaerys: Haha I just committed it to Option. We can change it later when it makes sense.

jesaerys:
Nope, all done. I'm waiting for the build to finish and then I'll merge

Theodlz:
> Nope, all done. I'm waiting for the build to finish and then I'll merge

I was writing the merge message oups, I had:

```
* Cross-match ZTF alerts with LSST objects, with a dec cutoff and 2 arcsec radius

* Create indexes on the aux tables when starting up alert workers

* added unit tests to validate the alert stream cross-matching for ZTF w/ LSST

Co-authored-by: Theodlz <theophile.dulaz@gmail.com>
```

jesaerys:
Is that what you want me to use?

PR #120, opened 2025-04-30
Remove unused ZTF avro schema

Theodlz:
We used to **require** the avro schemas for ZTF in the codebase in the early days of boom, to preload the schema to the alert worker. Now that we use a schema registry-like (or rather just caching the schema we load from the first alert we process) approach, we don't need that in the codebase. 

The only thing that uses a schema is `rsgen` but it's only relevant when one wants to add support for a new schema, so keeping an old - and already integrated - schema here is useless.

In this PR we:
- remove the `schema/ztf` directory and its content
- update the README's `rsgen` instructions with links to the repos that contain the up-to-date schemas for ZTF and LSST


jesaerys (approved):
Awesome, love me some housekeeping.

Issue #121, opened 2025-04-30
Redis connection: configurable


Theodlz:
Right now the connection to redis on localhost and on a specific port is hardcoded everywhere. Just like the connection to the database, that should be specified in the config file instead

PR #122, opened 2025-04-30
Redis connection: configurable

Theodlz:
In this PR, we make the redis connection configurable from the config file (defaulting to the usual localhost) just like the MongoDB connection already was, and replace all occurences of a hardcoded connection to redis with a call to that new function.

Addresses issue: https://github.com/boom-astro/boom/issues/121


jesaerys (approved):
Looks great

PR #130, opened 2025-05-27
Add API to workspace

petebachant:
Resolves https://github.com/boom-astro/boom-api/issues/10

Used [Cargo workspaces](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) to allow for some decoupling of dependencies between the different services. If in the future BOOM becomes somewhat of a shared library between the services it can be added as a dependency with:

```sh
cargo add --path . -p boom-api
```

### DIFF
--- a/api/src/api/alerts.rs
+++ b/api/src/api/alerts.rs
@@ -1,9 +1,9 @@
 use crate::models::response;
-use actix_web::{get, web, HttpResponse};
+use actix_web::{HttpResponse, get, web};
 use futures::TryStreamExt;
 use mongodb::{
-    bson::{doc, Document},
     Client, Collection,
+    bson::{Document, doc},
 };
 
 const DB_NAME: &str = "boom";
--- a/api/src/api/filters.rs
+++ b/api/src/api/filters.rs
@@ -1,8 +1,8 @@
 use crate::models::filter_models::*;
-use actix_web::{patch, post, web, HttpResponse};
+use actix_web::{HttpResponse, patch, post, web};
 use mongodb::{
-    bson::{doc, Document},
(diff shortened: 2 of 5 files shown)


Theodlz:
@petebachant looking into this tomorrow. May I ask why you decided to add an `api` folder in `src` that has its own lib and `main.rs`? Would it perhaps be easier to have the `main.rs` just turned into a bin in `src/bin/api.rs`? That way no need to run commands like the one in your PR description?

petebachant:
It might be helpful to put `api` under `crates`, following the pattern in [`uv`](https://github.com/astral-sh/uv), in case the codebase is modularized even further.

Theodlz:
> It might be helpful to put `api` inside a `crates` directory as `crates/boom-api`, following the pattern in [`uv`](https://github.com/astral-sh/uv), in case the codebase is modularized even further.

I like that idea. For the sake of other PRs Jake and I are working on let's go for your approach for now, and then we could have a folder for the api, a folder for the current boom code (we just need a name for it, could be `alert-ingestion` or similar), and a shared library of some kind so it's clear what is crate/program specific, vs shared between both?

Theodlz:
@petebachant I added Matthew Jenson who developed most of the API as a co-author, just to give a shout out to his work (which otherwise does not mention him if just merged to another repo like this!)

jesaerys:
Late to the party... if we're doing down the monorepo road, setting up a workspace to keep some distance between boom and the api is absolutely the correct design.

Theodlz:
Honestly the default logging Middleware that one can add to actix really isn't too bad at all, so we could just use that for now.

PR #128, opened 2025-05-26
Improve README installation instructions

anantgupta210:
This PR updates the README.md to fix issues I encountered during installation.
Changes include:
- Added missing dependencies for Linux
- Clarified usage of kafka_producer and kafka_consumer
- Noted optional requirement for requirements.txt

PR #133, opened 2025-05-28
Use inline dependency declaration in convert2onnx.py

petebachant:
Sort of related to https://github.com/boom-astro/boom/pull/128#pullrequestreview-2872908564. When run with `uv`, a script can declare its dependencies in line, saving the trouble of creating and maintaining a virtual environment.


petebachant:
> Well that depends of the script but I'd say from the script dir is best.

Alright, well I just made the script change working directories to the script dir so the user doesn't need to think about that.

PR #138, opened 2025-05-31
FilterWorker Alert output: add classifications

Theodlz:
This PR adds the classifications associated to an alert to the output of the filter worker, and associated unit tests.

It also adds the Alert struct to avro bytes conversion test to the LSST and ZTF-related unit tests, which so far wasn't tested anywhere.


petebachant (approved):
I don't know enough to be super helpful on this review, so LGTM!

PR #141, opened 2025-06-02
Add more endpoints to API

petebachant:
I also added a Makefile and target to hot reload for development.

PR #145, opened 2025-06-05
API: Basic logging

Theodlz:
While reviewing #141, I found it very useful to see logs of what I was trying through an API client. This PR adds some basic logging based on the official actix web example: https://actix.rs/docs/middleware/#logging

The formatting is simple but effective (ala nginx), as it records the endpoints being hit, the IP it came from, the time it took to respond, the response status, ...

PS: We might want to NOT use env_logger and somehow use tracing to make it look like the logging system of the rest of boom.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -527,6 +527,7 @@ dependencies = [
  "actix-rt",
  "actix-web",
  "config",
+ "env_logger",
  "futures",
  "mongodb",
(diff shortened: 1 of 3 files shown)


petebachant:
Going to merge this so I can fix the conflicts sooner rather than later. Thanks @Theodlz!

PR #157, opened 2025-06-10
WIP - API Authentication w/ JWT

Theodlz:
In this PR, we implement JWT-based authentication for the API. Here's how it works:
- When the API starts, we create an `AuthProvider` which defines some of the parameters of the JWT auth we are trying to set up (such as if we set expiration time for tokens and validate them, and also creates an admin user using credentials defined in the config for convenience), and adds it as `AppData` to actix web.
- During that setup stage, we also wrap all endpoints (but a newly added `auth` endpoint, described next) behind an `auth_middleware` which valides JWT tokens sent by the client as regular `Bearer Auth`.
- A client first queries the `/auth` endpoint, providing username/password. The password is compared to its hash saved in the database's user table, and if valid creates a new JWT token that is returned to the client. This way the password hashing and comparison against the database is only performed once (otherwise it can be a somewhat lengthy process! Hashing takes ~200 ms in release mode and you have the round trip to the DB. So using username/password as auth for every query is bad for performance and also not great for security, as the users share their password constantly over the internet).
- Thereafter, the client uses the JWT token to authenticate instead of username/password.

What I noticed while working on this is that there isn't any nice error handling yet in the API (which I probably just forgot about!), so I need to work a little bit on that before merging this first PR.

PS: By default, the config is set so that tokens do not expire at all. This makes it easier for development, and one can just set the expiration. Similarly in production one has to change the config's `secret_key` so the JWT auth remains secure.

### DIFF
--- a/api/src/auth.rs
+++ b/api/src/auth.rs
@@ -1,0 +1,145 @@
+use mongodb::bson::doc;
+
+use crate::{api::users::User, conf::AppConfig, conf::AuthConfig};
+
+use jsonwebtoken::{Algorithm, DecodingKey, EncodingKey, Header, Validation, decode, encode};
+use serde::{Deserialize, Serialize};
+
+/// Our claims struct, it needs to derive `Serialize` and/or `Deserialize`
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Claims {
+    sub: String,
+    iat: usize,
+    exp: usize,
+}
+
+#[derive(Clone)]
+pub struct AuthProvider {
+    encoding_key: EncodingKey,
+    decoding_key: DecodingKey,
+    validation: Validation,
+    users_collection: mongodb::Collection<User>,
+    token_expiration: usize,
+}
+
+impl AuthProvider {
+    pub async fn new(config: AuthConfig, db: &mongodb::Database) -> Self {
+        let encoding_key = EncodingKey::from_secret(config.secret_key.as_bytes());
+        let decoding_key = DecodingKey::from_secret(config.secret_key.as_bytes());
+        let mut validation = Validation::new(Algorithm::HS256);
+        validation.validate_exp = config.token_expiration > 0; // Set to true if tokens should expire
+
+        let users_collection: mongodb::Collection<User> = db.collection("users");
+
+        // always create the admin user if it doesn't exist
+        // using the admin_username from the config and admin_password
+        let admin_username = config.admin_username.clone();
+        // first check if the admin user already exists
+        match users_collection
+            .find_one(doc! { "username": &admin_username })
+            .await
+        {
+            Ok(Some(_)) => {
+                // Admin user already exists
+            }
+            Ok(None) => {
+                // Admin user does not exist, create it
+                let user_id = uuid::Uuid::new_v4().to_string();
+                let admin_password = config.admin_password.clone();
+                let hashed_admin_password = bcrypt::hash(&admin_password, bcrypt::DEFAULT_COST)
+                    .expect("failed to hash admin password");
+                let admin_user = User {
+                    id: user_id,
+                    username: admin_username.clone(),
+                    email: "<admin_email@example.com>".to_string(),
+                    password: hashed_admin_password,
+                };
+                // DEBUG print the username, password
+                println!("Creating admin user:");
+                println!("Username: {}", &admin_username);
+                println!("Password: {}", &admin_password);
+                match users_collection.insert_one(admin_user).await {
+                    Ok(_) => {
+                        println!("Admin user created successfully.");
+                    }
+                    Err(e) => {
+                        eprintln!("Failed to create admin user: {}", e);
+                    }
+                }
+            }
+            Err(e) => {
+                eprintln!("Failed to check for admin user: {}", e);
+            }
+        }
+
+        AuthProvider {
+            encoding_key,
+            decoding_key,
+            validation,
+            users_collection,
+            token_expiration: config.token_expiration,
+        }
+    }
+
+    pub async fn create_token(&self, user: &User) -> Result<String, jsonwebtoken::errors::Error> {
+        let iat = chrono::Utc::now().timestamp() as usize;
+        let exp = iat + self.token_expiration;
+        let claims = Claims {
+            sub: user.id.clone(),
+            iat,
+            exp,
+        };
+
+        encode(&Header::default(), &claims, &self.encoding_key)
+    }
+
+    pub async fn decode_token(&self, token: &str) -> Result<Claims, jsonwebtoken::errors::Error> {
+        decode::<Claims>(token, &self.decoding_key, &self.validation).map(|data| data.claims)
+    }
+
+    pub async fn validate_token(&self, token: &str) -> Result<String, jsonwebtoken::errors::Error> {
+        let claims = self.decode_token(token).await?;
+        Ok(claims.sub)
+    }
+
+    pub async fn authenticate_user(
+        &self,
+        username: &str,
+        password: &str,
+    ) -> Result<String, std::io::Error> {
+        let filter = mongodb::bson::doc! { "username": username };
+        let user = self.users_collection.find_one(filter).await.map_err(|e| {
+            std::io::Error::new(
+                std::io::ErrorKind::Other,
+                format!("Database query failed: {}", e),
+            )
+        })?;
+
+        // if the user exists and the password matches, create a token
+        // otherwise return an error
+        if let Some(user) = user {
+            match bcrypt::verify(&password, &user.password) {
+                Ok(true) => self.create_token(&user).await.map_err(|e| {
+                    std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        format!("Token creation failed: {}", e),
+                    )
+                }),
+                _ => Err(std::io::Error::new(
+                    std::io::ErrorKind::InvalidInput,
+                    "Invalid credentials",
+                )),
+            }
+        } else {
+            Err(std::io::Error::new(
+                std::io::ErrorKind::NotFound,
+                "User not found",
+            ))
+        }
+    }
+}
+
+pub async fn get_auth(db: &mongodb::Database) -> AuthProvider {
+    let config = AppConfig::from_default_path().auth;
+    AuthProvider::new(config, db).await
+}
--- a/config.default.yaml
+++ b/config.default.yaml
@@ -7,6 +7,11 @@ database:
   username: mongoadmin
   password: mongoadminsecret
   srv: false
+auth:
+  secret_key: "1234"
+  token_expiration: 0 # in seconds, 0 to disable
+  admin_username: admin
+  admin_password: adminsecret
 redis:
   host: localhost
   port: 6379
--- a/api/Cargo.toml
+++ b/api/Cargo.toml
@@ -13,6 +13,8 @@ mongodb = "3.1.0"
 serde = "1.0.215"
 serde_json = "1.0.138"
 env_logger = "0.11.8"
+jsonwebtoken = "9.3.1"
+chrono = "0.4.41"
 
 [dependencies.uuid]
 version = "1.16.0"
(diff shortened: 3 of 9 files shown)


Theodlz:
@petebachant I tagged you for review but just realized that some unit tests are failing, feel free to ignore this PR until it's green

api/src/routes/users.rs
```diff
@@ -33,15 +33,15 @@ pub struct UserInsert {
 )]
 #[post("/users")]
 pub async fn post_user(db: web::Data<Database>, body: web::Json<UserPost>) -> HttpResponse {
```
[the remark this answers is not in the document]
Theodlz: Yes, the claims contain the `sub` field which is the `user.id`. In my last commit we read it from the JWT, check that the user exists in the DB, and inject the user object in the request so handlers can use it. I - somewhat arbitrarily - edited the FilterHandler to show how the user.id can be retrieved and used. I think this is more than enough for this one PR, and we can think about `permissions` in a separate PR/ticket.

config.default.yaml
```diff
@@ -7,6 +7,12 @@ database:
   username: mongoadmin
   password: mongoadminsecret
   srv: false
+api:
+  auth:
+    secret_key: "1234" # Secret key for JWT authentication, change it to a strong secret
+    token_expiration: 0 # jwt token expiration in seconds, 0 to disable
+    admin_username: admin # Admin username for the API
```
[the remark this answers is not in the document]
Theodlz: good point, will add

api/src/auth.rs
```diff
@@ -0,0 +1,206 @@
...
+        let decoding_key = DecodingKey::from_secret(config.secret_key.as_bytes());
+        let mut validation = Validation::new(Algorithm::HS256);
+        validation.validate_exp = config.token_expiration > 0; // Set to true if tokens should expire
+
+        let users_collection: mongodb::Collection<User> = db.collection("users");
+
+        // always create the admin user if it doesn't exist,
+        // if it does exist, check that the password matches
+        let admin_username = config.admin_username.clone();
+        let admin_password = config.admin_password.clone();
```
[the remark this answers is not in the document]
Theodlz: I would love us to do a better job at handling secrets, but as you hinted I think we deserve a separate ticket for that.

api/src/routes/auth.rs
```diff
@@ -0,0 +1,24 @@
...
+    pub password: String,
+}
+
+#[post("/auth")]
+pub async fn post_auth(auth: web::Data<AuthProvider>, body: web::Json<AuthPost>) -> HttpResponse {
+    // Check if the user exists and the password matches
+    match auth.authenticate_user(&body.username, &body.password).await {
+        Ok(token) => HttpResponse::Ok().json(json!({
+            "status": "success",
+            "token": token,
```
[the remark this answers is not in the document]
Theodlz: done

api/tests/test_auth.rs
```diff
@@ -0,0 +1,127 @@
...
+        assert_eq!(resp.status(), StatusCode::OK);
+
+        let body = test::read_body(resp).await;
+        let body_str = String::from_utf8_lossy(&body);
+        let resp: serde_json::Value =
+            serde_json::from_str(&body_str).expect("failed to parse JSON");
+        assert_eq!(resp["status"], "success");
+
+        // the response is a jwt token
+        let token = resp["token"].as_str().expect("token should be a string");
```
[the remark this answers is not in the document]
Theodlz: done

Theodlz:
> Overall this looks good to me (once tests pass). A few comments and two general concerns:
> 
> 1. More secrets being added to the config, adding cognitive overhead for admins, i.e., they need to remember to set those as secrets and delete them from their config files before putting them in version control. We actually currently don't have a mechanism for patching these in from environmental variables. I could work on this in a separate PR if that would be helpful.
> 2. Extracting the user from the token in the endpoints, or some other way to restrict certain routes to admins, e.g., posting new users. We also probably want to restrict editing filters to the user that created them (or admins), so when a filter is posted, we want to put the user ID in the filter document. This will be easy if we can get the user ID or User object out of the token and into the function signature.

Hi Pete, I addressed most of your comments, and:
- I agree about the secrets stuff, a dedicated ticket + PR for handling secrets not just in the API but also the main app should definitely be done. I would rank that as fairly low in term of priority.
- In some of my more recent commits I get the `AuthProvider` to query for the user using the `user_id` for the JWT claims, and inject it into each request. That way endpoints that want to access the user can do so. Once we start looking into permissions we could have a simple method that takes a user + some list of permissions to enforce, and simply have that as some middleware or call it per endpoint.

api/src/routes/auth.rs
```diff
@@ -0,0 +1,27 @@
...
+use serde::Deserialize;
+use serde_json::json;
+
+#[derive(Deserialize, Clone)]
+pub struct AuthPost {
+    pub username: String,
+    pub password: String,
+}
+
+#[post("/auth")]
```
petebachant, line 13: Should probably add to utoipa.

api/src/routes/auth.rs
```diff
@@ -0,0 +1,27 @@
...
+    match auth
+        .create_token_for_user(&body.username, &body.password)
+        .await
+    {
+        Ok(token) => HttpResponse::Ok().json(json!({
+            "status": "success",
+            "access_token": token,
+            "message": "authentication successful"
+        })),
+        Err(e) => HttpResponse::Unauthorized().body(format!("authentication failed: {}", e)),
```
petebachant, line 25: Is it guaranteed that any error returned is due to an invalid username/password?

api/src/routes/auth.rs
```diff
@@ -0,0 +1,27 @@
...
+#[post("/auth")]
+pub async fn post_auth(auth: web::Data<AuthProvider>, body: web::Json<AuthPost>) -> HttpResponse {
+    // Check if the user exists and the password matches
+    match auth
+        .create_token_for_user(&body.username, &body.password)
+        .await
+    {
+        Ok(token) => HttpResponse::Ok().json(json!({
+            "status": "success",
+            "access_token": token,
```
petebachant, line 22: `token_type` is also required via OAuth2 spec:

```suggestion
            "access_token": token,
            "token_type": "Bearer",
```

api/src/routes/users.rs
petebachant, line 35: Along with this, I recommend when we initialize the DB with the first user, we set `is_admin = true` for them.

```suggestion
pub async fn post_user(
    db: web::Data<Database>,
    body: web::Json<UserPost>,
    admin_user: Option<web::ReqData<User>>
) -> HttpResponse {
if not admin_user.is_admin ... // Not authorized
```

api/src/db.rs
```diff
@@ -8,16 +11,90 @@ use mongodb::{Client, Database};
...
+        println!(
+            "Admin user does not exist, creating a new one with username: {}",
+            username
+        );
+        let admin_user = User {
+            id: uuid::Uuid::new_v4().to_string(),
+            username: username,
+            password: bcrypt::hash(&password, bcrypt::DEFAULT_COST)
+                .expect("failed to hash password"),
+            email: email,
```
petebachant, line 38:
```suggestion
            email: email,
            is_admin: true,
```

api/src/routes/filters.rs
```diff
@@ -283,7 +287,12 @@ pub struct FilterPost {
     )
 )]
 #[post("/filters")]
-pub async fn post_filter(db: web::Data<Database>, body: web::Json<FilterPost>) -> HttpResponse {
+pub async fn post_filter(
+    db: web::Data<Database>,
+    body: web::Json<FilterPost>,
+    user: Option<web::ReqData<User>>,
```
petebachant, line 293: Beautiful!

api/tests/test_auth.rs
```diff
@@ -0,0 +1,136 @@
...
+        let random_name = uuid::Uuid::new_v4().to_string();
+
+        let new_user = serde_json::json!({
+            "username": random_name,
+            "email":
+            format!("{}@example.com", random_name),
+            "password": "password123"
+        });
+
+        let req = test::TestRequest::post()
```
petebachant, line 36: Since we never delete this user, our local test databases are going to accumulate users each time we run the tests. Not the end of the world, but you could alternatively try to authenticate with the admin credentials in the config.

petebachant (commented):
A few more minor things. Happy to have it merged so long as we follow the @jesaerys's convention of adding non-addressed comments to the backlog for later discussion.

api/src/routes/auth.rs
Theodlz: Let me re-read through the method, I'm not sure when looking at that git diff

api/src/routes/users.rs
Theodlz: let me edit the user data model to include that

api/src/routes/filters.rs
Theodlz: yeah pretty happy with how that turned out, I'll rename stuff `current_user` or similar, just because `user` as is may conflict with other handlers where we have `user` variables

api/src/db.rs
Theodlz: done

api/src/routes/auth.rs
Theodlz: Ok so looking at `create_token_for_user` (which is the method that throws errors if any), it can be:

Theodlz: - Database query failed: ...
- Token creation failed
- Invalid credentials
- User not found

Theodlz: Does OAuth2 require us to return a specific error string?

Theodlz: here it does `authentication failed: <reason>`

petebachant: Not as far as I know.

Theodlz: Ok, in that case I think we can keep the current error schema then right?

petebachant: Yep, I think it's fine for now, but as we talked about in another PR, exposing internal details of the system like the full error messages to external users doesn't make much sense. It's not their concern that a database query or JWT encoding step failed; it's ours! Things like that should go to an observability system, and the user gets a plain 500 with no explanation when the server has an error.

api/src/routes/auth.rs
petebachant: Looks like this is still left to do.

config.default.yaml
```diff
@@ -7,6 +7,13 @@ database:
   username: mongoadmin
   password: mongoadminsecret
   srv: false
+api:
+  auth:
+    secret_key: "1234" # Secret key for JWT authentication, change it to a strong secret
+    token_expiration: 0 # jwt token expiration in seconds, 0 to disable
```
petebachant, line 13: May want to add a note here for anyone deploying a BOOM system that this is not recommended as the default for security concerns. Or just make the default 7 days or something, and in our own dev configs we can set to zero if desired.

petebachant (commented):
Couple minor things and we're good!

api/src/routes/auth.rs
Theodlz: Makes sense! Let me log it and always just return `Authentication failed` if it's a genuine failure, or `Incorrect credentials` otherwise (or similar).

Theodlz: Actually @petebachant, looking at: https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/, it looks like the error that's expected isn't just one error type, but a 400 or 401 (if its any error vs user/password incorrect/not found) with `error` and `error_description` fields. The description is optional, and "The error_description parameter can only include ASCII characters, and should be a sentence or two at most describing the circumstance of the error," "These are meant to give developers more information about the error, not intended to be shown to end users."

I guess I wonder how much detail we want to give, I think some minimal info isn't a bad idea perhaps, if that's the oauth2 recommendation. It should like they expect you to give some detail for whoever is using the API (considered here as the "developer").

Theodlz: Ok what I did is:
- edited the existing errors to remove any internal error detail and logged those instead, and made the error messages simple and clear (without revealing too much info about the underlying system, just enough to make sense of what might not be working).
- followed their recommended error json body format and error types

api/src/routes/auth.rs
Theodlz: added

petebachant (approved):
LGTM! 🚀

PR #158, opened 2025-06-10
Add Dockerfile

petebachant:
This is set up to be able to build a different image for each executable, e.g., with `docker build --target api -t boom-api .`. I've never done this before, but it looks pretty convenient since Docker Compose can set the `target` as a build arg:

```yaml
services:
  api:
    build:
      context: .
      dockerfile: Dockerfile
      target: api
    image: boom/boom-api:latest
    ports:
      - "4000:4000"
  scheduler:
    build:
      context: .
      dockerfile: Dockerfile
      target: scheduler
    image: boom/boom-scheduler:latest
 ```

The ultimate goal is to be able to run the entire app (all services) with `docker compose up`, merging in a separate dev Docker Compose config for local development, which would ideally use `cargo-watch` and mount the local working directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,10 +5,6 @@ WORKDIR /app
 # Copy the Cargo.toml and Cargo.lock files
 COPY Cargo.toml Cargo.lock ./
 
-# Build dependencies (this will be cached if Cargo.toml and Cargo.lock don't change)
-# TODO: This doesn't work without a lib or bin section in Cargo.toml
-# RUN cargo build --release --target x86_64-unknown-linux-gnu
-
 RUN apt-get update && \
     apt-get install -y curl gcc g++ libhdf5-dev perl make libsasl2-dev && \
     apt-get autoremove && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,12 +25,6 @@ WORKDIR /app
 # Copy the built executable from the builder stage
 COPY --from=builder /app/target/release/boom-api /app/boom-api
 
-# Copy the config file
-# TODO: This should probably be set some other way, e.g., all configuration
-# should be done via environmental variables set by Docker Compose or
-# Kubernetes
-COPY config.yaml /usr/local/etc/config.yaml
-
 # Expose the port your application listens on
 EXPOSE 4000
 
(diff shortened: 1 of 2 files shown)


petebachant:
Note it took a bit of fiddling with the system dependencies to get the build to work, but it's all good now.

Dockerfile
jesaerys, line 1: I think we'd at least want a stage for `kafka_consumer` as well.

petebachant:
> Can you explain a little more about what you have in mind? Bigger picture, I agree that it makes sense to have an "everything up" capability for deployment, I also want to make sure we land on something that's simple and flexible for local development, too. I get a little nervous when we start talking about mounting directories for local development, not that we shouldn't support that as an option, but I don't think we need to force people to work that way.

My goal is to be able to spin up something that resembles a production instance with a single command to observe and test changes against it, run benchmarks, etc. Maybe this one command would also ensure there is some historical data in the database, at least one admin user, etc. The scope of this PR is way smaller though--this Dockerfile just give us something that will successfully build.

Dockerfile
petebachant: I was also considering putting all binaries into a single image, and just using a different command for each container in the Docker Compose config. Would be a little simpler, and it seems like this will likely be deployed on a single node, so having one image that's slightly bigger than 3 other images is probably better.

Theodlz: We need to discuss it because the LSST and Kafka consuming is very different, since ZTF has one topic per night but LSST just has one giant topic where we read from the right offset that corresponds to the provided date and read everything after that.

jesaerys: That's effectively what Pete has in place now. The final scheduler stage is debian slim with the binary copied in (and other binaries can be added, too).

Theodlz (approved):
LGTM. Won't hurt to have a docker image even if we want to change it later, so I'm happy to add this on main. Pete, maybe you can already add an `api` and `scheduler` section to the default docker-compose but just comment those out? That would be nice so we can quickly run these along with other services. I'm hoping to test boom in a prod like setting but I only will have docker access on the test machine so this is perfect. A way to run the producer and consumer would be nice too, maybe just one container with both bins could do the trick for now!

Theodlz:
@petebachant we can just have a `docker.yaml` config which is a version of the config that uses the correct docker container names and not localhost, and you just copy/paste that into the docker images at build time as config.yaml, or as we do for a lot of application in productions if you don't need the right config to build the image itself then mounting the config file as a volume and then giving that path to the bins is an even better approach.

Issue #151, opened 2025-06-10
Throughput monitoring


jesaerys:
It would be nice to have some way to consistently measure the throughput of boom, to see how we're doing against the projected 10M alerts per night from LSST, plus any SLAs we put in place for how quickly we pump through each night of data.

We currently print an INFO message on shutdown that shows the total number of alerts processed, the total time elapsed, and the average rate (alerts per sec). This will probably be fine until we're running in prod, at which point we'll want to track the alert rate as a _**"metric"**_ (shudder) and monitor in grafana or whatever. Being able to track throughput over time will be really nice, especially as a way to watch for obvious step changes (good or bad) after a deployment, or to see when we start running into database limitations, and so on.

petebachant:
A Prometheus metric (sorry Jake) measuring the size of the Valkey queues over time could be helpful here too. You could see it fill up and how fast it drains.

jesaerys:
Some updates:

1. I have a very rudimentary demo app working that prints OpenTelemetry metrics to the console. The situation is a little like the `tracing` ecosystem in that multiple crates are required. So far my demo app involves three separate crates (all official from OpenTelemetry): `opentelemetry`, `opentelemetry_sdk`, and `opentelemetry-stdout`. It took a bit of reading to understand each of their roles and how they work together.

2. To get alert throughput, it seems that all we need is a simple counter. We can then turn the count data into a rate timeseries when we query the metrics backend. OpenTelemetry defines a handful of different kinds of metrics, and initially it wasn't immediately clear to me which one we should use for throughput monitoring. After reading through the spec as well as some crate API docs, it seems like a basic "sync" counter is all we need:

   * A counter rather than a gauge because gauges record a specific value at a moment in time, which isn't what we want. On the other hand, counters are incremented, so when we process an alert we can just increment the counter by 1.

   * A "sync" counter as opposed to an "async"/"observable" counter because the latter is used when the increment value needs to be obtained from a callback. We don't need that. We just need to add 1 after we process an alert, and that's what the "sync" counter is for.

   * Integer overflow won't be an issue. The counter is based on `u64`, which maxes out at ~1e19. If we process ~1e7 alerts per night from LSST, then we can run for ~1e12 nights or ~1e9 years without overflowing the counter.

[...]

PR #165, opened 2025-06-16
Throughput testing script

### LINKED ISSUE #156: Assess the impact of tracing on throughput
I'm super curious about how instrumenting with the `tracing` crate (including all of the spans introduced in #136) affects performance. It would be great to run some tests on this. Assuming we have a reliable method for measuring throughput on a given machine, maybe the easiest way to compare is to run boom once as-is and the run it again from a build with one of the static filtering features enabled ("max_level_error" or "max_level_warn", or maybe even "max_level_off"). Static filtering causes events and spans at higher levels to be skipped by the compiler so that they're completely absent from the binary (see https://docs.rs/tracing/latest/tracing/level_filters/index.html#compile-time-filters).

It may also be revealing to compare flamegraphs to see how much time is consumed by tracing.

If there is a significant cost to tracing, then we'll have to find ways to cut down on spans and events. If there's still an issue, it will be worth looking into the event recording/filtering process to see if there are any opportunities to optimize how we use `tracing` (breadcrumbs: read up on `Subscriber::{register_callsite,event_enabled,enabled,record}`, see https://docs.rs/tracing-core/latest/tracing_core/callsite/, and also https://docs.rs/tracing-subscriber/0.3.19/tracing_subscriber/layer/index.html#global-filtering). We can also possibly look into using an alternative crate, e.g., https://crates.io/crates/fastrace (I haven't used it).

jesaerys:
Closes #156. See the new report in `docs/tracing-throughput-test/tracing-throughput-test.md` for details.

This PR introduces a utility script that measures how fast boom processes a given night of alerts. See the Discussion section in the script's help message for details. This script is intended to be used by devs on any machine that's set up to run boom following the usual steps in our readme. The goal for #156 is to use this to see how throughput changes with different `tracing` compile-time filters and `RUST_LOG` settings.

I used the testing script to determine that `tracing` doesn't seem to have a measurable impact on our system. I wrote about it in `docs/tracing-throughput-test/tracing-throughput-test.md`.

I've hit my bash quota for at least a year, I hate it so much. I always forget how tricky it is to get things right in bash... maybe we can discuss sometime whether we're ok as a team making `uv` or whatever part of our dev toolbox.


scripts/throughput_test.sh
```diff
@@ -0,0 +1,409 @@
...
+}
+
+help() {
+  usage
+  echo
+  echo "Runs a throughput test for the alert processing pipeline."
+  echo
+  echo "Requires kafka, valkey, and mongodb to be running via docker compose."
+  echo
+  echo "**Important:** the following will be dropped and overwritten:"
```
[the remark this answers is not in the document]
jesaerys: Yeah, this is *not* to be run in production! It's a tool for developers, and we'll hopefully get a better method for measuring throughput (#151) making this script obsolete.

scripts/throughput_test.sh
```diff
@@ -0,0 +1,409 @@
...
+      echo "${count},${elapsed}"
+      break
+    fi
+
+    sleep "${check_interval}"
+  done
+}
+
+main() {
+  # TODO: number of iterations?
```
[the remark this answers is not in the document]
jesaerys: Running `docker compose up -d` crossed my mind at first, but I decided not because I'm not sure what it buys us plus the added overhead would really add up when running this a bunch of times in a row. The approach I've taken here is to assume the required containers are running and then manipulate them as I would manually to get a clean run each time (dropping the db before, deleting the topic).

I'm open to other approaches, if you have ideas.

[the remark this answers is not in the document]
jesaerys: Ah! Ok. Let's wait to see what we actually end up doing with this and the numbers it produces. This script isn't meant to be a formal benchmark of throughput. It really is focused on addressing #156, which is about getting a sense of the cost of instrumenting our code.

petebachant:
FYI, I have a stateless containerized (but non-parameterized) throughput workflow going here: https://github.com/boom-astro/boom-paper/blob/main/config/boom/compose.yaml

One important piece that's missing is the ability to shut down the containers as soon as all of the worker queues are empty. Is there anything in this script you could point me to for doing that?

jesaerys:
> Is there anything in this script you could point me to for doing that?

Hey @petebachant, yes, sorry for the delay. You can look at the `wait_for_scheduler` function. It basically runs `countDocuments` on the alert collection every second and then returns when the count reaches the expected number. The program ends after that, at which point the exit trap runs to terminate the subprocesses for the consumer and scheduler (see `cleanup`).

Theodlz (commented):
Some comments (same as on slack pretty much):
- a db name parameter is needed. This script has no knowledge of the db name specified in the config.yaml, so it needs to be configurable so it can match the settings of the user. hardcoding something that we made configurable somewhere else should always be avoided.
- maybe that should be documented, but again since this is decoupled from the config.yaml, how do I know the number of workers that are used for a run, etc. I understand its hard to add that in there (annoying to parse a yaml in bash) so at least mentioning that in some documentation (saying that the user should keep the number of threads used for processing in mind, if they want a per-thread metric).
- this is only benchmarking the alert worker, do you want to include benchmarking the rest? If not, best to tell the user to set the number of ml workers to 0, so that resource isn't used for nothing (which would have an impact on the end result, especially on smaller configs).

Theodlz (approved):
based on a discussion we had over slack, this is LGTM.

Jake suggested using the `-p` flag of docker compose to have this run in a custom namespace (project name, the thing that appears as a prefix of services names and volumes created by docker compose), ensuring that we do not use any container and volumes used by other processes.

jesaerys:
@Theodlz @petebachant Regarding the latest commit that includes the markdown report, I'm totally open to other ideas for places to store these kinds of documents. It's not really material for any paper, but it might also be nice to refer to later if the cost of tracing ever comes up again. I considered putting it in a wiki page here in this repo, but I don't like that there's no review process for that. Given this uncertainty, I also didn't work too hard on making the python env turnkey reproducible (but with just a few dependencies, which I listed in the script, and a somewhat modern python version, it should be fine in this case).

And regarding the inclusion of the svg files, each one is 10 or so kb, so I figured that shouldn't be an issue. I'm open to other ideas, though.

jesaerys:
Ok, @Theodlz @petebachant that's the last commit I plan for this PR.

docs/tracing-throughput-test/tracing-throughput-test.py
```diff
@@ -0,0 +1,57 @@
+# altair[all]>=5.5.0
+# pandas>=2.3.1
+# vl-convert-python>=1.8.0
```
petebachant, line 3: Is this just documentation, or does the script somehow automatically run with these dependencies? From what I've seen, if you write them like:

```python
# /// script
# dependencies = [
#   "altair[all]>=5.5.0",
#   "pandas>=2.3.1",
#   "vl-convert-python>=1.8.0",
# ]
# ///
```

`uv` will automatically run the script in a virtual environment for you with `uv run docs/tracing-throughput-test.py`.

May also want to document where this script is supposed to be run from.

docs/tracing-throughput-test/tracing-throughput-test.md
petebachant, line 1: Nit: Docs build systems like `mkdocs` would name this file `index.md` since it's in a subdirectory.

petebachant (approved):
A couple minor questions/comments, but LGTM! The arguments sound well supported so it feels safe to move on with instrumentation in place.

docs/tracing-throughput-test/tracing-throughput-test.py
Theodlz: Yeah I assumed this was already the `uv` automated thing but you are right, the syntax isn't quite the same as the other script we have in the codebase already. Would be good to use the same thing

jesaerys: 😱 cool! I did not know this.

I was hoping it was clear that the script should be run within the `docs/tracing-throughput-test` directly. Any ideas for how to make more obvious?

docs/tracing-throughput-test/tracing-throughput-test.md
jesaerys: Is the idea that using that convention makes it easier to adopt a tool like `mkdocs` if we decide to in the future?

PR #172, opened 2025-06-20
Add documentation for alert processing pipeline

petebachant:
Not sure if this belongs in `ARCHITECTURE.md`, but I figured I'd follow the pattern in #168. This is helpful for me to visualize how the production system will work and how it might scale with additional surveys. I'm sure there are errors.

To view the Mermaid diagram of the system architecture, click the "view file" link from the "files changed" tab:

image

### DIFF
--- a/docs/alert-processing.md
+++ b/docs/alert-processing.md
@@ -15,7 +15,7 @@ Each alert is processed with the following pipeline:
    of the object in the alert.
 3. Machine learning model classification scores are added.
 4. A set of user-defined filters are applied.
-   Any alert that passes through any filter is sent
+   Any alert that passes through at least one filter is sent
    to a dedicated Kafka output stream for that alert's input stream.
 
 Data flows through the system as follows:
--- a/docs/alert-processing.md
+++ b/docs/alert-processing.md
@@ -1,7 +1,7 @@
 # Alert processing
 
 BOOM consumes Kafka streams of alerts from astronomical surveys
-and outputs Kafka streams for consumers like Skyportal.
+and outputs Kafka streams for consumers like SkyPortal.
 
 Each alert is processed with the following pipeline:
 
@@ -17,3 +17,54 @@ Each alert is processed with the following pipeline:
    value for each. Any alert that passes through a given filter is sent
    to a dedicated Kafka output stream for that filter.
 4. The alert is save in a database for later querying and cross-matching.
+
+The implementation is as follows:
+
+```mermaid
+graph LR
+    subgraph Kafka
+        ZTF[ZTF Kafka stream]
+        LSST[LSST Kafka stream]
+        Output1[Output Kafka stream 1]
+        Output2[Output Kafka stream 2]
+        OutputN[Output Kafka stream N]
+    end
+
+    subgraph Valkey
+        ZTFAlertQueue[ZTF alert queue]
+        LSSTAlertQueue[LSST alert queue]
+    end
+
+    subgraph BOOM services
+        ZTFConsumer[ZTF Kafka consumer]
+        LSSTConsumer[LSST Kafka consumer]
+        CrossMatcher["Alert worker (database insertion and cross-matching)"]
+        MLWorker[ML worker]
+        FilterWorker[Filter worker]
+    end
+
+    subgraph BOOM consumers
+        SkyPortal[SkyPortal]
+    end
+
+    ZTF --> ZTFConsumer
+    LSST --> LSSTConsumer
+    ZTFConsumer --> ZTFAlertQueue
+    LSSTConsumer --> LSSTAlertQueue
+    ZTFAlertQueue --> CrossMatcher
+    LSSTAlertQueue --> CrossMatcher
+    CrossMatcher --> ZTFAlertQueue
+    CrossMatcher --> LSSTAlertQueue
+    ZTFAlertQueue --> MLWorker
+    LSSTAlertQueue --> MLWorker
+    MLWorker --> ZTFAlertQueue
+    MLWorker --> LSSTAlertQueue
+    ZTFAlertQueue --> FilterWorker
+    LSSTAlertQueue --> FilterWorker
+    FilterWorker -- Pass filter --> Output1
+    FilterWorker -- Pass filter --> Output2
+    FilterWorker -- Pass filter --> OutputN
+    Output1 --> SkyPortal
+    Output2 --> SkyPortal
+    OutputN --> SkyPortal
+```
--- a/docs/alert-processing.md
+++ b/docs/alert-processing.md
@@ -38,9 +38,11 @@ graph LR
     subgraph BOOM services
         ZTFConsumer[ZTF Kafka consumer]
         LSSTConsumer[LSST Kafka consumer]
-        CrossMatcher["Alert worker (database insertion and cross-matching)"]
-        MLWorker[ML worker]
-        FilterWorker[Filter worker]
+        subgraph Scheduler
+            CrossMatcher["Alert worker (database insertion and cross-matching)"]
+            MLWorker[ML worker]
+            FilterWorker[Filter worker]
+        end
     end
 
     subgraph BOOM consumers
--- a/docs/alert-processing.md
+++ b/docs/alert-processing.md
@@ -18,7 +18,7 @@ Each alert is processed with the following pipeline:
    Any alert that passes through any filter is sent
    to a dedicated Kafka output stream for that alert's input stream.
 
-The implementation is as follows:
+Data flows through the system as follows:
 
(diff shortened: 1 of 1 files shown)


petebachant:
Not sure if I got this right, but if I did, I see some room for simplification if we condense the entire pipeline into a single worker instead of 3, so we aren't sending messages back and forth via Valkey:

image

docs/alert-processing.md
```diff
@@ -0,0 +1,72 @@
...
+    LSSTAlertQueue --> AlertWorker
+    AlertWorker --> ZTFAlertQueue
+    AlertWorker --> LSSTAlertQueue
+    ZTFAlertQueue --> MLWorker
+    LSSTAlertQueue --> MLWorker
+    MLWorker --> ZTFAlertQueue
+    MLWorker --> LSSTAlertQueue
+    ZTFAlertQueue --> FilterWorker
+    LSSTAlertQueue --> FilterWorker
+    FilterWorker -- Pass filter --> Output1
```
Theodlz, line 66: To be clearer about one filter = one output stream, I would do

```
FilterWorker -- Pass filter 1 --> Output1
```

docs/alert-processing.md
```diff
@@ -0,0 +1,72 @@
...
+    FilterWorker -- Pass filter --> Output2
+    FilterWorker -- Pass filter --> OutputN
+    Output1 --> SkyPortal
```
Theodlz, line 69: Let's make it more general maybe and just say `listener` or `marshal`. It doesn't have to be skyportal specific.

docs/alert-processing.md
```diff
@@ -0,0 +1,72 @@
...
+            MLWorker[ML worker]
+            FilterWorker[Filter worker]
+        end
+    end
+
+    subgraph BOOM consumers
+        SkyPortal[SkyPortal]
+    end
+
+    ZTF --> ZTFConsumer
```
Theodlz, line 52: While ZTF and LSST are good examples because that's we support right now, we are already adding support for more surveys so making it so specific might not be a good representation of how extensible and flexible we are. I would use placeholders instead and mention in text that the first obvious application is to do this with ZTF that we cross match with LSST.

docs/alert-processing.md
```diff
@@ -0,0 +1,72 @@
...
+and outputs Kafka streams for consumers like SkyPortal.
+
+Each alert is processed with the following pipeline:
+
+1. Cross-matches with object IDs from other data catalogs are added.
+   This is done based on the location (right ascension and declination)
+   of the object in the alert. So, for example, if Rubin sends an alert
+   for a given object and then ZTF sends an alert for that same object later,
+   the ZTF alerts will have the object ID from the Rubin catalog.
+   However, the Rubin alert has already been sent out, so it will not have
```
Theodlz, line 13: I would not mention that, since well we could make that change if we wanted in theory (and we might).

docs/alert-processing.md
```diff
@@ -0,0 +1,72 @@
+# Alert processing
+
+BOOM consumes Kafka streams of alerts from astronomical surveys
+and outputs Kafka streams for consumers like SkyPortal.
+
+Each alert is processed with the following pipeline:
+
+1. Cross-matches with object IDs from other data catalogs are added.
```
Theodlz, line 8: While it definitely is one of the important steps carried by the alert worker, it might seem like an oversimplification to only talk about crossmatching. If you look at the alert workers a lot of the work we do is standardize the data we receive from each survey to unify the schemas where applicable, format it, split it for each table, ... and crossmatching with archival catalogs (which is omitted here btw) and other alert-based surveys. This should definitely be mentioned in here! Maybe https://docs.google.com/presentation/d/1PGpJQn2qcohc8fVJQqpfcVRy8ZKZmqF0J7xYygQOmWk/edit?usp=sharing can help with figuring out the content of this markdown document.

Theodlz (changes requested):
The diagram looks cool! I wonder if some elements can be organized as its a bit hard to navigate as is. I also made a number of suggestions as some important information about the processing that BOOM does is omitted here.

jesaerys (commented):
Two overall comments:

1. Add a rectangle for mongodb much like you already have for kafka and valkey. The alert worker pops from its valkey queue, inserts into mongo, and pushes the candid onto the ml valkey queue (rather than pushing the alert back into same queue, as it's currently drawn). Similarly, the ml worker pops from its queue, updates to mongodb, and pushes candids onto the filter valkey queue. I'm less clear on what the filter worker does, I haven't done much with it yet.

2. The pipeline is the same for each survey, so maybe it would be less cluttered to show the flow for one generic survey. Or, maybe use subgraphs, e.g.,

    ```mermaid
    ...
        subgraph Kafka
                subgraph Streams
                ZTF[ZTF Kafka stream]
                LSST[LSST Kafka stream]
                S[Survey *S* Kafka stream]
            end
            ...
        end
        ...
        subgraph BOOM services
            subgraph Consumers
                ZTFConsumer[ZTF Kafka consumer]
                LSSTConsumer[LSST Kafka consumer]
                SConsumer[Survey *S* Kafka consumer]
            end
            ...
        end
        ...
        ZTF --> ZTFConsumer
        LSST --> LSSTConsumer
        S --> SConsumer
        ...
    ```

    (Hopefully that's clear)

docs/alert-processing.md
petebachant: Gotcha. I will definitely add the normalization stuff.

Theodlz:
Also the ML worker and Filter worker, as consequence of being decoupled, can naturally run on batches of alerts at a time rather than one by one (which is required for the alert worker as it's not like we can vectorize over processing each message, whereas we can run the models and filters on batches), which is exactly what we do and where a single alert worker that processes one alert at a time from A to Z does not.

Theodlz:
I think we can easily see that effect in Kowalski, where MLing and filtering represent > 50% of the total processing time, with most of that time spent filtering. That forces us in prod to have > 28 workers just because that one processing step is slow, when here we could only scale that up as needed. If you have a end-to-end worker model then if you add more workers because step C is long, well you also add more resoures to step A and B so C remains a bottleneck by definition since you get more alerts to that step C. I think the approach we took with boom has the big advantage of being a lot less resource hungry actually, given how cheap the valkey round trips are I don't think it offsets that advantage.

Theodlz:
btw @petebachant one thing that was part of the original boom design was to, based on something like a prometheus metric (or whatever lets us measure throughput), scale up/down the number of workers of each type, independently. So if the input queue for filtering starts to grow overtime we can scale the number of filtering workers, and so on.

docs/alert-processing.md
```diff
@@ -9,64 +9,73 @@ Each alert is processed with the following pipeline:
...
-    MLWorker --> LSSTAlertQueue
-    ZTFAlertQueue --> FilterWorker
-    LSSTAlertQueue --> FilterWorker
-    FilterWorker -- Pass filter --> Output1
-    FilterWorker -- Pass filter --> Output2
-    FilterWorker -- Pass filter --> OutputN
-    Output1 --> SkyPortal
-    Output2 --> SkyPortal
-    OutputN --> SkyPortal
+    Input --> KafkaConsumer
```
Theodlz, line 53: Any chance to move the mongodb stuff at the top rather than between the workers and redis? Might help a lot with readability and following which arrows goes where. Also all the workers basically read from all 3 collections. Might be worth just having some bi-directional arrow from mongodb and each worker, rather than to specific collections?

docs/alert-processing.md
```diff
@@ -5,68 +5,77 @@ and outputs Kafka streams for consumers like SkyPortal.
...
-   information about the ZTF alert.
-2. Machine learning model classification scores are added.
-3. A series of user-defined filters are applied, resulting in one boolean
-   value for each. Any alert that passes through a given filter is sent
-   to a dedicated Kafka output stream for that filter.
-4. The alert is saved in a database for later querying and cross-matching.
+   of the object in the alert.
+3. Machine learning model classification scores are added.
+4. A set of user-defined filters are applied.
+   Any alert that passes through any filter is sent
```
Theodlz, line 18:
```suggestion
   Any alert that passes through at least one filter is sent
```

nitpick, but somehow I feel like that helps a little bit? Likely just personal preference so you can ignore that comment

Theodlz (approved):
2 minor comments, main one is about the diagram. Otherwise LGTM

petebachant:
Refactored a bit to include the database. Would like it to be more obvious how the system scales with surveys, but I put it in text.

image

Theodlz:
@petebachant is there anyway to just have the workers point to the MongoDB box rather than individual collections? They all feed from all 3 tables, not just the alert one as this suggests.

PR #176, opened 2025-06-23
FilterWorker(s) module: instrumentation

Theodlz:
This PR adds instrumentation to most of the key methods in the filter worker module, for both ztf and lsst.


jesaerys (approved):
LGTM

PR #177, opened 2025-06-23
Kafka Consumer: custom vs hardcoded topic bugfix

Theodlz:
There is some confusion in the kafka consumer structs and associated binary, as we used to be able to pass custom topics to read from (useful before we were able to automagically infer the topic name from a date) but for ZTF the parameter still existed but was ignored, and for LSST it still existed when we always want to read from the same one topic which holds the simulated data from Rubin.

This in this PR:
- remove the unused topic parameter from the `new` function to instantiate the `ZtfAlertConsumer`
- replace the `topic` parameter by a safer and more explicit `simulated` (bool) parameter which for now we forced to true, and will change to a clap parameter once we have real LSST data and an associated topic to read from.


petebachant (approved):
Seems reasonable to me.

PR #188, opened 2025-06-25
Add necessary dependencies to Dockerfile

petebachant:
Should probably check if these are actually necessary, but they were for me to be able to build a similar image and run the Kafka producer.


petebachant:
cc @Theodlz this might be what you need to run in Docker containers. Might want to just merge.

PR #189, opened 2025-06-25
Fix subtraction overflow in ml worker

### LINKED ISSUE #182: panic, subtraction overflow for command check counter in ml worker
There's a small bug in `run_ml_worker`: the line `command_check_countdown -= nb_candids - 1` will panic with "attempt to subtract with overflow" when `nb_candids -1` is larger than `command_check_countdown`. Maybe we want to use `saturating_sub` instead.

jesaerys:
Closes #182 

The main fix is in the first commit, 09cee74. We now use `saturating_sub` so that decrementing the counter evaluates to zero instead of panicking when the number of candids processed is greater than the counter's value.

The next commit, 3339df3, simplifies the manipulation of the counter. Looking at the code, I realized we didn't need to have little +1/-1 offsets. Hopefully this clarifies the logic a little.

Finally, commit b48f2ce simplifies things a little more by removing what appears to be a redundant check of whether there are any candids to process. In the original code, we first checked the length of the queue, handled the empty-queue case, then later popped from the queue and did another check to handle the empty-vector case. Both checks amount to the same thing, so this commit consolidates that logic. @Theodlz please correct me if I've gotten anything wrong on this point.

### DIFF
--- a/src/ml/base.rs
+++ b/src/ml/base.rs
@@ -65,9 +65,8 @@ pub async fn run_ml_worker<T: MLWorker>(
             if should_terminate(&mut receiver) {
                 break;
             }
-            command_check_countdown = command_interval + 1;
+            command_check_countdown = command_interval;
         }
-        command_check_countdown -= 1;
         // if the queue is empty, wait for a bit and continue the loop
         let queue_len: i64 = con.llen(&input_queue).await?;
         if queue_len == 0 {
@@ -89,7 +88,7 @@ pub async fn run_ml_worker<T: MLWorker>(
         }
 
         let processed_alerts = ml_worker.process_alerts(&candids).await?;
-        command_check_countdown = command_check_countdown.saturating_sub(nb_candids - 1); // As if iterated this many times
+        command_check_countdown = command_check_countdown.saturating_sub(nb_candids); // As if iterated this many times
 
         // push that back to redis to the output queue
         con.lpush::<&str, Vec<String>, usize>(&output_queue, processed_alerts)
--- a/src/ml/base.rs
+++ b/src/ml/base.rs
@@ -64,9 +64,8 @@ pub async fn run_ml_worker<T: MLWorker>(
         if command_check_countdown == 0 {
             if should_terminate(&mut receiver) {
                 break;
-            } else {
-                command_check_countdown = command_interval + 1;
             }
+            command_check_countdown = command_interval + 1;
         }
         command_check_countdown -= 1;
         // if the queue is empty, wait for a bit and continue the loop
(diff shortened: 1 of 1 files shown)


Theodlz:
> Finally, commit [b48f2ce](https://github.com/boom-astro/boom/commit/b48f2ce79d7dca6230c01e62240e3acdd95bee6f) simplifies things a little more by removing what appears to be a redundant check of whether there are any candids to process. In the original code, we first checked the length of the queue, handled the empty-queue case, then later popped from the queue and did another check to handle the empty-vector case. Both checks amount to the same thing, so this commit consolidates that logic. @Theodlz please correct me if I've gotten anything wrong on this point.

So from what I remember, I checked the length before doing a pop because when popping in batches, it seemed to be blocking (i.e. it would not return None or empty but would wait until there was at least one thing to return) and therefore would never continue the loop and get an interrupt signal if there wasn't anything to process. But maybe that was wrong?

Theodlz:
I'm weirded out by the fact that I never got this specific error, but I might have just gotten lucky

Theodlz:
> > and therefore would never continue the loop and get an interrupt signal if there wasn't anything to process.
> 
> Ah, that's a really good thing to check. I think it doesn't block, but I'll cook up a little demo using the same connection and methods used here and verify.



> > and therefore would never continue the loop and get an interrupt signal if there wasn't anything to process.
> 
> Ah, that's a really good thing to check. I think it doesn't block, but I'll cook up a little demo using the same connection and methods used here and verify.

Thank you!

jesaerys:
@Theodlz is this convincing?

```rust
use std::num::NonZero;

use redis::AsyncCommands;

#[tokio::main]
async fn main() {
    // docker run --rm --detach --name foo-valkey --publish 6379:6379 valkey/valkey:7.2.6

    let client = redis::Client::open("redis://localhost:6379").unwrap();
    let mut con = client.get_multiplexed_async_connection().await.unwrap();

    let _: usize = con.del("myqueue").await.unwrap();

    // Basic push and pop:
    let _: usize = con.lpush("myqueue", vec![12345]).await.unwrap();
    let candids: Vec<i64> = con.rpop("myqueue", NonZero::new(1000)).await.unwrap();
    assert_eq!(candids, vec![12345]);

    // Pop again, this time from an empty queue:
    let candids: Vec<i64> = con.rpop("myqueue", NonZero::new(1000)).await.unwrap();
    assert!(candids.is_empty());
}
```

If `rpop` blocked until something was available on the queue, then `main` wouldn't terminate. Yet it does, and it runs without errors.

jesaerys:
Oh hey, maybe you're thinking of `brop`? From https://docs.rs/redis/0.32.2/redis/trait.AsyncCommands.html#method.brpop,

> Remove and get the last element in a list, or block until one is available.

PR #193, opened 2025-06-26
Add ML models to scheduler Docker image

### LINKED ISSUE #192: Add ML models to scheduler Docker image
These currently aren't there, so the ML worker doesn't work!

petebachant:
Resolves #192


Theodlz (approved):
LGTM! Thanks for the fix

PR #196, opened 2025-06-28
DECAM Support: Kafka module

Theodlz:
This PR implements a kafka producer (from file ala ZTF, but without automated download from archive, the user needs to have the data if they wish to produce for decam) and consumer for DECAM.

~~PS: This points should be edited to point to main, once the branch it points to has been merged.~~

PS: As I mentioned, happy to send some data over to the reviewer(s) so they can test this out if they'd like.


Theodlz:
now that we'll have 3 of them, it's becoming clearer and clearer what are the common blocks, so I'm mostly hoping we can deduplicate stuff as much as possible.

src/kafka/ztf.rs
```diff
@@ -313,7 +313,13 @@ impl AlertProducer for ZtfAlertProducer {
...
                 continue;
             }
-            let payload = std::fs::read(path).unwrap();
+            let payload = match std::fs::read(&path) {
+                Ok(data) => data,
+                Err(e) => {
+                    error!("Failed to read file {:?}: {}", path.to_str(), e);
+                    continue;
+                }
+            };
```
jesaerys, line 322: This could be a `let-else`:

```rust
let Ok(payload) = std::fs::read(&path) else {
    error!("Failed to read file {:?}: {}", path.to_str(), e);
    continue
}
```

src/kafka/decam.rs
```diff
@@ -0,0 +1,247 @@
...
+        n_threads: usize,
+        max_in_queue: Option<usize>,
+        output_queue: Option<&str>,
+        group_id: Option<&str>,
+        server: Option<&str>,
+        config_path: &str,
+    ) -> Self {
+        if 15 % n_threads != 0 {
+            panic!("Number of threads should be a factor of 15");
+        }
```
jesaerys, line 37: The other surveys take a different approach to this as of #201. Do you want to make those changes here, too?

src/kafka/decam.rs
```diff
@@ -0,0 +1,247 @@
...
+
+    #[instrument(skip(self))]
+    async fn consume(&self, timestamp: i64) {
+        let partitions_per_thread = 15 / self.n_threads;
+        let mut partitions = vec![vec![]; self.n_threads];
+        for i in 0..15 {
+            partitions[i / partitions_per_thread].push(i as i32);
+        }
+
+        // DECAM uses nightly topics, and no user/pass (IP whitelisting)
```
jesaerys, line 78: For my own understanding: this is exactly the same as for ZTF, then, right?

jesaerys (commented):
Seems straightforward. My main two comments have to do with matching the changes made in #201.

src/kafka/decam.rs
Theodlz: Yes, this PR pre dates 201 so the changes that are in there aren't in sync yet.

src/kafka/decam.rs
Theodlz: Truth is right now they dont even have topic whatsoever, so Ill just replace that comment

src/kafka/decam.rs
```diff
@@ -0,0 +1,247 @@
...
+                Ok(data) => data,
+                Err(e) => {
+                    error!("Failed to read file {:?}: {}", path.to_str(), e);
+                    continue;
+                }
+            };
+
+            let record = FutureRecord::to(&topic_name)
+                .payload(&payload)
+                .key("decam")
```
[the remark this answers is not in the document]
Theodlz: Yes, so will be added once 201 is wrapped up and merged, if we merge 201 before. Actually I'm fine merging this one first since it's pretty much ready to go.

PR #199, opened 2025-06-30
DECAM Support: Filter module

Theodlz:


src/filter/decam.rs
```diff
@@ -0,0 +1,378 @@
...
+pub struct DecamFilter {
+    id: i32,
+    pipeline: Vec<Document>,
+}
+
+#[async_trait::async_trait]
+impl Filter for DecamFilter {
+    #[instrument(skip(filter_collection), err)]
+    async fn build(
+        filter_id: i32,
```
petebachant, line 21: This method might be better named as `from_db`, or at least have some sort of docstring explaining what "build" means. It seems that we're prepending some steps onto the user-defined filter in order to make them work properly. `from_db_user_filter` might also be a decriptive method name.

PR #200, opened 2025-07-04
make consumer able to use given offsets

### LINKED ISSUE #170: Intermittent kafka issues for consumer
While working on #165, which has involved repeatedly running the producer-consumer-scheduler setup many, many times---way more than I've ever run it before---it has become apparent that the consumer often fails to read alerts from kafka. I have no details other than the following observations:

1. The scheduler says "queue is empty" even though I *know* there are alerts in the kafka topic.
2. The issue is intermittent. I can run the very same consumer binary multiple times in a row (no code changes, no changes to the docker containers) and approximately half of the runs will go smoothly and the other half will manifest this issue.
3. Sometimes the consumer will read *some* of the alerts from the topic and push them to the queue and then stop, though most of the time the issue is that the consumer doesn't read any alerts at all.
4. This doesn't seem to be an offset issue. If it were, I don't think it would be possible for some runs to work and others to fail.
5. It's not a valkey issue. I can trace the issue back to `boom::kafka::base::consume_partitions` where we call `consumer.poll` (that's rdkafka) and get None.

Maybe we can look how partitions are assigned.

I think the first thing to do is get better logging coverage so we can troubleshoot more easily.

jesaerys:
Closes #170.

The call to `consumer.subscribe` is removed so that `consumer.assign` can work properly. tl;dr `subscribe` and `assign` are incompatible APIs, and calling `subscribe` was preventing the subscriber from using the timestamp-based offsets we assign to it. See https://github.com/boom-astro/boom/issues/170#issuecomment-3034302595 for a more detailed explanation.

I've run this a number of times on the same topic (only running the producer once, not touching the topic after that), and it has worked every time. *If* there is still an issue with the consumer, I'm confident that it's going to be pretty rare.

Other changes in this PR:
* I Added some tracing stuff for rdkafka/librdkafka so we can now optionally log their events. For instance, by setting `RUST_LOG=librdkafka=trace,rdkafka=debug,boom=info`.
* I added some instrumentation
* I added `ConsumerError` for a little extra error handling
* I added a healthcheck for the kafka container. This isn't critical for anything in our workflow, but our other containers had healthchecks so it seemed good for kafka to have one, too.
* I've touched up the consumer CLI (see commit message for details)

### DIFF
--- a/src/kafka/ztf.rs
+++ b/src/kafka/ztf.rs
@@ -267,6 +267,7 @@ impl AlertProducer for ZtfAlertProducer {
             .set("acks", "1")
             .set("max.in.flight.requests.per.connection", "5")
             .set("retries", "3")
+            .set("debug", "broker,topic,msg")
             .create()
             .expect("Producer creation error");
 
--- a/src/kafka/ztf.rs
+++ b/src/kafka/ztf.rs
@@ -4,7 +4,7 @@ use crate::{
     utils::{
         data::{count_files_in_dir, download_to_file},
         enums::ProgramId,
-        o11y::as_error,
+        o11y::{as_error, log_error},
     },
 };
 use indicatif::ProgressBar;
@@ -108,15 +108,17 @@ impl AlertConsumer for ZtfAlertConsumer {
                     &config_path,
                 )
                 .await;
-                if let Err(e) = result {
-                    error!("Error consuming partitions: {:?}", e);
+                if let Err(error) = result {
+                    log_error!(error, "failed to consume partitions");
                 }
             });
             handles.push(handle);
         }
 
         for handle in handles {
-            handle.await.unwrap();
+            if let Err(error) = handle.await {
+                log_error!(error, "failed to join task");
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


src/kafka/ztf.rs
```diff
@@ -113,15 +118,20 @@ impl AlertConsumer for ZtfAlertConsumer {
         for handle in handles {
             handle.await.unwrap();
```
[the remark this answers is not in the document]
jesaerys: Addressed in [a13e7e7](https://github.com/boom-astro/boom/pull/200/commits/a13e7e7ff58f85f6670ac65743229bd58324f366)

src/kafka/ztf.rs
```diff
@@ -267,6 +279,7 @@ impl AlertProducer for ZtfAlertProducer {
             .set("acks", "1")
             .set("max.in.flight.requests.per.connection", "5")
             .set("retries", "3")
+            .set("debug", "broker,topic,msg")
```
Theodlz, line 282: Just curious, does that have any known performance impact when set?

Theodlz (approved):
LGTM! Thanks a lot for going into the - painful - rabbit hole that is kafka

src/kafka/ztf.rs
jesaerys: I'd be surprised if there's any impact on performance because and have to enable these messages using RUST_LOG and tracing doesn't bother constructing messages unless there's a broker that's interested in consuming them. But, getting an actual answer to this question is part of #156.

Issue #194, opened 2025-06-27
`kafka_producer` only uses one partition


jesaerys:
We have this bit of code in the producer:

```rust
let record = FutureRecord::to(&topic_name)
    .payload(&payload)
    .key("ztf")
    .timestamp(chrono::Utc::now().timestamp_millis());
```

This means that every message we produce has the exact same partition key ("ztf"), so all messages are produced on the same partition. It seems like this isn't what we want, given that we've specified in our compose file that the kakfa instance should have 15 partitions. But then again, assigning all messages the same key guarantees ordering within that partition, and maybe we want to make sure the consumer gets all of the alerts in order. We have to decide what's more important to us. (And if ordering isn't so important, it's probably easiest to omit the `key` method altogether and let kafka distribute the messages across the partitions automatically.)

I don't know if this causes, contributes to, or is unrelated to the consumer issue #170, but it's certainly something that needs to be fixed.

## Evidence: tracing

I enabled the "tracing" feature in `rdkafka`, added `.set("debug", "broker,topic,msg")` to the client config, then ran the producer at the debug level:

```
$ RUST_LOG=librdkafka=trace,rdkafka=debug ./target/debug/kafka_producer ztf 20240617
...(a bunch of startup stuff, including the creation of 15 partitions 0..14)
2025-06-27T13:30:57.138327Z DEBUG /.../rdkafka-0.37.0/src/client.rs:86: librdkafka: PRODUCE [thrd:localhost:9092/bootstrap]: localhost:9092/1: ztf_20240617_programid1 [10]: Produce MessageSet with 1 message(s) (69476 bytes, ApiVersion 7, MsgVersion 2, MsgId 0, BaseSeq -1, PID{Invalid}, uncompressed)
2025-06-27T13:30:57.138348Z DEBUG /.../rdkafka-0.37.0/src/client.rs:86: librdkafka: MSGSET [thrd:localhost:9092/bootstrap]: localhost:9092/1: ztf_20240617_programid1 [10]: MessageSet with 1 message(s) (MsgId 0, BaseSeq -1) delivered
...(etc. for all messages)
```

Every one of these PRODUCE and MSGSET lines shows `ztf_20240617_programid1 [10]` in the included context, which indicates that the messages are being produced on partition 10.

[...]

PR #201, opened 2025-07-09
Kafka module: Deduplication + topic initialization + proper partitioning

Theodlz:
Initially, I just wanted to make sure that we do not produce to only one partition anymore: we currently hardcode the "key" of each record to "ztf" at the moment, but the key is supposed to be what defines what partition a record goes to => they all go to one partition.

While doing so, I realized that if we do want to assign records to unique keys (so they populate all the topics) it would be best to know the exact number of topics in use by the topic we are producing to / consuming from.

- In the producer case, we now create the topic if it does not exist already, with the expected number of partitions for that survey. If it already exists, we get the number of partitions in use for that existing topic.
- In the consumer case, we wait until the topic exists before we start consuming anything, and once we find it retrieve the number of partitions used by that topic so we can do the proper partition assignment between consumer threads.

EDIT: After merging #196 which adds a kafka consumer and producer for a 3rd survey, duplication (or rather, annoying duplication) became a lot more obvious, so this PR was updated to:
- remove all `consume()` methods from the survey's `AlertConsumer` implementations, in favor of a generic method directly implemented on the trait, and instead require each class to implement some simple accessors so the generic method can access the relevant quantities.
- same changes for the `producer()` methods from the survey's `AlertProducer` implementations, to deduplicate the main logic.
- it does feel like these accessors are adding some duplication too but my opinion is that level of duplication is very very minimal and requires very little effort for a developer to implement vs re-implementing the full consume() and produce() methods, which then need to be updated for all surveys once we add new features. I wonder if there is a more elegant way to remove more duplication. But also, I'm pretty happy w/ how thinks look now compared to what they did earlier.

~PS: #203 should be reviewed and merged first, and this PR merged with main once that's done~

### DIFF
--- a/src/kafka/base.rs
+++ b/src/kafka/base.rs
@@ -1,14 +1,73 @@
 use crate::{conf, utils::o11y::as_error};
 
+use rdkafka::admin::{AdminClient, AdminOptions, NewTopic, TopicReplication};
+use rdkafka::client::DefaultClientContext;
 use rdkafka::config::ClientConfig;
 use rdkafka::consumer::{BaseConsumer, Consumer};
 use rdkafka::message::Message;
+use rdkafka::producer::{FutureProducer, Producer};
 use redis::AsyncCommands;
 use tracing::{debug, error, info, instrument, trace};
 
 #[async_trait::async_trait]
 pub trait AlertProducer {
     async fn produce(&self, topic: Option<String>) -> Result<i64, Box<dyn std::error::Error>>;
+}
+
+pub async fn ensure_kafka_topic_partitions(
+    bootstrap_servers: &str,
+    topic_name: &str,
+    expected_nb_partitions: usize,
+) -> Result<usize, Box<dyn std::error::Error>> {
+    let producer: FutureProducer = ClientConfig::new()
+        .set("bootstrap.servers", bootstrap_servers)
+        .create()?;
+
+    let admin_client: AdminClient<DefaultClientContext> = ClientConfig::new()
+        .set("bootstrap.servers", bootstrap_servers)
+        .create()?;
+
+    // Fetch all topics
+    let existing_topics = producer
+        .client()
+        .fetch_metadata(None, std::time::Duration::from_secs(5))?
+        .topics()
+        .iter()
+        .map(|t| t.name().to_string())
+        .collect::<Vec<String>>();
+
+    let topic_exists = existing_topics.contains(&topic_name.to_string());
+    info!("Existing topics: {:?}", existing_topics);
+
+    let nb_partitions = if topic_exists {
+        info!("Topic {} exists, using existing partitions", topic_name);
+        producer
+            .client()
+            .fetch_metadata(Some(topic_name), std::time::Duration::from_secs(5))?
+            .topics()
+            .iter()
+            .find(|t| t.name() == topic_name)
+            .map(|t| t.partitions().len())
+            .unwrap_or(expected_nb_partitions)
+    } else {
+        info!("Topic {} does not exist, creating it", topic_name);
+        let opts = AdminOptions::new().operation_timeout(Some(std::time::Duration::from_secs(5)));
+        admin_client
+            .create_topics(
+                &[NewTopic::new(
+                    topic_name,
+                    expected_nb_partitions as i32,
+                    TopicReplication::Fixed(1),
+                )],
+                &opts,
+            )
+            .await
+            .map_err(|e| format!("Failed to create topic {}: {}", topic_name, e))?;
+        info!("Topic {} created successfully", topic_name);
+        expected_nb_partitions
+    };
+
+    Ok(nb_partitions)
 }
 
 #[derive(Debug, thiserror::Error)]
--- a/src/kafka/base.rs
+++ b/src/kafka/base.rs
@@ -1,74 +1,129 @@
 use crate::{
-    conf,
-    utils::{enums::Survey, o11y::as_error},
+    conf::{self, SurveyKafkaConfig},
+    utils::o11y::{as_error, log_error},
 };
 
-use rdkafka::admin::{AdminClient, AdminOptions, NewTopic, TopicReplication};
 use rdkafka::client::DefaultClientContext;
 use rdkafka::config::ClientConfig;
 use rdkafka::consumer::{BaseConsumer, Consumer};
 use rdkafka::message::Message;
-use rdkafka::producer::{FutureProducer, Producer};
+use rdkafka::{
+    admin::{AdminClient, AdminOptions, NewTopic, TopicReplication},
+    error::KafkaError,
+};
 use redis::AsyncCommands;
-use tracing::{debug, error, info, instrument, trace};
+use tracing::{debug, error, info, instrument, trace, warn};
 
-#[async_trait::async_trait]
-pub trait AlertProducer {
-    async fn produce(&self, topic: Option<String>) -> Result<i64, Box<dyn std::error::Error>>;
-}
-
-pub async fn ensure_kafka_topic_partitions(
+// same as ensure_kafka_topic_partitions, but do not attempt to create the topic
+// simply check that the topic exists and return the number of partitions
+pub async fn check_kafka_topic_partitions(
     bootstrap_servers: &str,
     topic_name: &str,
-    expected_nb_partitions: usize,
-) -> Result<usize, Box<dyn std::error::Error>> {
-    let producer: FutureProducer = ClientConfig::new()
+) -> Result<Option<usize>, KafkaError> {
+    let consumer: BaseConsumer = ClientConfig::new()
         .set("bootstrap.servers", bootstrap_servers)
         .create()?;
 
+    let metadata = consumer.fetch_metadata(None, std::time::Duration::from_secs(5))?;
+
+    debug!("Existing topics:");
+    for topic in metadata.topics() {
+        debug!(" - {}", topic.name());
+    }
+
+    let topic_metadata = metadata.topics().iter().find(|t| t.name() == topic_name);
+    match topic_metadata.map(|t| t.partitions().len()) {
+        Some(partitions) => Ok(Some(partitions)),
+        None => Ok(None),
+    }
+}
+
+pub async fn assign_partitions_to_consumers(
+    topic_name: &str,
+    nb_consumers: usize,
+    kafka_config: &SurveyKafkaConfig,
+) -> Result<Vec<Vec<i32>>, ConsumerError> {
+    // call check_kafka_topic_partitions to ensure the topic exists (it returns the number of partitions)
+    // do so in a while until the topic exists
+    let mut nb_partitions = 0;
+    let mut topic_exists = false;
+    while !topic_exists {
+        match check_kafka_topic_partitions(&kafka_config.consumer, &topic_name).await {
+            Ok(Some(partitions)) => {
+                nb_partitions = partitions;
+                topic_exists = true;
+            }
+            Ok(None) => {
+                info!("Topic {} does not exist yet, retrying...", &topic_name);
+                std::thread::sleep(core::time::Duration::from_secs(5));
+            }
+            Err(e) => {
+                return Err(ConsumerError::from(e));
+            }
+        }
+    }
+
+    let nb_consumers = nb_consumers.clone().min(nb_partitions);
+
+    let mut partitions = vec![vec![]; nb_consumers];
+    for i in 0..nb_partitions {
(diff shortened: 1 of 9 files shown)


Theodlz:
This adressed #194

src/kafka/lsst.rs
```diff
@@ -87,11 +83,33 @@ impl AlertConsumer for LsstAlertConsumer {
...
+        let mut nb_partitions = 0;
+        let mut topic_exists = false;
+        while !topic_exists {
+            match check_kafka_topic_partitions(LSST_SERVER_URL, &topic).await {
+                Ok(Some(partitions)) => {
+                    nb_partitions = partitions;
+                    topic_exists = true;
+                }
+                Ok(None) => {
+                    error!("Topic {} does not exist yet, retrying...", topic);
```
jesaerys, line 98:
```suggestion
                    info!("Topic {} does not exist yet, retrying...", topic);
```

src/kafka/ztf.rs
```diff
@@ -74,16 +75,38 @@ impl AlertConsumer for ZtfAlertConsumer {
...
+        let mut nb_partitions = 0;
+        let mut topic_exists = false;
+        while !topic_exists {
+            match check_kafka_topic_partitions(ZTF_SERVER_URL, &topic).await {
+                Ok(Some(partitions)) => {
+                    nb_partitions = partitions;
+                    topic_exists = true;
+                }
+                Ok(None) => {
+                    error!("Topic {} does not exist yet, retrying...", topic);
```
jesaerys, line 93:
```suggestion
                    info!("Topic {} does not exist yet, retrying...", topic);
```

src/kafka/ztf.rs
```diff
@@ -74,16 +75,38 @@ impl AlertConsumer for ZtfAlertConsumer {
...
+                    std::thread::sleep(core::time::Duration::from_secs(5));
+                }
+                Err(e) => {
+                    error!("Error checking topic {}: {}", topic, e);
+                    return;
+                }
+            }
+        }
+
+        let n_threads = self.n_threads.clone().min(nb_partitions);
```
jesaerys, line 103: I *think* this shouldn't be necessary. If there are fewer partitions than threads, I think kafka just doesn't assign any partitions to the leftover consumers.

That said, I'm totally ok with conserving threads as you're doing here.

src/kafka/ztf.rs
```diff
@@ -315,16 +343,19 @@ impl AlertProducer for ZtfAlertProducer {
...
+                .key(&key_string)
                 .timestamp(chrono::Utc::now().timestamp_millis());
 
             producer
                 .send(record, std::time::Duration::from_secs(0))
                 .await
                 .unwrap();
 
+            current_key = (current_key + 1) % nb_partitions as u64;
+
```
jesaerys, line 358: Have you looked into not specifying a key? My understanding is that kafka uses round-robin by default, which is basically what you're implementing here.

src/kafka/base.rs
```diff
@@ -11,6 +14,60 @@ pub trait AlertProducer {
...
+        .collect::<Vec<String>>();
+
+    let topic_exists = existing_topics.contains(&topic_name.to_string());
+
+    let nb_partitions = if topic_exists {
+        info!("Topic {} exists, using existing partitions", topic_name);
+        metadata
+            .topics()
+            .iter()
+            .find(|t| t.name() == topic_name)
```
jesaerys, line 47: There may be an opportunity to streamline this function. `find` returns `Some` if the topic is found and `None` if it doesn't exist. If I'm thinking about this right, this means we can get rid of the `existing_topics` part above.

src/kafka/ztf.rs
```diff
@@ -283,6 +306,11 @@ impl AlertProducer for ZtfAlertProducer {
             .create()
             .expect("Producer creation error");
 
+        let nb_partitions =
+            ensure_kafka_topic_partitions(ZTF_SERVER_URL, &topic_name, ZTF_NB_PARTITIONS).await?;
```
jesaerys, line 310: I'm trying to understand why we need `ensure_kafka_topic_partitions`. It creates the topic that would have been produced anyway below and it determines the number of partitions which we already know (`ZTF_NB_PARTITIONS`).

jesaerys (commented):
Left some comments. I'm not too bothered about the style of error handling, especially for the producer. Revisiting/simplifying/unifying our error handling strategy is still on my todo list.

src/kafka/ztf.rs
Theodlz: I'd rather prevent spawning threads if those will be idle anyways

src/kafka/ztf.rs
Theodlz: I think the idea is that:
- in production I have no idea how many partitions (I mean I know by experience but better not make any assumptions that could break things down the line if that changes upstream) the distant server may have created for a given topic, so better query the server for that info (that's a comment for the consumer part)
- in development, I want to mimick the number of partitions expected to have in production. For ZTF that's 15, for LSST that's 45. So, this methods lets us define an expected number of partitions and topics created by our consumer binary will now be guaranteed to use that amount. But if for whatever reason the topic already exists (could have been created by some other software), I need to know the number of partitions that were used so my key logic is valid.

jesaerys: Couldn't you just use `check_kafka_topic_partitions` to get the number of partitions if the topic already exists and otherwise fall back to `ZTF_NB_PARTITIONS`? Then, `producer.send` part will automatically create the topic if needed so we don't need to use the `AdminClient` to create it first. And *if* it's true that kafka does round-robin partitioning by default, then you don't even need the partition count, anyway.

Theodlz: > Couldn't you just use `check_kafka_topic_partitions` to get the number of partitions if the topic already exists and otherwise fall back to `ZTF_NB_PARTITIONS`? Then, `producer.send` part will automatically create the topic if needed so we don't need to use the `AdminClient` to create it first. And *if* it's true that kafka does round-robin partitioning by default, then you don't even need the partition count, anyway.

Because we need to explicitly create the topic with the expected number of partitions, otherwise it auto creates it with the default number of partitions from docker compose, not the one we might want to use for a given survey to emulate its prod setup. So yeah, creating the topic explicitly is best as it's the only way to set a specific number of partitions to match a real world scenario.

Theodlz: That said I think the method could be refactored to use the other one (the one we use in the consumer), and avoid duplication.

jesaerys: > otherwise it auto creates it with the default number of partitions from docker compose

Good point, I'm on board now.

jesaerys (commented):
Left some comments/suggestions.

Regarding accessors/getters for the `AlertProducer` trait (https://github.com/boom-astro/boom/pull/201#discussion_r2207912877), I think it makes sense to get rid of them and have the `produce` method take arguments instead. The accessor methods aren't used anywhere else, they're only used so supply values to `produce`, so they might as well be args. The trait could then look like this:

```rust
pub trait AlertProducer {
    async fn download_alerts_from_archive(&self) -> Result<i64, Box<dyn std::error::Error>>;

    async fn produce(
        &self,
        topic_name: &str,
        server_url: &str,
        expected_nb_partitions: usize,
        data_folder: &str,
        limit: i32,
        verbose: bool,
    ) -> Result<i64, Box<dyn std::error::Error>> {
        // Pretty much the same implementation
    }
}
```

This makes things easy for the implementors, e.g.,

```rust
#[async_trait::async_trait]
impl AlertProducer for ZtfAlertProducer {
    async fn download_alerts_from_archive(&self) -> Result<i64, Box<dyn std::error::Error>> {
        // Same impl that you currently have
    }
}
```

Then it's used like this in `kafka_producer`:

```rust
    ...
    match args.survey {
        Survey::Ztf => {
            let producer = ZtfAlertProducer::new(date, limit, program_id, &server_url, true);
            producer
                .produce(
                    &producer.topic_name(),
                    &server_url,
                    producer.default_nb_partitions(),
                    &producer.data_directory(),
                    limit as i32,
                    true,
                )
                .await?;
        }
        ...
    }
    ...
```

where `topic_name`, `default_nb_partitions`, and `data_directory` could be regular methods in `impl ZtfAlertProducer`. Actually, it could be simpler because the only reason `ZtfAlertProducer::new` takes `limit`, `server_url`, and `verbose` is to expose them through the accessor methods. So if you remove those fields from `ZtfAlertProducer`, then you'll have,

```rust
    ...
    match args.survey {
        Survey::Ztf => {
            let producer = ZtfAlertProducer::new(date, program_id);
            producer
                .produce(
                    &producer.topic_name(),
                    &server_url,
                    producer.default_nb_partitions(),
                    &producer.data_directory(),
                    limit as i32,
                    true,
                )
                .await?;
        }
        ...
    }
    ...
```

You could instead keep the `ZtfAlertProducer` fields and `new` as-is, and add a method that passes all the args for you:

```rust
impl ZtfAlertProducer {
    ...

    pub async fn run(&self) -> Result<i64, Box<dyn std::error::Error>> {
        let total_pushed = self.produce2(
            &self.topic_name(),
            &self.server_url,
            self.default_nb_partitions(),
            &self.data_directory(),
            self.limit as i32,
            self.verbose,
        ).await?;
        Ok(total_pushed)
    }
}
```

Then `kafka_producer` would look like,

```rust
    ...
    match args.survey {
        Survey::Ztf => {
            let producer = ZtfAlertProducer::new(date, limit, program_id, &server_url, true);
            producer.run().await?;
        }
        ...
    }
    ...
```

These are just some ideas.

src/kafka/base.rs
```diff
@@ -1,17 +1,243 @@
...
+    async fn produce(&self, topic: Option<String>) -> Result<i64, Box<dyn std::error::Error>> {
+        match self.download_alerts_from_archive().await {
+            Ok(count) => count,
+            Err(e) => {
+                error!("Error downloading alerts: {}", e);
+                return Err(e);
+            }
+        };
+
+        let topic_name = topic.unwrap_or_else(|| self.topic_name());
```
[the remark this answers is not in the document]
Theodlz: testing/consume from an arbitrary topic name is very handy for unit testing, so we can test these things without interfering with other tests that are running on the same topic names

Theodlz:
@jesaerys quick question just to clarify one thing: what did you think of the current state of the consumers? In there I already tried to remove most accessors (except those that are actually survey specific), and hoping to do the same for the producer based on your suggestions, but yeah was just curious to know if you think the changes made for the consumer are good or not.

src/kafka/base.rs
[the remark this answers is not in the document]
jesaerys: Is it more straightforward to just let the topic name be an argument?

jesaerys:
@Theodlz For the consumers, I would personally go further and remove the AlertConsumer trait entirely. The AlertConsumer::consume method could easily just be a function. Or, if you feel like there are too many arguments, and you often pass None for the Option args, then you could make a struct named AlertConsumer that’s constructed using the builder pattern.

Honestly, I'm ok with what you have currently if you'd rather not refactor. These are all just suggestions.

src/kafka/base.rs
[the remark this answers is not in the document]
Theodlz: done

src/kafka/base.rs
```diff
@@ -1,17 +1,243 @@
...
+            }
+            Ok(None) => {
+                info!("Topic {} does not exist yet, retrying...", &topic_name);
+                std::thread::sleep(core::time::Duration::from_secs(5));
+            }
+            Err(e) => {
+                return Err(ConsumerError::from(e));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
Theodlz: done

Theodlz:
> Initially, I just wanted to make sure that we do not produce to only one partition anymore: we currently hardcode the "key" of each record to "ztf" at the moment, but the key is supposed to be what defines what partition a record goes to => they all go to one partition.
> 
> While doing so, I realized that if we do want to assign records to unique keys (so they populate all the topics) it would be best to know the exact number of topics in use by the topic we are producing to / consuming from.

Ok so just going back to the PR description for a second, I think that this currently fixes the issues it meant to address initially. I'm fine with further refactoring but I think this should be done in a dedicated ticket @jesaerys @petebachant (mostly saying this because I'd really like to have these bugfixes in and cannot get to such a refactor right now)

src/kafka/base.rs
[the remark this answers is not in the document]
jesaerys: fair

jesaerys (approved):

PR #202, opened 2025-07-10
Rationalize filter IDs as internal UUIDs, not integers from external systems

### LINKED ISSUE #183: Rationalize filter IDs as UUIDs instead of ints
Options:

1. Duplicate logic in API and main package
2. Depend on a shared library
3. Make the filter worker fetch filters from the API (a more proper separation of concerns for service-based architecture, but adds complexity)

petebachant:
BOOM gives a random UUID to each of its filters as an internal ID. It does not accept IDs from outside systems like Kowalski did, since that's not very good from the standpoint of separation of concerns, i.e., ideally only one system should be responsible for storing the filters, and if other systems need filter information, they query the filter manager's API.

As we can see, changes had to be made all over the place to make this happen, so some centralization would be good in a future PR (TODO: Add to backlog). I would recommend a `boom-filters` crate that is totally dedicated to filter handling, and the API can use functions from there.

Resolves #183 

## TODO

- [x] Write our UUID to `_id` in MongoDB, not `id`, and use that for querying as well.


Theodlz:
> BOOM gives a random UUID to each of its filters as an internal ID. It does not accept IDs from outside systems like Kowalski did, since that's not very good from the standpoint of separation of concerns, i.e., ideally only one system should be responsible for storing the filters, and if other systems need filter information, they query the filter manager's API.
> 
> As we can see, changes had to be made all over the place to make this happen, so some centralization would be good in a future PR (TODO: Add to backlog). I would recommend a `boom-filters` crate that is totally dedicated to filter handling, and the API can use functions from there.
> 
> Resolves #183

+1 for the repackaging of boom to avoid duplication between the API and the alert processing stuff, we should do it soon-ish I think

src/filter/ztf.rs
```diff
@@ -437,16 +441,10 @@ impl FilterWorker for ZtfFilterWorker {
...
             let mut results_map: HashMap<i64, Vec<FilterResults>> = HashMap::new();
 
-            let filter_indices = self
-                .filters_by_permission
-                .get(&programid)
-                .ok_or(FilterWorkerError::GetFilterByQueueError)?;
-
-            for i in filter_indices {
-                let filter = &self.filters[*i];
+            for filter in &self.filters {
```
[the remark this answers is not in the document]
Theodlz: Looks like you incorrectly removed the filter by permission implementation @petebachant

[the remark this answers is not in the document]
Theodlz: It's not like LSST here, ZTF alerts has programids and filters are only allowed to run on some, so that's absolutely needed

Theodlz (commented):
LGTM, but one general comment:
While generating custom ids that aren't as cryptic as mongo's default ID system is a good idea for serialization/deserialization, I'm not such a fan of having it be called `id`, because essentially our mongodb documents end up with:
- an `_id` from mongodb, unique
- a `id`, not unique unless we add an index on it
So we have 2 ids and one of them isn't unique unless we remember to create an index.
It would make more sense to call the `id` field here `_id`, to effectively overwrite Mongo's ID system for ours. That way uniqueness is ensured and we don't have 2 ID systems. Deserializating/Serializing `id` to/from `_id` is as simple as using one serde macro on the field of the struct.
Also, the current implementation constrasts with all the alert stuff where we do overwrite the `_id` with the unique ids from the alert data.

petebachant:
> LGTM, but one general comment: While generating custom ids that aren't as cryptic as mongo's default ID system is a good idea for serialization/deserialization, I'm not such a fan of having it be called `id`, because essentially our mongodb documents end up with:
> 
> * an `_id` from mongodb, unique
> * a `id`, not unique unless we add an index on it
>   So we have 2 ids and one of them isn't unique unless we remember to create an index.
>   It would make more sense to call the `id` field here `_id`, to effectively overwrite Mongo's ID system for ours. That way uniqueness is ensured and we don't have 2 ID systems. Deserializating/Serializing `id` to/from `_id` is as simple as using one serde macro on the field of the struct.
>   Also, the current implementation constrasts with all the alert stuff where we do overwrite the `_id` with the unique ids from the alert data.

I remember we ended up discussing this a bit in the API PR a while back, e.g., for users. We ended up creating our own IDs, and in the `get_db` function, we create any necessary indexes. We can do the same for filters. Maybe we need a centralized `boom-db` crate to encapsulate all database interaction. It feels like there is too much direct database manipulation scattered around and it could get fragile.

Theodlz:
> > LGTM, but one general comment: While generating custom ids that aren't as cryptic as mongo's default ID system is a good idea for serialization/deserialization, I'm not such a fan of having it be called `id`, because essentially our mongodb documents end up with:
> > 
> > * an `_id` from mongodb, unique
> > * a `id`, not unique unless we add an index on it
> >   So we have 2 ids and one of them isn't unique unless we remember to create an index.
> >   It would make more sense to call the `id` field here `_id`, to effectively overwrite Mongo's ID system for ours. That way uniqueness is ensured and we don't have 2 ID systems. Deserializating/Serializing `id` to/from `_id` is as simple as using one serde macro on the field of the struct.
> >   Also, the current implementation constrasts with all the alert stuff where we do overwrite the `_id` with the unique ids from the alert data.
> 
> I remember discussing this a bit in the API PR a while back, e.g., for users. We ended up creating our own IDs, and in the `get_db` function, we create any necessary indexes. We can do the same for filters. Maybe we need a centralized `boom-db` crate to encapsulate all database interaction. It feels like there is too much direct database manipulation scattered around and it could get fragile.

I think it's just unnecessary to create an additional id when we could just overwrite mongo's, which removes the index creation requirement entirely + makes the id system match the logic of all the alert processing stuff.

src/bin/add_filter.rs
```diff
@@ -75,17 +73,12 @@ async fn main() {
...
     let collection = db.collection::<mongodb::bson::Document>("filters");
 
-    // add a unique index on filter_id
-    match create_index(&collection, doc! {"filter_id": 1}, true).await {
-        Ok(_) => {}
-        Err(e) => {
-            error!("error creating index on filter_id: {}", e);
-        }
-    }
-
```
jesaerys, line 85: Won't we want to index on this new id field?

jesaerys (commented):
UUID strings as IDs seems much more sensible to me than ints. Other than that, I don't have much to contribute to this one. I'll let you and @Theodlz hash it out if that's ok.

src/bin/add_filter.rs
```diff
@@ -38,11 +35,12 @@ async fn main() {
...
     };
 
-    // create a bson document with filter_id, active, catalog, permissions
+    // Create a bson document with id, active, catalog, permissions
     // group_id, and a fv array with one doc that has a fid field and a pipeline field
+    let filter_id: String = uuid::Uuid::new_v4().to_string();
 
     let filter = doc! {
-        "filter_id": filter_id,
+        "id": filter_id.clone(),
```
petebachant, line 43: Something else strange about this script is that we're going to allow adding filters that have no users attached to them. Maybe that's okay, i.e., we don't need to know where the filters came from so long as they're there and they work? I guess that makes sense if we have one output stream per survey, but then anyone who subscribes to BOOM gets the results of everyone else's filters?

src/filter/ztf.rs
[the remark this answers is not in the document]
petebachant: It might be worth documenting this somewhere as it's confusing to me as an outsider that permissions applies to program IDs. Would it make sense to call the key in the filter `programids` instead of `permissions`?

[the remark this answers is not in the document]
Theodlz: I think filters should ALL have that permissions key. If its for ZTF then well if its empty the filter is ignored. For other surveys (caus currently only ZTF has permissions), that field is ignored and they run on the whole stream.

src/bin/add_filter.rs
Theodlz: - If you are querying filters to have them deserialize/serialize from/to the struct directly, you just add a the following macro on top of your `id` field in the struct:

  ```
  #[serde(rename = "_id")]
  ```
  So when being read from/to mongo, the rust structs `id` is mapped to the document's `_id`. (I believe we might not be reading things directly as filter structs, which we should be.)
- if you are querying the document as is (not serializing/deserializing to/from a struct) then you can just use a projection, where you have:
  ```
  "id": "$_id"
  ```

src/bin/add_filter.rs
Theodlz: This script really is just here for development. We used to not have an API, so that script was the only way to add filters without dealing with mongo directly. We could just make it so that this script gets the admin user assigned to it (which rn is generated when you run the API, but it should really just be part of some generic DB crate/module that both api and "nightly processing" stuff use).

Theodlz:
@petebachant any progress on this? DO you need me to review it again?

api/src/routes/filters.rs
```diff
@@ -9,19 +10,48 @@ use std::vec;
...
+    pub permissions: Vec<i32>,
+    pub user_id: String,
+    pub catalog: String,
+    pub active: bool,
+    pub active_fid: String,
+    pub fv: Vec<FilterVersion>,
+}
+
+impl FilterPublic {
+    pub fn from_filter(filter: Filter) -> Self {
```
petebachant, line 44: Couldn't figure out any way to avoid this, since we have to serialize a response, and we don't want the underscore there.

Theodlz: Don't we `deserialize` when we send to Mongo? In which case maybe you can just have the one private filter struct.

Theodlz:
@petebachant I quickly added a simple endpoint to retrieve a specific filter by ID. This is needed for other systems to easily interact with boom.

Theodlz (approved):
LGTM

Issue #184, opened 2025-06-24
Write deployment documentation


petebachant:
Setting up a CD pipeline here might be part of this, or perhaps it belongs in a different repo, where we keep the Caltech configs. We could, however, at least build the Docker images in GitHub Actions and push them to this repo's container registry.

Theodlz:
@petebachant I think that ...
- updating the docker compose to read the config.yaml from a mounted volume (so we can pass any config and aren't forced to use the one that was shipped in the container if any)
- adding a CI here to push things to the github container registry or docker hub

... would be perfect. For the deployment we can have a private repo with configs + some stuff to version control which image we use to deploy

Theodlz:
How does that sound? I like the idea of having the CI here, since the build itself isn't config dependent

petebachant:
Sounds good! FYI I already took the config YAML out of the Docker image.

PR #204, opened 2025-07-15
Start deployment docs

petebachant:
Towards #184 

Mostly to start discussion and prep for the production deployment.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -1,0 +1,30 @@
+# Deploying a BOOM system
+
+## Option 1: Single node with Docker Compose and a GitHub Actions self-hosted runner
+
+TODO: Point to the Caltech deployment repo as an example?
+Will need to keep secrets out of committed text files.
+
+1. Create a new GitHub repo to store your Docker Compose and BOOM configuration
+   files.
+   This should be private unless BOOM can be configured with secrets as
+   environmental variables instead of in committed text files.
+1. Copy in `docker-compose.default.yaml` and `config.default.yaml`, removing
+   `.default` from the file names.
+1. Change all important secrets in the configs to values that are unique and
+   hard to guess.
+   Change API token expiration time to be greater than zero, e.g., 3600.
+   TODO: These should be set as secrets in the GitHub settings, not out
+   in plain text in the repo files.
+1. Add a BOOM API service to your Docker Compose config.
+   TODO: Or we could deploy with `docker compose up --profile=prod`?
+1. Add one BOOM Kafka consumer service to your Docker Compose config for each
+   survey BOOM will subscribe to.
+1. Add one BOOM scheduler service to your Docker Compose config for each survey
+   BOOM will subscribe to.
+1. TODO: Set up a Traefik reverse proxy config and run that in Docker Compose.
+1. Configure your node as a GitHub Actions self-hosted runner.
+1. Create a `deploy.yaml` GitHub Actions workflow that runs on each release.
+   Ensure its `environment` is set to the environment that points to your
+   GitHub Actions self-hosted runner.
+   TODO: Deploy a staging instance?


README.md
```diff
@@ -44,12 +44,31 @@ BOOM runs on macOS and Linux. You'll need:
...
+by copying it to `.env`:
+
+```sh
+cp .env.example .env
+```
+
+**Note:** Do not commit `.env` to Git or use the example values
+in production.
+
+### Docker and Build Setup
```
petebachant, line 61:
```suggestion
### Start services for local development
```

petebachant:
@Theodlz @jesaerys I think this is ready to at least start a discussion. One major question to figure out is the Kafka consumer one in the PR description.

.env.example
Theodlz, line 1: Kafka-related stuff should live in there too maybe? Right now we expect the user to set LSST kafka credentials as ENV variables but perhaps we should read them from such an .env file instead. We should absolutely avoid having too many ways or rather places to set important parameters, so I'd like to have kafka-related secrets set here too at some point.

README.md
```diff
@@ -44,12 +44,31 @@ BOOM runs on macOS and Linux. You'll need:
...
+### Environment configuration
+
+BOOM uses environment variables for sensitive configuration like passwords
+and API keys.
+For local development, you can use the defaults in `.env.example`
+by copying it to `.env`:
+
+```sh
+cp .env.example .env
+```
```
Theodlz, line 56: reminds me a bit of the default config vs just one config debate here. Let's just make sure that the code uses some sensible defaults if some vars are not defined in there? (might already be the case, just asking)

README.md
```diff
@@ -44,12 +44,31 @@ BOOM runs on macOS and Linux. You'll need:
...
+### Start services for local development
+
+1. Launch `Valkey`, `MongoDB`, and `Kafka` using Docker, with the provided `docker-compose.yaml` file:
     ```bash
     docker compose up -d
     ```
     This may take a couple of minutes the first time you run it, as it needs to download the docker image for each service.
     *To check if the containers are running and healthy, run `docker ps`.*
+
+    **Note:** Docker Compose will automatically use the environment variables from your `.env` file to configure the MongoDB container with your specified credentials.
```
Theodlz, line 70: another big win for docker compose

api/src/conf.rs
```diff
@@ -1,4 +1,100 @@
...
+/// Loads environment variables from a .env file if it exists.
+/// This function should be called early in the application startup,
+/// typically before any configuration loading.
+///
+/// The function looks for .env files in this order:
+/// 1. .env in the current working directory
+/// 2. .env in the parent directory (useful when running from subdirs)
+/// 3. If none found, continues without error (env vars may be set by system)
+pub fn load_dotenv() {
+    // Try current directory first
```
Theodlz, line 26: that does make me wonder how we should handle config locations. right now, the path to the config.yaml is often a parameter we pass to binaries we run. But I have a feeling that the convention might be different for .env files, that are just expected to be where you run your command from, correct?

api/src/conf.rs
```diff
@@ -77,47 +153,52 @@ pub fn load_config(config_path: Option<&str>) -> AppConfig {
...
-        Some(admin_email) => admin_email
-            .clone()
-            .into_string()
-            .unwrap_or_else(|e| panic!("Invalid admin_email: {}", e)),
-        None => default_config.auth.admin_email.clone(),
-    };
+
+    let secret_key = auth_conf
+        .get("secret_key")
+        .expect("secret_key must be set in config")
```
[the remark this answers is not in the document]
Theodlz: I agree with our robot friend here

api/src/conf.rs
```diff
@@ -77,47 +153,52 @@ pub fn load_config(config_path: Option<&str>) -> AppConfig {
...
+        .map(|au| {
+            au.clone()
+                .into_string()
+                .unwrap_or_else(|e| panic!("Invalid admin_username: {}", e))
+        })
+        .unwrap_or_else(|| "admin".to_string()); // Username can have a reasonable default
+
+    let admin_password = auth_conf
+        .get("admin_password")
+        .expect("admin_password must be set in config")
```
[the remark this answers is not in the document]
Theodlz: here too

api/src/conf.rs
```diff
@@ -126,39 +207,62 @@ pub fn load_config(config_path: Option<&str>) -> AppConfig {
...
         .get("username")
         .and_then(|username| username.clone().into_string().ok())
-        .unwrap_or(default_config.database.username);
+        .unwrap_or_else(|| "mongoadmin".to_string()); // Username can have a reasonable default
+
     let password = db_conf
         .get("password")
-        .and_then(|password| password.clone().into_string().ok())
-        .unwrap_or(default_config.database.password);
+        .expect("database password must be set in config")
```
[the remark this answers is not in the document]
Theodlz: same

api/src/db.rs
```diff
@@ -13,29 +13,34 @@ pub const PROTECTED_COLLECTION_NAMES: [&str; 2] = ["users", "filters"];
 
 async fn init_api_admin_user(
     auth_config: AuthConfig,
-    users_collection: &mongodb::Collection<User>,
+    database: &mongodb::Database,
```
Theodlz, line 16: why pass it the whole DB object rather than just the collection it needs?

api/tests/test_auth.rs
```diff
@@ -3,17 +3,39 @@ mod tests {
...
+        let client = mongodb::Client::with_uri_str(uri)
+            .await
+            .expect("failed to connect");
+        let temp_db = client.database(&db_config.name);
+        temp_db
+            .collection::<mongodb::bson::Document>("users")
+            .drop()
+            .await
+            .ok(); // Ignore errors if collection doesn't exist
+
```
Theodlz, line 35: This is a problem. Our tests are async i.e. they can run in parallel and wiping a table clean (which is used by multiple tests) can add flakiness and create unexpected results.

Theodlz: + that lengthy-ish piece of code is duplicated btw

api/tests/test_conf.rs
Theodlz, line 1: Very nice!

api/tests/test_conf.rs
```diff
@@ -0,0 +1,120 @@
...
+mod tests {
+    use boom_api::conf::{load_config, load_dotenv};
+    use std::panic;
+
+    #[test]
+    fn test_token_expiration_validation_fails_with_zero() {
+        load_dotenv();
+
+        // This should panic because token_expiration is 0
+        let result = panic::catch_unwind(|| {
```
Theodlz, line 11: Seems more complicated then I though, any reason to wrap the whole thing in this `panic` statement, including the string definition and such?

config.yaml
```diff
@@ -1,19 +1,19 @@
 database:
   host: localhost
   port: 27017
-  name: boom
+  name: ${BOOM_DB_NAME:-boom}
   max_pool_size: 200
   replica_set: null
-  username: mongoadmin
-  password: mongoadminsecret
+  username: ${BOOM_DB_USERNAME:-mongoadmin}
```
Theodlz, line 7: is the `:-something` syntax there to set defaults? Just curious

Theodlz: asking because I see it for some, but not for others (e.g. the DB password)

src/conf.rs
```diff
@@ -19,6 +23,99 @@ pub enum BoomConfigError {
...
+
+/// Loads environment variables from a .env file if it exists.
+/// This function should be called early in the application startup,
+/// typically before any configuration loading.
+///
+/// The function looks for .env files in this order:
+/// 1. .env in the current working directory
+/// 2. .env in the parent directory (useful when running from subdirs)
+/// 3. If none found, continues without error (env vars may be set by system)
+pub fn load_dotenv() {
```
Theodlz, line 47: I think this PR makes the duplication issue between the main app and the API even more obvious then before, we should really get to this soon-ish...

docs/deployment.md
```diff
@@ -0,0 +1,30 @@
...
+   This should be private unless BOOM can be configured with secrets as
+   environmental variables instead of in committed text files.
+1. Copy in `docker-compose.default.yaml` and `config.default.yaml`, removing
+   `.default` from the file names.
+1. Change all important secrets in the configs to values that are unique and
+   hard to guess.
+   Change API token expiration time to be greater than zero, e.g., 3600.
+   TODO: These should be set as secrets in the GitHub settings, not out
+   in plain text in the repo files.
+1. Add a BOOM API service to your Docker Compose config.
```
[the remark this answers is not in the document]
petebachant: I basically completely rewrote this doc, so I will resolve these comments.

docs/deployment.md
```diff
@@ -0,0 +1,30 @@
...
+   TODO: Or we could deploy with `docker compose up --profile=prod`?
+1. Add one BOOM Kafka consumer service to your Docker Compose config for each
+   survey BOOM will subscribe to.
+1. Add one BOOM scheduler service to your Docker Compose config for each survey
+   BOOM will subscribe to.
```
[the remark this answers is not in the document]
petebachant: I don't think so. This is something I've done with Helm charts for Kubernetes, so if this gets unmanageable, maybe we need to shift to a Kubernetes deployment at some point. Or we can maybe rearchitect to have each service handle multiple surveys?

docker-compose.yaml
```diff
@@ -73,41 +71,103 @@ services:
...
+    networks:
+      - boom
+    profiles:
+      - prod
+    volumes:
+      - ${PWD}/config.yaml:/app/config.yaml
+    command:
+      - /app/kafka_consumer
+      - ztf
+      - --auto-switch-date
```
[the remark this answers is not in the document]
jesaerys: I don't think it should be too hard. If we provide a start date AND a changeover time on the CLI, then we can have a task that shuts down the current consumer and starts up a new one every 24 hours at the changeover time.

src/conf.rs
```diff
@@ -19,6 +23,99 @@ pub enum BoomConfigError {
...
+    debug!("No .env file found, using system environment variables only");
+}
+
+/// Expands environment variable placeholders in a string.
+/// Supports both ${VAR_NAME} and ${VAR_NAME:-default_value} syntax.
+///
+/// Examples:
+/// - "${BOOM_DB_PASSWORD}" -> reads from BOOM_DB_PASSWORD env var
+/// - "${BOOM_DB_PASSWORD:-defaultpass}" -> reads from BOOM_DB_PASSWORD, falls back to "defaultpass"
+fn expand_env_vars(input: &str) -> Result<String, ExpandError> {
```
jesaerys, line 76: Are we sure we need this? It doesn't seem like we should be in the business of parsing shell expansion syntax. If I'm understanding the use case correctly, I think we'd be better served taking a layered config approach where env vars override whatever is in the yaml.

petebachant: So what would you propose as a solution that will allow:
1. Secrets stored in GitHub Actions secrets
2. An app that will not run with weak/exposed secrets unless done very intentionally, i.e., only for local dev

## Concept A

I could take secrets out of the config altogether and only allow setting them via environmental variable, but @Theodlz didn't like that the keys would no longer be visible. On the other hand `.env.example` shows what env vars must be set for BOOM to run properly, so maybe that's illustrative enough.

## Concept B

I could leave them empty by default and have the app error if they aren't either there or in the env vars. However, there's still a risk someone adds secrets into `config.yaml` and ends up committing them.

I'd lean towards concept A if we don't want to mimic Docker Compose and parse the syntax out of YAML as that is the most fool proof against weak/exposed secrets.

petebachant: Just pushed up something that should be a bit less offensive--no parsing env vars from YAML, and use dunderscores for nested env vars that map directly into the config.

Dockerfile-benchmark
Theodlz, line 1: Nice, I also felt like this was just extra noise

config.yaml
```diff
@@ -5,15 +5,15 @@ database:
...
   srv: false
 api:
   auth:
-    secret_key: "1234" # Secret key for JWT authentication, change it to a strong secret
-    token_expiration: 0 # jwt token expiration in seconds, 0 to disable. Should NOT be 0 in production
-    admin_username: admin # Admin username for the API
-    admin_password: adminsecret # Admin password for the API
-    admin_email: admin@example.com # Admin email for the API
+    secret_key: "" # Set via BOOM_API__AUTH__SECRET_KEY environment variable
+    token_expiration: 604800 # JWT token expiration in seconds (7 days) -- must be > 0
```
Theodlz, line 13: any reason for that change? I actually find it quite annoying in a dev setup to not have 0 as the value. Then all the postman/insomnia/... kind of setup where you test API calls break weekly...

docker-compose.yaml
```diff
@@ -73,41 +71,155 @@ services:
...
+    networks:
+      - boom
+    profiles:
+      - prod
+    volumes:
+      - ${PWD}/config.yaml:/app/config.yaml
+    command:
+      - /app/scheduler
+      - lsst
+  babamul-worker:
```
Theodlz, line 199: Nice!

src/conf.rs
```diff
@@ -19,6 +23,99 @@ pub enum BoomConfigError {
...
+    let re = Regex::new(r"\$\{([^}:]+)(?::-(.*?))?\}").unwrap();
```
Theodlz, line 77: I wonder if these patterns can't be set at compile time somehow, at the top of the file rather in the functions. With how regex works, in theory stuff can be pre-computed before runtime, but maybe not with this crate? (which is ok, just making sure we don't repeat work we don't need to)

config.yaml
petebachant: Isn't it just as easy to store a username/password with those tools? My goal is to make it harder to deploy with an less secure config.

Theodlz:
looking at this today @petebachant

src/conf.rs
petebachant: We actually shouldn't need this functionality any longer. We'll read from env vars, but won't expand them from their use in the YAML file.

PR #205, opened 2025-07-16
SchemaCache struct (ala SchemaRegistry)

Theodlz:
Now that we have DECAM alert processing implementing, it's obvious that the `alert_from_avro_bytes` method is pretty much the same as ZTF's. The exception is Lsst which uses a `SchemaRegistry` and therefore has a different implementation. DECAM and ZTF both use have some sort of caching of the schema they read from one alert to another so it can be reused thereafter and we skip re-reading the schema everytime. 

In this PR:
- To  remove some duplication and make things a bit cleaner, we add a `SchemaCache` struct, similar to the `SchemaRegistry` one, with the properties that were hardcoded on the ZTF and DECAM alert workers. 
- Then, we move the `alert_from_avro_bytes` method out of the AlertWorker trait, and remove their survey-specific implementations from all 3 surveys, to move them to the SchemaRegistry and newly added SchemaCache struct. Now, surveys implement one of these 2 data structures to handle schema's and reading data from avro.

PS: SchemaRegistry and SchemaCache now both implement a method with the same name. I'm thinking that we should have a trait that requires the `alert_from_avro_bytes`, for these 2 structs to implement. Any thoughts? Trying to stay rusty here.

### DIFF
--- a/src/filter/ztf.rs
+++ b/src/filter/ztf.rs
@@ -564,18 +564,19 @@ impl FilterWorker for ZtfFilterWorker {
                 )
                 .await?;
 
-                // if the array is empty, continue
+                info!(
+                    "{}/{} ZTF alerts with programid {} passed filter {}",
+                    out_documents.len(),
+                    candids.len(),
+                    programid,
+                    filter.id,
+                );
+
+                // If we have output documents, we need to process them
+                // and create filter results for each document (which contain annotations)
+                // however, if the array is empty, there's nothing to do
                 if out_documents.is_empty() {
                     continue;
-                } else {
-                    // if we have output documents, we need to process them
-                    // and create filter results for each document (which contain annotations)
-                    info!(
-                        "{} alerts passed ztf filter {} with programid {}",
-                        out_documents.len(),
-                        filter.id,
-                        programid,
-                    );
                 }
 
                 let now_ts = chrono::Utc::now().timestamp_millis() as f64;


src/alert/decam.rs
```diff
@@ -98,8 +92,7 @@ pub struct DecamAlertWorker {
     alert_collection: mongodb::Collection<Document>,
     alert_aux_collection: mongodb::Collection<Document>,
     alert_cutout_collection: mongodb::Collection<Document>,
-    cached_schema: Option<Schema>,
-    cached_start_idx: Option<usize>,
+    pub schema_cache: SchemaCache,
```
jesaerys, line 95: `pub` fields are often a sign that something isn't quite right. As far as I can tell, this field is `pub` only because the test `test_decam_alert_from_avro_bytes` (in `test_decam.rs`) needs to access it in this bit:

```rust
        let alert = alert_worker
            .schema_cache
            .alert_from_avro_bytes(&bytes_content);
```

The thing I notice about this test is that it should really be a *unit* test. All it needs is a randomized alert, it doesn't depend on any external services. In Rust, unit tests have access to all private details of the code under test, whereas integration test (all of our tests to date) only have access to the public API. By turning `test_decam_alert_from_avro_bytes` into a unit test, the need for a public field goes away.

In general, I'd like to see us shift to using unit tests whenever possible. And when it isn't possible, to take the time to consider why; it could be a hint that there's a better design out there.

Anyway, my suggestion here is to make this field private:

```suggestion
    schema_cache: SchemaCache,
```

Then, move `test_decam_alert_from_avro_bytes` into a `tests` module at the bottom of the `alert::decam` module:

```rust
#[cfg(test)]
mod tests {
    #[tokio::test]
    async fn test_decam_alert_from_avro_bytes() {
        ...
    }
}
```

src/alert/lsst.rs
```diff
@@ -722,7 +720,7 @@ impl Alert for LsstAlert {
 
 pub struct LsstAlertWorker {
     stream_name: String,
-    schema_registry: SchemaRegistry,
+    pub schema_registry: SchemaRegistry,
```
jesaerys, line 723: See `decam.rs`.

src/alert/ztf.rs
```diff
@@ -400,8 +398,7 @@ pub struct ZtfAlertWorker {
     alert_collection: mongodb::Collection<Document>,
     alert_aux_collection: mongodb::Collection<Document>,
     alert_cutout_collection: mongodb::Collection<Document>,
-    cached_schema: Option<Schema>,
-    cached_start_idx: Option<usize>,
+    pub schema_cache: SchemaCache,
```
jesaerys, line 401: See `decam.rs`

jesaerys (changes requested):
My main feedback is to start using proper unit tests so we don't have to make fields public just for the integration tests to work. Everything else looks great 👍

Regarding having two methods named `alert_from_avro_bytes`, I agree it's worth thinking about what that tells us, though I don't see an obvious issue offhand. Both methods get called in the different `AlertWorker::process_alert` impls
where the concrete type is known anyway. It would only make sense to unify the
two methods if we wanted one generic `alert_from_avro_bytes` method that we call
when the alert type is truly arbitrary. I think it's perfectly ok as-is.

src/alert/base.rs
```diff
@@ -302,6 +304,97 @@ impl SchemaRegistry {
...
+impl SchemaCache {
+    pub fn new(cached_schema: Schema, cached_start_idx: usize) -> Self {
+        SchemaCache {
+            cached_schema: Some(cached_schema),
+            cached_start_idx: Some(cached_start_idx),
+        }
+    }
+
+    #[instrument(skip_all, err)]
+    pub async fn alert_from_avro_bytes<T: Alert>(
```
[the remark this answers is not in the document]
Theodlz: How is that possible? In the LSST cache we need to interrogate the schema registry api, which is async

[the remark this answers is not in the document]
jesaerys: This is about `SchemaCache::alert_from_avro_bytes`. LSST uses `SchemaRegistry::alert_from_avro_bytes`, which indeed needs to be async as you said.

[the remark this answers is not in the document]
Theodlz: Oh my bad, you are 100% right then. Sorry reading through this omw to the airport...

src/alert/decam.rs
Theodlz: How do you feel about doctests (in docstrings) too? I used those a bit in a project, and I really liked how they show up as examples in auto-generated docs. The only thing I don't link is how they may impact readability, though for anyone using a modern-ish IDE that won't be an issue.

src/alert/base.rs
```diff
@@ -302,6 +304,97 @@ impl SchemaRegistry {
...
+
+        let alert: T = from_value::<T>(&value)?;
+
+        Ok(alert)
+    }
+}
+
+pub struct SchemaCache {
+    cached_schema: Option<Schema>,
+    cached_start_idx: Option<usize>,
```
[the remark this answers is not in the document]
Theodlz: Yeah I think it changed the whole thing quite a bit, and the alert worker itself would now have to be responsible of knowing if the cache is initialized or not, so for now I'd rather keep it this way.

Theodlz:
@jesaerys I made the schema_cache and schema_registry fields of the various alert workers private, and moved the unit testing of the from_avro_bytes methods to the various alert modules, as you suggested. I'll let you take another look at this one?

src/alert/decam.rs
jesaerys: Doctests are wonderful.

jesaerys (approved):
LGTM.

I had a minor question about removing `init_indexes`, but I don't think it's a blocker for merging.

PR #209, opened 2025-07-18
produce only when needed

### LINKED ISSUE #101: ZTF Archive Kafka Producer - adding persistence

jesaerys:
Closes #101 

TODO:

- [x] figure out how to count topics without consuming
- [x] update `produce` to only download and produce if the topic has the same number of messages as avro files in the data directory
- [x] refactor as needed so the new code plays nicely with the rest of the module
- [x] figure out how to persist the kafka data in a docker volume


Theodlz:
@jesaerys can I suggest looking into https://hub.docker.com/r/bitnami/kafka ? Looks like that docker image for kafka is WAY more popular than the apache/kafka one we are using, and way more documented (on both docker hub and the web in general). Actually, they even have a documentation section about persisting data!

jesaerys:
I spent some time looking into the `AccessDeniedException` when we try to mount kafka's data directory, and I figured it out. Here's what I did.

First, I wanted to take a step back first and make sure boom would still run without the `KAFKA_LOG_DIRS: /tmp/kraft-combined-logs` part. I removed that setting, and everything still worked. According to the kafka config docs, the default of `log.dir` is `/tmp/kafka-logs`. *Without* the volume definition (how we've normally been running the kafka container), I was able to see partition files showing up there after running the producer. Furthermore, I noticed the owner of that directory was `appuser`.

Then, I added the volume definition back to the compose file (adding `- kafka_data:/tmp/kafka-logs` under `volumes` for the `boom` service), and I again got the `AccessDeniedException` error. I then set `command: sleep 60` for the `boom` service to prevent the container from exiting right away, giving me some time to look around. When I restarted docker compose, I found the owner of `/tmp/kafka-logs` was set to `admin`. *So*, it doesn't matter whether we have `KAFKA_LOG_DIRS` set to `/tmp/kraft-combined-logs` or if we just use the default `/tmp/kafka-logs`. The issue is that trying to mount kafka's data directory to a docker volume sets that directory's owner to `admin`, which breaks the container because kafka can't write to that directory unless the owner is `appuser`.

At one point, I noticed the `apache/kafka:latest` Dockerfile already defines some volumes:

```
$ docker image inspect apache/kafka:latest | jq '.[].Config.Volumes'
{
  "/etc/kafka/secrets": {},
  "/mnt/shared/config": {},
  "/var/lib/kafka/data": {}
}
```

I checked `/var/lib/kafka/data` in a successfully running broker container, and it was completely empty; kafka doesn't do anything with that directory. But, the volumes really do get created, they're just anonymous:

```
$ docker inspect boom-broker-1 | jq '.[].Mounts'
[
  {
    "Type": "volume",
    "Name": "98ea9ff9d88220054894f10d859609c2e3e96eb2c5fc2e6119ba7ad5ac10b0a9",
    "Source": "/var/lib/docker/volumes/98ea9ff9d88220054894f10d859609c2e3e96eb2c5fc2e6119ba7ad5ac10b0a9/_data",
    "Destination": "/etc/kafka/secrets",
    "Driver": "local",
    "Mode": "",
    "RW": true,
    "Propagation": ""
  },
  {
    "Type": "volume",
    "Name": "d64e14728bf8fdf40f6ea866d84fa5c7e1af4c2b854bb0b2809f09f8acadf5df",
    "Source": "/var/lib/docker/volumes/d64e14728bf8fdf40f6ea866d84fa5c7e1af4c2b854bb0b2809f09f8acadf5df/_data",
    "Destination": "/mnt/shared/config",
    "Driver": "local",
    "Mode": "",
    "RW": true,
    "Propagation": ""
  },
  {
    "Type": "volume",
    "Name": "41825d8a6044454d2e3b6d1ae386de0921521f690cabd618cfc327119f36256b",
    "Source": "/var/lib/docker/volumes/41825d8a6044454d2e3b6d1ae386de0921521f690cabd618cfc327119f36256b/_data",
    "Destination": "/var/lib/kafka/data",
    "Driver": "local",
    "Mode": "",
    "RW": true,
    "Propagation": ""
  }
]
```

And, it turns out that `/var/lib/kafka/data` is owned by `appuser`, which means kafka can write to it. I therefore set `KAFKA_LOG_DIR: /var/lib/kafka/data` in the compose file, and it worked! The container was able to run and I could see partition files in `/var/lib/kafka/data` after running the producer.

The only issue was that the volume was anonymous. To fix that, I added the volume definition back to the compose file, using `- kafka_data:/var/lib/kafka/data` instead of `- kafka_data:/tmp/kafka-logs` from before. And that solves it; by sending kafka's data to `/var/lib/kafka/data` and giving the associated volume a name, the state of the `broker` service is persisted.

petebachant:
@jesaerys are there any unknowns here I can help research/debug so we can get it merged for the paper benchmarking?

jesaerys:
@Theodlz @petebachant Ready for review.

src/kafka/base.rs
```diff
@@ -5,56 +5,82 @@ use crate::{
...
-    let topic_metadata = metadata.topics().iter().find(|t| t.name() == topic_name);
-    match topic_metadata.map(|t| t.partitions().len()) {
-        Some(partitions) => Ok(Some(partitions)),
-        None => Ok(None),
-    }
+    Ok(metadata.partition_ids(topic_name).map(|ids| ids.len()))
 }
 
-pub async fn assign_partitions_to_consumers(
+pub fn assign_partitions_to_consumers(
```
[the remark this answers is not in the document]
jesaerys: We could be more careful about our use of `pub`. It would help avoid false positives like this and is just good practice in general.

src/kafka/base.rs
```diff
@@ -5,56 +5,82 @@ use crate::{
...
+                .collect::<Vec<_>>();
+            (name, partition_ids)
+        })
+        .collect::<HashMap<_, _>>();
+    Ok(Metadata(inner))
+}
+
 // check that the topic exists and return the number of partitions
-pub async fn check_kafka_topic_partitions(
+pub fn check_kafka_topic_partitions(
```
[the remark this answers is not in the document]
jesaerys: ditto

tests/test_kafka.rs
```diff
@@ -1,46 +1,93 @@
...
+    let result = producer.produce(Some(topic.clone())).await;
     assert!(result.is_ok());
-    assert!(result.unwrap() == 710);
-    assert!(std::path::Path::new(&format!("data/alerts/ztf/public/{}", &date_str)).exists());
+    assert_eq!(result.unwrap().unwrap(), expected_count as i64);
+
+    // Verify that the messages were actually produced:
+    let message_count = count_messages(&producer.server_url(), &topic)
+        .unwrap()
+        .unwrap();
```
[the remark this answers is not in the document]
jesaerys: This is test code, unwrapping is fine. It implicitly asserts that counting the messages didn't result in an error and that the topic exists.

src/kafka/base.rs
```diff
@@ -358,8 +480,8 @@ pub async fn consume_partitions(
     client_config
         .set("bootstrap.servers", &survey_config.consumer)
         .set("security.protocol", "SASL_PLAINTEXT")
-        .set("group.id", group_id)
-        .set("debug", "consumer,cgrp,topic,fetch");
+        .set("group.id", group_id);
+    // .set("debug", "consumer,cgrp,topic,fetch"); // Seems ignore EnvFilter directives
```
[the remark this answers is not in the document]
jesaerys: 👍

src/kafka/base.rs
```diff
@@ -154,7 +276,7 @@ pub trait AlertProducer {
             .set("acks", "1")
             .set("max.in.flight.requests.per.connection", "5")
             .set("retries", "3")
-            .set("debug", "broker,topic,msg")
+            // .set("debug", "broker,topic,msg") // Seems to ignore EnvFilter directives
```
petebachant, line 279: Want to delete this line instead of having commented code?

src/kafka/base.rs
```diff
@@ -358,8 +480,8 @@ pub async fn consume_partitions(
     client_config
         .set("bootstrap.servers", &survey_config.consumer)
         .set("security.protocol", "SASL_PLAINTEXT")
-        .set("group.id", group_id)
-        .set("debug", "consumer,cgrp,topic,fetch");
+        .set("group.id", group_id);
+    // .set("debug", "consumer,cgrp,topic,fetch"); // Seems to ignore EnvFilter directives
```
petebachant, line 484: Same here. Want to leave commented code in the codebase?

src/kafka/base.rs
```diff
@@ -129,7 +206,53 @@ pub trait AlertProducer {
...
+                    _ => Err(error),
+                })?
+            {
+                debug!(
+                    "{} avro files found in {}",
+                    avro_count,
+                    self.data_directory()
+                );
+                // If the counts match, then nothing to do, return early.
+                if total_messages == avro_count as u32 {
```
petebachant, line 230: Just want to double check that it's not possible for the directory and topic to exist, but with zero files and messages, respectively, which would cause us to exit early.

docker-compose.default.yaml
petebachant, line 1: This is a dev experience thing and outside the scope of this PR (more relevant to #184 probably), but it's kind of a pain to have to manually merge updates from `docker-compose.default.yaml` and `config.default.yaml` into `docker-compose.default.yaml` and `config.yaml` each time these change, because devs can potentially lose any changes they made if they copy/overwrite. Maybe we should simply have dev versions of these files that are committed to the repo and used directly instead of having devs maintain their own configs outside version control.

petebachant (commented):
Overall LGTM, but might be worth adding a test case for empty topic, zero files in data dir. Seems like it's possible for that to happen, but if it's too much work for too rare of a situation, `docker compose down --volumes` should clean things up for devs.

src/kafka/base.rs
jesaerys: I'd like to keep that around somehow because I had to do some digging to figure out how to get log messages from kafka. How about this instead?

```suggestion
            // To get logs from kafka, add .set("debug", "broker,topic,msg").
            // Setting an EnvFilter directive, e.g., RUST_LOG=rdkafka=debug, doesn't work.
```

src/kafka/base.rs
jesaerys: Likewise, are you ok with this?
```suggestion
    // To get logs from kafka, add .set("debug", "consumer,cgrp,topic,fetch").
    // Setting an EnvFilter directive, e.g., RUST_LOG=rdkafka=debug, doesn't work.
```

src/kafka/base.rs
jesaerys: It is possible for both to be empty, though it would be super weird. I figured the simplest logic is to skip production only if the counts match, even when the counts are zero, because it's not clear what boom should do instead.

* Maybe the downloaded tarball really is empty, and so we end up with an empty topic, which is correct. If that's not what we expect, then we should talk to the folks upstream about their tarball for that date.

* Or, maybe the directly exists but it's empty because we accidentally deleted all of the files earlier, and so we again end up with an empty topic. This would be our fault, not boom's, and we shouldn't expect boom to automatically recover from it, especially if the directory can be empty for legitimate reasons. If we know that there really should be alerts for that date, then we either restore them to the directory or we remove the directory so boom can start over.

What I can do is have boom emit a warning when the data directory is empty, just to make sure we're aware when that happens.

Thoughts?

docker-compose.default.yaml
jesaerys: 💯. I agree completely. Do the right thing by default. And if a dev needs to needs to override something, it's easy enough to tell docker and boom to use different files.

src/kafka/base.rs
jesaerys: Pete and I discussed on slack. The empty-dir case is really unlikely, so this approach should be ok. We can always change it later.

petebachant (approved):
LGTM!

PR #211, opened 2025-07-22
Finish up idempotent Kafka producer logic

petebachant:
This seems to be working fine, so I'm going to use this in the benchmarking.

### DIFF
--- a/src/kafka/base.rs
+++ b/src/kafka/base.rs
@@ -163,8 +163,40 @@ pub trait AlertProducer {
             )?;
             debug!(?total_messages);
 
-            // TODO: need to get the expected count
-            // TODO: finish the produce-or-not logic
+            // Count the number of Avro files in the data directory
+            let avro_count = count_files_in_dir(&self.data_directory(), Some(&["avro"]))?;
+
+            // If they don't match, delete all messages from the topic and
+            // start fresh
+            if total_messages != avro_count as u32 {
+                warn!(
+                    "Topic {} already exists with {} messages, but {} Avro files found in data directory",
+                    self.topic_name(),
+                    total_messages,
+                    avro_count
+                );
+                // Delete all messages from Kafka topic so we start fresh
+                let admin_client: AdminClient<DefaultClientContext> = ClientConfig::new()
+                    .set("bootstrap.servers", &self.server_url())
+                    .create()?;
+                let opts =
+                    AdminOptions::new().operation_timeout(Some(std::time::Duration::from_secs(5)));
+                admin_client
+                    .delete_topics(&[&self.topic_name()], &opts)
+                    .await
+                    .map_err(|e: KafkaError| {
+                        error!("Failed to delete topic {}: {}", self.topic_name(), e);
+                        e.into()
+                    })?;
+                info!("Deleted topic {}", self.topic_name());
+            } else {
+                info!(
+                    "Topic {} already exists with {} messages, no need to produce more",
+                    self.topic_name(),
+                    total_messages
+                );
+                return Ok(total_messages as i64);
+            }
         }
 
         match self.download_alerts_from_archive().await {
--- a/src/kafka/base.rs
+++ b/src/kafka/base.rs
@@ -175,20 +175,11 @@ pub trait AlertProducer {
                     total_messages,
                     avro_count
                 );
-                // Delete all messages from Kafka topic so we start fresh
-                let admin_client: AdminClient<DefaultClientContext> = ClientConfig::new()
-                    .set("bootstrap.servers", &self.server_url())
-                    .create()?;
-                let opts =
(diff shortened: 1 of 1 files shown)


src/kafka/base.rs
```diff
@@ -163,8 +163,31 @@ pub trait AlertProducer {
...
             debug!(?total_messages);
 
-            // TODO: need to get the expected count
-            // TODO: finish the produce-or-not logic
+            // Count the number of Avro files in the data directory
+            let avro_count = count_files_in_dir(&self.data_directory(), Some(&["avro"]))?;
+
+            // If they don't match, delete all messages from the topic and
+            // start fresh
+            if total_messages > 0 && total_messages != avro_count as u32 {
```
jesaerys, line 171:
```suggestion
            if total_messages != avro_count as u32 {
```

src/kafka/base.rs
```diff
@@ -163,8 +163,31 @@ pub trait AlertProducer {
...
+                    self.topic_name(),
+                    total_messages,
+                    avro_count
+                );
+                // TODO: Delete all messages from Kafka topic so we start fresh
+                // For now, just return an error
+                return Err(
+                    "Topic exists with different message count; manually delete the topic".into(),
+                );
+            } else if total_messages == avro_count as u32 {
```
jesaerys, line 183:
```suggestion
            } else {
```

src/kafka/base.rs
```diff
@@ -163,8 +163,31 @@ pub trait AlertProducer {
             )?;
             debug!(?total_messages);
 
-            // TODO: need to get the expected count
-            // TODO: finish the produce-or-not logic
+            // Count the number of Avro files in the data directory
+            let avro_count = count_files_in_dir(&self.data_directory(), Some(&["avro"]))?;
```
jesaerys, line 167: One thing we haven't discussed is what to do when the topic exists and the data directory does not. There would be no way to verify the message count in the topic. Do we trust what's there, or do we start over?

Whatever the case, it's safe to say we don't want propagate a "directory not found" error here.

jesaerys (commented):
Thanks for the contribution!

src/kafka/base.rs
petebachant: Good catch! This only works because I've run it in the past on the same dates--the perils of caching...

src/kafka/base.rs
petebachant: I started with this logic, but ended up wanting to handle the case where we had zero messages, which means we can go ahead and send them in. Am I missing the fact that the size of the topic can't be zero?

src/kafka/base.rs
```diff
@@ -163,8 +163,34 @@ pub trait AlertProducer {
...
+            if total_messages > 0 && total_messages != avro_count as u32 {
+                warn!(
+                    "Topic {} already exists with {} messages, but {} Avro files found in data directory",
+                    self.topic_name(),
+                    total_messages,
+                    avro_count
+                );
+                // TODO: Delete all messages from Kafka topic so we start fresh
+                // For now, just return an error
+                return Err(
```
Theodlz, line 183: Oh I didn't realize that's the bahavior you guys had in mind? I thought we wanted to skip pushing alerts when the count matches and otherwise wipe the topic and push again, but here it seems like you want to alert the user and have them clear the topic manually?

If we don't want to automate the topic deletion that's fine, but I think that providing the users with a `--clear` flag to let them delete the topic and push again using boom rather than dive into a kafka container or similar to run these commands themselves would be a LOT nicer.

petebachant: I was trying to write the code to clear the topic but I couldn't figure it out after trying for a bit (async return types this that and the other), so I landed on just raising an error message.

src/kafka/base.rs
jesaerys: I'm looking at it this way: only skip producing if the message count equals the avro count; in all other cases, delete the topic if it exists and then download+produce. To me, it shouldn't be a big deal to delete an empty topic (though maybe I'm overlooking a reason why that's a bad idea).

The fact that deleting a topic is a TODO item here shouldn't matter for us in practice. Really, having the message count not equal the avro file count is a pretty weird edge case. I can't think of any normal workflow that would regularly leave one of us with an empty topic or a topic whose message count is more/less than it should be. And when that does happen, we're already used to dropping the topic manually.

jesaerys: > Hmmmm that doesn't sound right to me. If we have more than the expected number of messages for benchmarking our runs won't be consistent, since it does take some processing time to figure out that we already have an alert to avoid reprocessing it fully.

Ah you're right. I retract my last comment that overproducing would be ok for benchmarking.

jesaerys:
@petebachant fyi I'm going to wrap this up and merge it into my branch.

jesaerys (approved):
Self-approving my changes :p

PR #213, opened 2025-07-24
WIP - Summerschool inspired bugfixes/improvements

Theodlz:


petebachant (approved):
Would recommend a docstring on one function but overall changes look minor and if it works, it works. I suppose having tests to catch the bugs fixed here would be good, but that can be added to the backlog in the spirit of getting this running.

PR #217, opened 2025-07-30
Provide evidence of filtering in logs

petebachant:
Without this, there's no real way to know if filtering was attempted. I do think it would be good to save filter results, even for alerts that don't pass the filters, perhaps inside the alert object itself, like:

```json
{
  "_id": {
    "$numberLong": "2991147293515015045"
  },
  "objectId": "ZTF18acwpwgq",
  "candidate": {
    "jd": 2460745.6472917,
    "band": "r",
    "pid": {
      "$numberLong": "2991147293515"
    }
    ...
  },
  "coordinates": {
    "radec_geojson": {
      "type": "Point",
      "coordinates": [
        -76.0545582,
        -16.742982
      ]
    },
    "l": -131.61575014218297,
    "b": -6.615562065786586
  },
  "created_at": 2460887.227974537,
  "updated_at": 2460887.227974537,
  "classifications": {
    "acai_b": 0.2063591182231903,
    "acai_h": 0.000014662742614746094,
    "acai_n": 0.0006379783153533936,
    "acai_o": 0.2871479392051697,
    "acai_v": 0.00823184847831726,
    "btsbot": 0.001475214958190918
  },
  "filters": [
    {"filter_id": 1, "filter_version": "sdk43k", "result": false}
  ]
}
```

So, we could store filter results indexed by filter ID and filter revision ID, and we could easily check if a filter was run, and if doing a historical query, quickly return results with simple boolean matching.

### DIFF
--- a/src/filter/ztf.rs
+++ b/src/filter/ztf.rs
@@ -564,18 +564,18 @@ impl FilterWorker for ZtfFilterWorker {
                 )
                 .await?;
 
-                // if the array is empty, continue
+                info!(
+                    "{} alerts passed ztf filter {} with programid {}",
+                    out_documents.len(),
+                    filter.id,
+                    programid,
+                );
+
+                // If we have output documents, we need to process them
+                // and create filter results for each document (which contain annotations)
+                // however, if the array is empty, there's nothing to do
                 if out_documents.is_empty() {
                     continue;
-                } else {
-                    // if we have output documents, we need to process them
-                    // and create filter results for each document (which contain annotations)
-                    info!(
-                        "{} alerts passed ztf filter {} with programid {}",
-                        out_documents.len(),
-                        filter.id,
-                        programid,
-                    );
                 }
 
                 let now_ts = chrono::Utc::now().timestamp_millis() as f64;


Theodlz:
I agree that logging this would be better. That said , this isnt exactly a blocker. Filter outputs go to kafka so you could just interrogate the output topic to get the offsets = number of things that passed filters (and see if you get the expected number).

Oh and actually this makes me think, but in boom-paper you should maybe clear the kafka topic that has the filter outputs, to keep things clean between runs?

jesaerys (approved):
LGTM, and doesn't seem to be controversial based on the discussion so far.

jesaerys:
> LGTM, and doesn't seem to be controversial based on the discussion so far.

Not to shut down the conversation, @petebachant @Theodlz. Please continue.

Theodlz:
Depending on how the xmatch config is set up, yes. If only the NED xmatch is specified then yes for sure if it's a list of catalogs and it ain't the first one then it should be fine.

Theodlz (approved):
LGTM, I would just change the string to:

```
"{} ztf alerts with programid {} passed filter {}",
```

which makes it more clear that the programid is that of the alerts, not the filter (which runs on a list of them, in a way)

PR #220, opened 2025-08-04
Timeseries update: fix + improvements

Theodlz:
In this PR:
- we fix the update_timeseries_op() function, where a missing `$` in front the name of the array field to update broke that function
- the $sortArray operator used by update_timeseries_op() is only supported in aggregation pipelines, so we turn the update document passed to the `update_one()` method into a vec of doc, i.e. an aggregation pipeline used for update. *Note: the performance impact of that change is unknown and will be assessed using Pete's benchmarking suite in `boom-astro/boom-paper`*.
- we handle a potential edge-case where we try to update an array with new entries, but that array does not exist yet (in which case we want to avoid an error and proceed as usual). *Note: The logic there could be improved to use a simple insert of the new values when the field is missing but that isn't much of an issue nor a priority at the moment. That slight improvement only becomes useful if we have some schema change introducing new fields.*

### DIFF
--- a/src/alert/decam.rs
+++ b/src/alert/decam.rs
@@ -227,16 +227,16 @@ impl AlertWorker for DecamAlertWorker {
         survey_matches: &Option<Document>,
         now: f64,
     ) -> Result<(), AlertError> {
-        let update_doc = doc! {
+        let update_pipeline = vec![doc! {
             "$set": {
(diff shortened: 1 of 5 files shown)


petebachant (approved):
LGTM!

PR #230, opened 2025-08-22
[DNM] Throughput test - self hosted runner

Theodlz:
Just me having fun with using my home server as a self hosted runner :)


Theodlz:
It didn't run at all and I had to start it manually on the machine, hence these numbers. Looks like producing + consuming is what failed.

PR #236, opened 2025-08-27
Quantities: first batch of changes

Theodlz:
This addresses Jake's comments on the existing PR (#234):
- [x] rename all `ml worker` and `feature worker` occurrences to the new naming schema `enrichment worker`
- [ ] add some unit tests to `lightcurves.rs`
- [ ] extract the feature computation logic in each ML workers as it's own function, to make the main processing loop shorter and easier to read.


Theodlz:
self merging, so changes can be reviewed on the other PR

PR #239, opened 2025-08-28
Instrument other workers and consumer

jesaerys:
This PR instruments the enrichment worker, the filter worker, and the kafka consumer to emit metrics.

The README now includes links to the Prometheus UI with pre-entered queries for the consumer and workers.

The metrics tracked so far give us a good sense of throughput for each stage in boom and give us insight into other things like queuing and batching behavior and the effect of adding/removing a worker.

One thing we can't get yet is a query that gives the *overall* throughput of the system as a whole. In general it works to use the min throughput as the overall throughput, but it turns out to not be straightforward to do this in PromQL. And actually it's more complicated than just taking the min because a fast stage is eventually put on hold until the other stages catch up (e.g., the consumer filling its queue faster than the alert worker can process the messages), and when this happens its throughput drops to zero. Taking the min would show the system has zero throughput, even though it's still working. It's probably possible to combine these different throughput time series into an average, but I'm guessing it will require adding a label/attribute. (It should work to combine time series with the `or` operator, but each time series would need to have a unique value of a common label, and right now we don't have that across all four alert counter meters.)

Another idea to have each alert/candid carry around a timestamp from it was first consumed. E.g., the consumer would pair each message with a timestamp value and then push that tuple onto the alert queue; the alert worker pops a tuple, processes the alert message like it currently does, but then pairs the candid with the the timestamp value it received and pushes that tuple onto the enrichment queue; etc. When the filter worker finishes processing an alert, it can record the elapsed time in a histogram, giving us a distribution of "seconds per alert", which can be flipped to get throughput.

I'll make a ticket for revisiting overall throughput.


Theodlz:
@jesaerys I suggest syncing with main, now that we have a new baseline there

src/scheduler/base.rs
```diff
@@ -178,6 +180,7 @@ impl Worker {
     /// config_path: path to the config file we are working with
     #[instrument]
     fn new(worker_type: WorkerType, survey_name: Survey, config_path: String) -> Worker {
+        let id = Uuid::new_v4();
```
Theodlz, line 183: wait is this reintroducing a uuid-based worker id?

Theodlz (commented):
Thanks for all the `ml` vs `enrichment` fixes, these evaded me... sorry about the inconvenience!

Only 1 comment and 1 question, I'll let you take a look. Otherwise this looks great. How it all works is still a little confusing to me but that's just me not being used to using static/global variables in rust. I'm not aware of how they work, and will need to look into them on my own time.

src/scheduler/base.rs
jesaerys: Yeah, there needs to be a way to distinguish workers in Prometheus. I was started by stringifying the `ThreadId` value, but that resulted in strings like "ThreadId(7)" and I thought UUID would be better. I was kind of plus or minus on making the UUID a field of the `Worker` struct: we don't have a use for it now, but I'm not sure we won't ever need it, so I named it with a leading underscore. I'm happy to remove the field if you don't want it.

Theodlz (approved):
LGTM!

PR #242, opened 2025-08-28
Throughput CI: rust build caching (in docker)

Theodlz:
In this PR, we edit the `Dockerfile-benchmark` docker file to:
- first create an empty application with the same layout as booms, and copy our Cargo.toml files to it
- then we build this empty project. This will download and build dependencies. This way, this step only runs when there are Cargo.toml changes, and otherwise is cached along with the build dependencies
- then we copy the source code of boom and build that. Since dependencies are already built, this will only require build time for the app itself. So, now that is the only build time that needs to happen when there are source code changes, and we don't need to rerun the whole dependencies update.

This is in an effort to cut down the docker build time for the throughput test. Similar changes could be made to the proper Dockerfile we use to run the app, to make builds a lot faster when we change the code, which would be a big big help.

### DIFF
--- a/Dockerfile-benchmark
+++ b/Dockerfile-benchmark
@@ -1,19 +1,23 @@
 FROM rust:1.87-slim-bookworm AS builder
 
-WORKDIR /app
-
-# Copy the Cargo.toml and Cargo.lock files
-COPY Cargo.toml Cargo.lock ./
-
 RUN apt-get update && \
     apt-get install -y curl gcc g++ libhdf5-dev perl make libsasl2-dev pkg-config && \
     apt-get autoremove && \
     apt-get clean && \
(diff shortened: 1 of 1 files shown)


Theodlz:
I tested making a dummy code change locally and this definitely works. I don't want to jinx it so once this first run finishes, I'll make a dummy code change to a binary and check how long it takes then (we take 20 minutes total currently, the hope is that this at least cuts it down to half of that even when we have code changes). Then will revert that change before merging.

Theodlz:
ok so the caching worked (docker build took 2m38s vs 10m39s), the filter worker now being measured properly adds a lot more variation to the results it looks like, so this CI run failed though :/

petebachant (approved):

PR #248, opened 2025-09-10
API: Pipeline Query endpoint

Theodlz:
This PR addresses part of #166, by adding a `post_pipeline_query` endpoint, that lets a user of the API run a MongoDB aggregation pipeline on a collection.

What this does not do is enforce some permission-level authorization on what stages can be used. Some stages can be very resource-hungry or data-access-sensitive. For example, the `$lookup` stage can performs a join and could just not be suitable/reasonable if too many documents match, and in the ZTF case - where some users should only access some data - would let users query data they might not be supposed to query. That said:
- the data access problem is true for all `queries` endpoint, and should be thought about. This is very hard to solve, we could just inject user filters/pipelines with operators to restrict data access but that may break some queries that rely heavily on indexes, so it's not that simple. It's a longer discussion item we should have. All in all, I believe that querying alert collections (not archival catalogs) should simply be banned altogether in favor of dedicated API endpoints that expose only a handful of parameters which we know we have indexes on so that performance is consistent and data access can easily be controlled. It's limiting in so many ways, but makes data access more feasible and might be a great way to avoid performance issues with users querying these gigantic tables...
- the "which stage can be used" is mostly a data access issue, and as stated above that needs to be thought of more careful. So, for now, I'd rather make things very permissive so we can test things well. We should revisit a lot of the API before letting users play with it anyways.

ALSO, this PR reorganizes the `queries.rs` file into a `queries` module where we have a file per "category" of endpoint, to keep each feature a little more self-contained and the code readable.


petebachant (approved):
I think this is fine to merge, but I have a couple non-gating comments: One around filter versus pipeline that Copilot brought up, and one is whether or not we can unit test the new filter endpoint.

Theodlz:
@petebachant ok I'm think I'm done for this PR, please take another look when you have a chance to

PR #252, opened 2025-09-10
Filter worker: compress avro output (snappy)

Theodlz:
This PR adds the `snappy` feature to apache-avro crate, so we can use the snappy codec when making the avro packets that the filter worker outputs to kafka. The avro packets we put out with the alert worker contain images and lots of metadata, so adding compression on top is very very appreciated. This means less data volume sent to Kafka and fetched by clients, helping with networking.

Not: LSST alert cutouts actually aren't compressed at all, unlike ZTF's. Instead, the avro packets themselves are compressed! So before that change was made that meant we were sending uncompressed images to Kafka which is clearly suboptimal. MongoDB uses snappy/gzip to compress records, so the images we send there should be compressed anyways and we are fine. But, for consistency sake we may want to consider compressing incoming LSST images before saving them to Mongo (that could let us use a better-suited compression algorithm, and remove some survey-specific image processing logic we would need in all workers that come after the alert one)). Something to think about!


petebachant (approved):
Sounds like a good idea to me!

PR #253, opened 2025-09-10
LSST alert: Missing IDs deserialization bugfix

Theodlz:
fix issue with lsst alerts giving us ids=0 (diaObjectId and ssObjectId, only one of these should be defined and the other None) instead of None. So, we simply deserialize 0 values as Nones when reading these ids.

For context, this created a problem because only one of diaObjectId or ssObjectId should be defined values, but since the missing one was 0 we ran into our custom `AmbiguousIds` issue.

So in this PR, we remove the distinct deserialize objid vs ssoid functions to combine them into one function that can be used for both, and shorten it a bit.

### DIFF
--- a/src/alert/lsst.rs
+++ b/src/alert/lsst.rs
@@ -181,6 +181,7 @@ pub struct DiaSource {
     /// Filter band this source was observed with.
     pub band: Option<String>,
     /// Source well fit by a dipole.
+    #[serde(rename = "isDipole")]
     pub is_dipole: Option<bool>,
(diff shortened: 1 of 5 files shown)


petebachant (approved):
Seems reasonable to me. Is it worth adding one of these problematic alert cases to a unit test though?

Theodlz:
> Seems reasonable to me. Is it worth adding one of these problematic alert cases to a unit test though?

Hmmm I thought about it, but unclear how to easily simulate that without extending the functionality of the alert randomizer. All in all, this was a proper bug on the LSST end, and I think there's so much we can do about that. Today it's these ids, tomorrow it could be something else. It would just be a never ending loop of us writing schema validation unit tests that quite frankly, should exist on their end.

PR #254, opened 2025-09-10
ZTF & LSST forced photometry: bugfixes

Theodlz:
In this PR, we:
- ztf and lsst alert workers: compute magnitude even for negative forced photometry fluxes
- ztf filter worker: fix fp zeropoint
- lsst filter worker: add missing flux unit conversion (so that `Alert` `Photometry` values from different surveys use the same flux units).

Note: This flux unit conversion does make me wonder: should we make sure at deserialization time that fluxes are converted to the same unit for ALL surveys? This way, in the DB itself units are matched and can be used in queries. I really doubt folks would use fluxes in queries but you never know, and folks using LSST data might assume the flux is in the unit of the original alert packet and some other folks might assume it's in another unit (like ZTF's flux for example). So... I honestly don't know what's best. In my opinion consistency is best in a data model, so we should convert things and simply document things properly OR rename the field names (just like we replace LSST's mjd values to jd to match ZTF, which is another debate in itself as perhaps we should do the opposite but whatever) so that the unit is clearly mentioned in them... Something to think about!

### DIFF
--- a/src/alert/lsst.rs
+++ b/src/alert/lsst.rs
@@ -529,15 +529,14 @@ impl TryFrom<DiaForcedSource> for ForcedPhot {
         // may revisit this later
         let (magpsf, sigmapsf, isdiffpos, snr) = match dia_forced_source.psf_flux {
             Some(psf_flux) => {
-                let psf_flux = psf_flux * 1e-9;
-                if (psf_flux / psf_flux_err) > SNT {
-                    let isdiffpos = psf_flux > 0.0;
-                    let (magpsf, sigmapsf) = flux2mag(psf_flux, psf_flux_err, ZP_AB);
+                let psf_flux_abs = psf_flux.abs() * 1e-9;
+                if (psf_flux_abs / psf_flux_err) > SNT {
+                    let (magpsf, sigmapsf) = flux2mag(psf_flux_abs, psf_flux_err, ZP_AB);
                     (
                         Some(magpsf),
                         Some(sigmapsf),
-                        Some(isdiffpos),
-                        Some(psf_flux / psf_flux_err),
+                        Some(psf_flux > 0.0),
+                        Some(psf_flux_abs / psf_flux_err),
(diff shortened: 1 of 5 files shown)


petebachant:
> Note: This flux unit conversion does make me wonder: should we make sure at deserialization time that fluxes are converted to the same unit for ALL surveys? This way, in the DB itself units are matched and can be used in queries. I really doubt folks would use fluxes in queries but you never know, and folks using LSST data might assume the flux is in the unit of the original alert packet and some other folks might assume it's in another unit (like ZTF's flux for example). So... I honestly don't know what's best. In my opinion consistency is best in a data model, so we should convert things and simply document things properly OR rename the field names (just like we replace LSST's mjd values to jd to match ZTF, which is another debate in itself as perhaps we should do the opposite but whatever) so that the unit is clearly mentioned in them... Something to think about!

In my understanding BOOM's job is to enrich (add to or extend) the alerts and filter them, not necessarily transform them. However, uniformity between different streams sounds smart if they are consumed and viewed together. Is there any way to signal to consumers what the output units are, perhaps by providing a schema definition to the community for all output streams?

Do they feel like they are consuming the LSST stream, or BOOM's stream, which includes LSST data? Maybe focus on SkyPortal's use case since that's the most definite and urgent.

Theodlz:
@petebachant would you mind reviewing this with the new changes?

Theodlz:
Oh and to add to the above: I absolutely think we want to have a place somewhere in the docs where we clearly display the schema of the data we have in the DB. That shouldn't be too hard to implement since we already have structs that define the schema of the upstream survey. All we do in BOOM is add new fields and split the alert in 3 collections. What we are missing are structs that define the exact DB schema (because when we do insert in the alert worker, we create documents directly from our structs, rather than structs that we let the driver serialize to documents). Maybe that's the thing we should change, so we have somewhere structs that clearly define the data model.

Issue #208, opened 2025-07-17
Consuming from Kafka: dev vs prod


Theodlz:
At the moment, the consuming is somewhat identical between surveys. Topic names and auth may vary, but thereafter its all the same, we:
- assign consumer(s) to partitions of a topic
- query for the offsets that give us records pushed after a certain date/timestamp (always the 00:00:00 UTC, which happens before the start of the night for ZTF).

However, a number of things are incorrect here, and it all boils down to the fundamental difference between LSST and ZTF/DECAM:
- LSST has ONE TOPIC where everything goes, so if we want to read data pushed after a certain date, we need to query for the offsets matching with that date.
- ZTF/DECAM (and all other surveys I've worked with so far, for Kowalski & Boom) have nightly topics, as in they have one topic per UTC date. So, we don't need to query for the offsets at all.

Moreover, right now everything has been designed for development. By that, I mean that we set our consumers to new offsets (to match the date) everytime, when really we want to do such a thing only while developing (or in rare occasions in production, where we want to force consuming from the beginning of a night or re-consume a past night of data). So, instead we should:
- for development and all surveys, keep the current behavior (minus querying for offsets for surveys other than LSST where it's not needed)
- for production in the LSST case, only query for the offsets that match a date ONCE in the LSST case and thereafter subscribe again and keep reading where we  stopped consuming earlier.
- for production in the ZTF case,  never query offsets but just keep the same group.id and such so we just resume where we left off.

So all in all, I think we need to do one of (or both):
- have an optional group.id be set in the kafka configuration (in `config.yaml`) for each surveys. When it's None, we assume we are in dev mode and do what is needed to restart reading from the very beginning of whatever is the provided date.
- have some `--resume` argument passed to kafka_consumer binary to tell it if it should keep reading from where kafka tells it to (for the provided group.id if any) or if it should restart from the first alert of the provided date.

PS: I feel like we should really have the group.id in the config and not in the code, as changing it in the codebase could have pretty bad consequences in production systems that use it. So keeping track of it on their end makes more sense to me.

jesaerys:
I'm happy to look into it.

PR #255, opened 2025-09-10
Kafka Consumer: partition checks missing auth fix

Theodlz:
In this PR, we:
- use kafka credentials for consumer partition checks, where needed (e.g. for LSST).
- add a const to define the timeout to use when talking to kafka, so we don't define a duration again and again and keep things consistent
- for lsst: make the simulated flag (which picks which topic to read from for LSST, simulated alerts for testing or the real topic) accessible from the CLI, defaulting to false.

### DIFF
--- a/src/bin/kafka_consumer.rs
+++ b/src/bin/kafka_consumer.rs
@@ -47,6 +47,10 @@ struct Cli {
     #[arg(long, value_name = "MAX", default_value_t = 15000)]
     max_in_queue: usize,
 
+    /// Simulated mode (for testing purposes, LSST only)
+    #[arg(long, default_value_t = false)]
+    simulated: bool,
+
     /// UUID associated with this instance of the consumer, generated
     /// automatically if not provided
     #[arg(long, env = "BOOM_CONSUMER_INSTANCE_ID")]
@@ -74,12 +78,6 @@ async fn run(args: Cli, meter_provider: SdkMeterProvider) {
     .and_utc()
     .timestamp();
 
-    // TODO: let the user specify if they want to consume real or simulated LSST data
-    let simulated = match args.survey {
-        Survey::Lsst => true,
-        _ => false,
-    };
-
     match args.survey {
         Survey::Ztf => {
             let consumer = ZtfAlertConsumer::new(None, Some(args.program_id));
@@ -103,7 +101,7 @@ async fn run(args: Cli, meter_provider: SdkMeterProvider) {
             };
         }
         Survey::Lsst => {
-            let consumer = LsstAlertConsumer::new(None, simulated);
+            let consumer = LsstAlertConsumer::new(None, args.simulated);
             if args.clear {
                 let _ = consumer.clear_output_queue(&args.config);
             }
--- a/src/kafka/lsst.rs
+++ b/src/kafka/lsst.rs
@@ -40,7 +40,7 @@ impl AlertConsumer for LsstAlertConsumer {
         if self.simulated {
             "alerts-simulated".to_string()
         } else {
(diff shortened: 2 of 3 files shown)


jesaerys (approved):
These changes look good, but I'm wondering about how we use "simulated" for the consumer. Does it make more sense to treat LSST the same as ZTF, in that it's up to the producer to publish alerts to a local topic (either simulated alerts or real alerts downloaded from an archive)? This way, the consumer works the same in dev and in prod, the only difference is the bootstrap server and auth.

jesaerys:
By the way, my preference is to merge this one in before the stuff I'm working on for #208. I'm making some fairly sweeping changes and it's easier for me to incorporate this than the other way around.

src/bin/kafka_consumer.rs
```diff
@@ -47,6 +47,10 @@ struct Cli {
     #[arg(long, value_name = "MAX", default_value_t = 15000)]
     max_in_queue: usize,
 
+    /// Simulated mode (for testing purposes, LSST only)
+    #[arg(long, default_value_t = false)]
+    simulated: bool,
+
```
petebachant, line 53: Nit: Consider `BOOM_SIMULATE_LSST` env var or something for config in lieu of a command line arg.

src/kafka/lsst.rs
```diff
@@ -40,7 +40,7 @@ impl AlertConsumer for LsstAlertConsumer {
         if self.simulated {
             "alerts-simulated".to_string()
         } else {
-            "alerts".to_string()
+            "lsst-alerts-v9.0".to_string()
```
petebachant, line 43: Maybe worth making configurable?

petebachant (approved):
LGTM!

Theodlz:
> By the way, my preference is to merge this one in before the stuff I'm working on for #208. I'm making some fairly sweeping changes and it's easier for me to incorporate this than the other way around.

Ah cool!

src/kafka/lsst.rs
jesaerys: Yeah. My new consumer CLI has a subcommand for each survey (`decam`, `lsst`, `ztf`), and each subcommand has whatever args it needs to get the topic name. For LSST, there's an arg for the schema version. For ZTF and DECam, there's an arg for the date.

src/bin/kafka_consumer.rs
petebachant: I agree. It's a minor thing--the logic being that an uncommon "dev" option is better slightly hidden, but it's only ever going to be run by devs or an automated process.

Theodlz:
> These changes look good, but I'm wondering about how we use "simulated" for the consumer. Does it make more sense to treat LSST the same as ZTF, in that it's up to the producer to publish alerts to a local topic (either simulated alerts or real alerts downloaded from an archive)? This way, the consumer works the same in dev and in prod, the only difference is the bootstrap server and auth.

I'd love to say yes, but we just can't do it like ZTF. We do not have a set of publicly accessible LSST alerts we can pull from to produce locally. We could host our own (some sample of real data), but data can't be advertised or shared with the public until late Oct (meaning I should not even talk about this here on git, and we can't add a feature that indicates that we are pulling some LSST data from somewhere)

Theodlz:
And it's just not like ZTF, LSST has some kafka topic for us with simulated alerts. That will be how we can get some fake alerts in future schema versions in the future before they promote them to prod.

jesaerys:
I'm only proposing that we move the "simulated" logic to the producer so that the consumer doesn't have to be aware of such details. That should be possible, no?

PR #261, opened 2025-09-29
Add back `classtar` to the ZTF candidate schema

Theodlz:
This PR adds the `classtar` field back to the ZTF candidate schema, also adds a missing `serde rename` macro on the `is_dipole` field (which is named `isDipole` in the schema).


mcoughlin (approved):
LGTM

Issue #159, opened 2025-06-10
Make error logging more ergonomic


jesaerys:
This is kind of a followup to #105. PRs #125 and #136 brought some nice improvements, but to me there's still room to do better. The ideal I have in mind is to be able to write code without having to constantly use `.map_err` or `.inspect_err` etc. on every single `Result` value, and that errors show up in the logs with clear provenance (including line numbers!) and enough context to understand the state of the system when the error occurred. I think we're almost there, there are just a couple issues I'd like to address:

1. My current solution to getting line numbers *requires* creating a dedicated event callsite (i.e., a place where `tracing::{error,info,warn,...}!` is used). This is why we see things like `.inspect_err(as_error("oh nos"))` in boom currently. It was meant to be a rare workaround to disambiguate similar errors from the same function/span, but it's turned out that we need it quite frequently and it's really no better than how we used to write `.map_err(FooError::Bad)` everywhere.

2. Errors are logged not just where they're handled, but also everywhere they're propagated. This means the log shows an ERROR event for each source error in an error's cause chain. @Theodlz correctly pointed out that this is a bit disconcerting, and though part of me sees it as "tracing in action", overall I've come to think that there must be a better way.

I've been looking into some refinements and what I can say so far is that adopting an "accumulator" error type like `anyhow::Error` or `eyre::Report` could be really useful. Both provide good error "reports" that include nearly all of the information I'm looking for. This comes at the expense of ERROR lines being less structured, but that's probably ok. I think the `tracing_error` crate will also play a role (indirectly in the case of `eyre`), which would allow us to stop auto logging ERROR events with the `instrument` attribute.

In the end, it may not be possible to get rid of the `map_err` pattern completely, but I can live with that.

PR #262, opened 2025-09-29
new consumer app

### LINKED ISSUE #208: Consuming from Kafka: dev vs prod

jesaerys:
Closes #208 

This is a rewrite of `kafka_consumer` that aims to address #208. Here are the highlights:

* The binary belongs to a new, independent crate named `consumer`. This takes advantage of the workspace structure established by `app`, promoting separation of concerns and clear dependencies. To do anything with this crate in `cargo`, you'll need the `-p` flag, e.g., `cargo build -p consumer`, `cargo test -p consumer`, etc.

* The CLI is a bit different from `kafka_consumer`. In summary,

  * Each survey is considered a separate "source" with its own survey-specific options in a dedicated subcommand. Running `consumer --help` should assist in discovering the new system. I'm used to running commands like the following during development:

  ```sh
  kafka_consumer --processes 2 ztf 20250205
  ```

  That's basically the same as running this with `consumer`:
  
  ```sh
  consumer --consumers 2 --from beginning ztf --date 20250205
  ```

  Here, `--date` is an option for the `ztf` subcommand that plays a role in determining which topic to read from. Another relevant option is `--program-id`. You can look at the `--help` for the subcommands to learn more.

  * The `--from` option tells the consumer tasks where to start. The default is "stored", telling the consumer group to pick up where it left off. Other choices are "beginning", "end", and timestamp-ish values prefixed with "@" ("@20250205", "@2025-02-05", "@2025-02-05T00:00:00Z"). For ZTF and Decam, it sounds like we'll always want to use `--from beginning`, where for LSST we'll use `--from @{whatever-date}`.

* As a separate crate, you'll find that `consumer` effectively treats `boom` as a library for some common stuff like logging, metrics, and config. Integrating this into the existing config system wasn't entirely straightforward, but I think that's because the config system could use some love. Overall, there's still a single config file and `consumer` only uses the parts it cares about. `consumer` also uses a layered config system in which environment variables can override whatever is in the "consumer" section of the config file. This allows for deployment-specific overrides as well as injecting things like LSST credentials. I've documented the new stuff in `config.yaml`.

* This PR is also begins to address #159. After playing around with some different error-handling stuff, I think `eyre`/`color-eyre` is probably the best bet for this project. My main goal is to reduce boilerplate and ceremony wherever possible, so I've had `anyhow` in mind for a while:

  * Its "backtrace" feature allows for tracing through each line that propagates an error with `?` back to where it originated, which seems like exactly what we want. However, this requires opting-in by setting `RUST_BACKTRACE=1`/`RUST_LIB_BACKTRACE=1`. More annoying is that the resulting backtrace report is often extremely noisy, so you kind of have to learn how to read it.

  * Without the backtrace, the `{:?}` error representation includes the cause chain, if there is one. But in practice this is only useful when every `?` propagation site (or at least many of them) have context attached. This basically forces us to annotate every `?`, which is what I'm trying to get away from. Also the cause chain does *not* include line numbers, so debugging means grepping for context messages.

  And then I looked into `eyre`, specifically the extension crate `color-eyre`. It does all the same stuff as `anyhow`, but the error reporting is a little nicer:
  
  * Includes the location of origin of the error by default. This by itself is *extremely* useful.

  * With `RUST_BACKTRACE=1`/`RUST_LIB_BACKTRACE=1` and `color_eyre::install()`,

    * Frames may be omitted from the backtrace depending on the situation, which can make it easier to find the relevant information.
    * The backtrace has some coloring applied to help make it easier to read.

  Consider the code in this PR as a demonstration of what error handling could look like. Notice there are very few places where `#[instrument]` is used; that functionality is now focused on creating spans to help make the logs easier to read, it's no longer also trying to be a traceback mechanism. And there are a lot fewer uses of `map_err`, `inspect_err` and things like that.

* I've included tests for the critical functionality around consuming for different offsets. I'm really not sure how it will work in CI, hopefully it works out.

* I've been running it locally as a replacement for `kafka_consumer` and so far so good. I can get metrics just like before, and I'm not noticing any change in performance for better or worse.

* Last but not least, this PR demonstrates one way to achieve modularity in terms of surveys. To add a new survey, one just needs to define a new type that implements the `Source` trait as well as ancillary types for args and config. See `consumer/src/sources` for examples.


Theodlz:
A quick comment that came to mind as soon as I tried running it: it would be good to have some level of feedback about the parameters that are used for consuming. I explain: I just ran `consumer ztf` and it started reading from kafka. But I don't know from what "kind" of offset, i don't know what topic, ... Even just for logging purposes, it would be great to have that. Now perhaps where there is indeed data flowing that gets logged at some point, but when you just init a consumer (that may not have alerts to read yet, which is likely to be how it will look like in production when we start a kafka process and wait for data to arrive) you don't really have some feedback/confirmation that the parameters used are those you gave, or what it defaulting to.

config.yaml
```diff
@@ -25,6 +25,73 @@ kafka:
...
+    # Username for the LSST Kafka cluster. Do *not* include it here!
+    # Use BOOM__CONSUMER__LSST__USERNAME instead.
+    username: ""
+
+    # Password for the LSST Kafka cluster. Do *not* include it here!
+    # Use BOOM__CONSUMER__LSST__PASSWORD instead.
+    password: ""
+
+  ztf:
+    bootstrap_servers: "localhost:9092"
```
petebachant, line 93: Maybe nest under `surveys`? It's possible we someday have a survey that conflicts with another section of the config, though maybe unlikely.

petebachant (commented):
Why the switch from `--processes` to `--consumers`? I could imagine other apps having a similar argument, e.g., the enrichment worker. `-n` could be a good shorthand flag for number of processes as well.

Regarding project organization, I've seen other Rust projects put crates inside a subdirectory, like:

```
boom/
    crates/
        boom-api/
        boom-consumer/
    src/
```

Would that make sense here? Our executable would then be `boom-consumer`.

petebachant:
> @petebachant I renamed `--processes` to `--consumers` because strictly speaking we're not dealing with processes, but rather async tasks that are probably executed on separate threads (but not necessarily).

Makes sense. Do we want to allow the number of concurrent consumer tasks to be set in the config like the workers? I don't have a strong feeling either way because it will be committed to the repo either there or the Docker Compose config via the service's command.

> As for crate organization, it's up to you guys at this point. IMO each crate getting a top-level directory makes sense in a workspace, but it's all personal preference.

Looking around at [trending Rust projects](https://github.com/trending/rust) to see what is the most common, I 
see a mix of [all crates as top-level](https://github.com/category-labs/monad-bft) and [crates in their own folder](https://github.com/typst/typst).

One pattern that seems to be consistent is to name the crates with a common prefix, which in our case would be `boom-`, so we'd have `boom-api` and `boom-consumer` (maybe even `boom` for the main lib).

This is also something that can easily be dealt with later on since I don't think our number of crates is going to multiply that quickly.

PR #264, opened 2025-10-04
cider images and metadata model added

frenbox:
@Theodlz Hi Theo,

Please take a look and suggest edits onto this.


Theodlz (changes requested):
First batch of comments @frenbox, but it's nice to see how adding a new model doesn't seem to require too much extra code.

The properties cloning thing to avoid could be tricky. If you can't figure it out let me know and I'll take a crack at it.

Theodlz (commented):
Looks a lot better sushant! For consistency sake, now that cider's `get_metadata` takes the alert_properties as a `&[&Document]` (rather than a `&[Document]`), can you make the same change to BTSbot's `get_metadata` method?

Other than that, I'm happy with the current state of the PR. I'll make some testing on my own time. A more general question I had about this image + metadata model ala BTSbot: why are the classes nuclear vs sn1 vs sn2 vs cv? I understand this PR is a proof of concept for implementing new models in boom but is this model something we actually want to run in production? I guess I'm surprised that this tries to distinguish between SN1 and SN2, which feels like a lightcurve-based model task. Does that just inherit from the lack of model fusion from previous iterations of cider, where the image and lightcurve models needed to have the same classes to be merged together? I know it's not related to that PR directly, but shouldn't the image model only be phenomenological ala ACAI?

frenbox:
> Looks a lot better sushant! 

Thanks Theo!

>For consistency sake, now that cider's `get_metadata` takes the alert_properties as a `&[&Document]` (rather than a `&[Document]`), can you make the same change to BTSbot's `get_metadata` method?

I think this is done in the PR. I changed the BTSBot `get_metadata` function to get alert_properties as `&[&Document]` 

> A more general question I had about this image + metadata model ala BTSbot: why are the classes nuclear vs sn1 vs sn2 vs cv? I understand this PR is a proof of concept for implementing new models in boom but is this model something we actually want to run in production? I guess I'm surprised that this tries to distinguish between SN1 and SN2, which feels like a lightcurve-based model task. Does that just inherit from the lack of model fusion from previous iterations of cider, where the image and lightcurve models needed to have the same classes to be merged together? I know it's not related to that PR directly, but shouldn't the image model only be phenomenological ala ACAI?

Metadata features provide some help for classification (although the cross contamination remains high between the two SN classes), and more broadly, we are very quickly pushing towards including the light curves in the fusion model which of course will combine this and the photometry model to produce the final results.

src/enrichment/models/cider.rs
```diff
@@ -150,4 +128,270 @@ impl CiderImagesModel {
...
+        &self,
+        photometry: Vec<PhotometryMag>,
+    ) -> Result<
+        (
+            ArrayBase<OwnedRepr<f32>, Dim<[usize; 3]>>,
+            ArrayBase<OwnedRepr<bool>, Dim<[usize; 2]>>,
+        ),
+        ModelError,
+    > {
+        let mut sorted_photometry = Vec::from(photometry);
```
Theodlz: It's already sorted (or at least should be?). Please double check, sorts can be expensive so we don't want to sort something that's already sorted. I wonder if it's not already deduplicated too.

src/enrichment/models/cider.rs
```diff
@@ -150,4 +128,270 @@ impl CiderImagesModel {
...
+        for point in sorted_photometry {
+            bands
+                .entry(point.band.clone())
+                .or_insert_with(Vec::new)
+                .push(point);
+        }
+        // Process each band separately
+        for (_band, mut points) in bands {
+            // Sort by time
+            points.sort_by(|a, b| a.time.partial_cmp(&b.time).unwrap());
```
Theodlz: you already sorted by time earlier, dividing by band doesn't "cancel" the sort. So, you shouldn't need to sort again.

src/enrichment/models/cider.rs
```diff
@@ -150,4 +128,270 @@ impl CiderImagesModel {
...
+                tail3.push(0.0);
+            } else if point.band == "r" {
+                tail1.push(0.0);
+                tail2.push(1.0);
+                tail3.push(0.0);
+            } else {
+                tail1.push(0.0);
+                tail2.push(0.0);
+                tail3.push(1.0);
+            }
```
Theodlz: maybe it would be a lot more efficient to create tail1, tail2, tail3 as arrays of the correct size to begin with (so maybe they don't have to be vecs anymore, just regular arrays) filled with 0s, and add 1s where needed? That way you just write to one already pre-allocated array at a time, instead of allocating more and more memory to 3 arrays

src/enrichment/models/cider.rs
```diff
@@ -150,4 +128,270 @@ impl CiderImagesModel {
...
+        let mut tail3: Vec<f32> = Vec::new();
+
+        for point in &result {
+            let diff_delta_t = point.time - initial_t;
+            let diff_dt = point.time - temp;
+            temp = point.time;
+            dt.push(diff_dt as f32);
+            delta_t.push(diff_delta_t as f32);
+            log_fluxes.push(point.mag.max(1e-6).log10() as f32);
+            sigma_fluxes.push(point.mag_err / (point.mag * 2.302585) as f32);
```
Theodlz: all of these normalizing factors should be documented somewhere. Would be good to have a document (if not here, in a cider repo) that lists them, to serve as a source.

frenbox:
This PR is no longer valid. Closing it.

PR #269, opened 2025-10-20
Kafka Consumers: rely on the group coordinator for consumer partition assignment

Theodlz:
First, some info about how we currently do stuff on main, to consume alerts from say LSST:
1. We query the server for metadata about the existing topics, to check if the topic we want to read from exists, and more importantly how many partitions it has
2. Given the number of `processes` specified when running `kafka_consumer`, we attribute a partition id to each process.
3. We start the consumer processes, and each of them uses the provided `timestamp` to manually assign partitions (by their ID) to themselves, at offsets matching the timestamps. This let's us have each consumer start reading partition messages that were emitted at or after a given start date, useful especially in LSST land where we have one gigantic topic with all nights of data, unlike ZTF where we have one topic per night.

This worked just fine, but the following happened:
- we see a bunch of kafka-related errors, telling us we failed to authenticate to the group coordinator. The group coordinator is a process from the kafka cluster (so the distant server we are reading data from) that, when multiple consumer processes are subscribing to a topic, gives them orders to re-balance "who read which partitions". This way if you have multiple consumers with the same `group.id` the server is responsible of telling which clients what partitions they should read from or not. Effectively, this means the server assignes the partitions to the client processes. But so far we didn't use that at all, since we manually assigned partitions on the client side. One big drawback of client-side manual partition assignment: if one of our processes dies, the others get no order to pick up the partitions from that dead consumer.
- at some point, for a few days we literally couldn't read data from LSST's kafka cluster! No idea why, but with a regular subscribe approach it worked, but the manual assignment didn't
- because we didn't use the group coordinator, LSST folks mentioned that they couldn't see us at all in their logs! Definitely an issue for observability.

Because of these various issues all coming from the manual partition assignment we are doing, after discussing with the folks handling the LSST stream I decided to implement the following, which BTW python has built in:
1. We spawn our `N` processes, which simply subscribe to the topic. After a few seconds, the kafka cluster will tell our consumers which partitions they are assigned to.
2. We await for the first message. When we get it, we call a new method where for each partition that was auto-assigned to our consumer, we seek the offset matching our timestamp. When that happens, the server will "revoke" the currently assigned partitions and will tell the consumers what new partitions they have been assigned (because we didn't remove/add new consumers between that offset seeking and now, in theory "partitons per consumer" should be identical, just the offset from which we read changes
3. We now enter our main loop, and receive the data we are owed, starting from the provided offset.

TODOs:
- [x] This works great. No more kafka errors, and LSST folks can see us in their logs. However since the server now tells the consumers of a given `group.id` which partitions they get assigned to, we run into another issue: `group.id` collisions. Right now we auto-generate a `group.id` in the code, so everyone running a kafka_consumer ends up in the same group. This means if we have more than 1 person running a `kafka_consumer`, the server will assign the partitions between these 2 people and none of them will get the full stream. So, from now on, we **require folks to set a `group.id` of their choice, which should remain private, in their .env file which is then read along with the config**.
- [x] LSST needs a username and password. We pass those as env variables which is great, but they live completely separately from the "config with fields overwritten by env variables" framework we now have on main. That should be fixed.
- [x] Right now our Kafka consumer config key is just a string value. Let's instead make it a dict with keys to specify the server, group_id, username and password (last 3 are set via config variables).
- [x] I found a capitalization-based issue with reading the config. values loaded from YAML keep their capitalization, but those read from env variables do not. To fix this, I updated all configs to stick with lowercase values, and updated the code where needed to lowercase survey names when reading from the config.


src/bin/test_xmatch.rs
petebachant: Is this intended to be its own app or should it be a separate test, maybe excluded from the default `cargo test`, but still executed with it?

Theodlz: Should be discussed somewhere else, but I'd love having the current repo broken into these crates:
- boom-core: contains the core features that all crates need to run boom, such as reading configs, interacting with the db, interacting with redis.
- boom-alert, containing the contain of the `src/alert` dir. It could have one binary to start a single alert worker, useful if we want to handle the scaling with docker/k8 rather than the scheduler babysitting all processes
- boom-enrichment, same as above
- boom-filter, same as above
- boom-scheduler, containing the current scheduler code
- boom-kafka, containing the kafka consuming code, it could also contain the producing code we have for local dev (so this is similar to Jake's currently opened PR, which unfortunately is conflicting with the other PR I have open with bugfixes to kafka reading, so I'll have to reconcile these)
- boom-api, containing the api code. that one will likely read from other crates such as boom-filter, to get the filter struct and methods to validate a filter and such, so we can finally deduplicate that.

Any thoughts? Main goal is to avoid duplication which could also be achieved by a mono crate like design, but there's already a lot to boom and separating different tasks seems the right path to me. I already find myself struggling to find things with the current structure.

PR #274, opened 2025-10-25
Throughput CI: Refine Wall time as first Kafka msg

Theodlz:
Between the start of any kafka consumer process (whether it's boom or not), there are initial steps to go through before we receive any message. Because most of these involve talking to the kafka server, the consumer itself has nothing to do with how long that takes. And in general, if what one wants to measure is the speed at which messages are flowing and thereafter being processed, accounting for that initialization time isn't relevant, and actually prevents you from measuring the actual time that your own application takes to process the data.

This PR redefines the wall time start of the throughput CI as the first message received from kafka. For that to work, I added a simple log message that occurs when the first message comes in, that way we can measure that precisely.

### DIFF
--- a/tests/throughput/_run.sh
+++ b/tests/throughput/_run.sh
@@ -29,7 +29,7 @@ EXPECTED_ALERTS=29142
 N_FILTERS=25
 TIMEOUT_SECS=300 # 5 minutes
 
-# Wait for the kafka consumer to start expecting messages (when it logs "Starting Kafka consumer loop...")
+# Wait for the kafka consumer to start expecting messages (when it logs "Consumer received first message, continuing...")
 echo "$(current_datetime) - Waiting for Kafka consumer to start"
 START_TIME=$(date +%s)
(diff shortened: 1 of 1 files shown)


Theodlz:
> Between the start of any kafka consumer process (whether it's boom or not), there are initial steps to go through before we receive any message. Because most of these involve talking to the kafka server, the consumer itself has nothing to do with how long that takes. And in general, if what one wants to measure is the speed at which messages are flowing and thereafter being processed, accounting for that initialization time isn't relevant, and actually prevents you from measuring the actual time that your own application takes to process the data.
> 
> This PR redefines the wall time start of the throughput CI as the first message received from kafka. For that to work, I added a simple log message that occurs when the first message comes in, that way we can measure that precisely.

Note: I'm not expecting this to necessarily report a lower wall time. The initialization time from the consumer can take no more than a second, like it could take 45 seconds (that's what I experienced locally). So, this is all about being a bit more accurate and avoiding some unwanted variance in the reported results. So that all that can vary, is the actual thing we can about measuring.

PR #278, opened 2025-10-28
Remove database host from deploy.yaml

### LINKED ISSUE #277: Database host does not need to be mapped in `deploy.yml`
We thought this was an issue with the deployment, since Mongo wasn't allowing the correct auth, but this was not a problem.

petebachant:
Resolves #277

This was not the issue preventing Mongo from initializing properly.


Theodlz (approved):
LGTM, thanks Pete

PR #282, opened 2025-11-05
Filter module & API deduplication

Theodlz:
* define the Filter struct in boom::filter::base.rs along with a LoadedFilter struct and methods to load Filter from the DB and retrieve the active filter version.
* For each survey, add a method to create a LoadedFilter from a filter's id
* re-use newly defined structs and methods in the API filter routes and unit tests

TODOs:
- [x] more deduplication in the filter loading process can be done, to shorten things a bit and avoid having duplicated logic between surveys.
- [x] look at the filter API endpoints, where testing a filter is somewhat reimplemented. Maybe that can be simplified and re-use some of the things added to the filter module in this PR

**THIS PR INCLUDES BREAKING CHANGES:**
Filters used to have a `catalog` field, with values like `ZTF_alerts`, referring to the name of the catalog on which they run. This feels quite odd, as we may customize/edit catalog names in code at any time, and pretty much everywhere else we rely on Survey (as a string or enum, more often the latter) which is a lot more robust. So, I decided to replace that field by a `survey` field, of value `Survey` (enum), indicating which survey's alerts will a filter run on. I updated the CI/CD (throughput pipeline) based on these changes. Various deployments will have to update their existing filter schemas. Given that we are still in dev mode, I do not see a need to create a migration script to serve in the codebase.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,9 +207,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
+checksum = "ddd31a130427c27518df266943a5308ed92d4b226cc639f5a8f1002816174301"
 dependencies = [
  "memchr",
 ]
@@ -502,6 +502,7 @@ dependencies = [
  "rdkafka",
  "redis",
  "reqwest",
+ "schemars 1.1.0",
  "serde",
  "serde_json",
  "serde_with",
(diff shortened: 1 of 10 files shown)


Theodlz:
> LGTM! I like the change from catalog to survey since catalog is more of an internal detail so users will find survey more intuitive.

Yep, 100%. This week I will open another PR which will change the permissions stuff to support the multi survey effort (so a filter which runs on alerts coming from survey A, can express its permissions for the other survey's data it may use when there is a crossmatch. For example, right now if you were to run a filter on LSST and use ZTF data there, we don't have a way to say that this filter may only use Public+Partnership data and not Caltech). So this permissions variable will become a document with survey names as keys and arrays of permissions as values, where applicable (all LSST data is public so we don't care about permissions much there).

PR #285, opened 2025-11-10
API: Test filter pipeline endpoint + filter module documentation

Theodlz:
Adds an API endpoint to test a filter pipeline on a batch of alerts (needs to be constrained on at least one of: candidate.jd range, objectIds, candids), provided permissions and which survey it should run on. 

A future version of that endpoint would rather take an existing filter ID as input, and use these permissions and survey. That would be a way to enforce data access. That said for now our API basically exposes everything to everyone so data access isn't this PR's problem.

This endpoint is needed for Thomas' filter building UI, where real time testing of filters is absolutely needed.

NOTE: We also add some guardrails here, so systems that use this endpoint don't abuse of it:
- no more than 1k objectIds at a time
- no more than 100k candids at a time
- no more than 1 JD at a time
These are subject to change as we test stuff on a production-sized DB. But these numbers are sensible defaults that the DB should handle just fine, without being too conservative either (testing one night at a time is the most likely usecase for that endpoint, so the max 1 JD constraint is great for that).

TODO: 
- [x] relax the JD max window to a week. We're not in prod yet and data isn't flowing constantly. Being able to select a bit more than a day would be useful.
- [x] Documenting some of the methods in the filter worker would be good. Since we now re-use them in the API here in there, it would make things a bit easier to develop.

### DIFF
--- a/src/filter/base.rs
+++ b/src/filter/base.rs
@@ -137,8 +137,6 @@ pub enum FilterError {
     InvalidFilterPipeline(String),
     #[error("invalid filter id")]
     InvalidFilterId,
-    #[error("unsupported survey for filter")]
-    UnsupportedSurvey(Survey),
     #[error("error during filter execution")]
     FilterExecutionError(String),
 }


Theodlz:
> @petebachant @thomasculino I pinged you both a little early, I still have some things to address before this is worth spending time on to review. I'll ping you again (likely today) when that's done.

Alright, I think it's all good now!

Theodlz:
> Code looks good to me. We might consider this more of a query than a filter management operation and put it in `POST /queries/filter-test`, but as long as we can easily change the client code (front end, Python package someday I guess), we don't need to fuss about naming at this stage.

In future iterations that endpoint will be more tightly coupled with filters (for data access reasons, and enabling testing existing filters by id + version id, not just passing a pipeline as input), so let's keep its route as if for now.

PR #288, opened 2025-11-17
Add driver: bridge to docker-compose

mcoughlin:
This PR adds a required network to the docker-compose to get it running on the Harvard cluster.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -8,6 +8,8 @@ networks:
   traefik-public:
     # Allow setting it to false for testing and local development
     external: true
+  default:
+    driver: bridge
 
 services:
   # Autoheal will monitor containers with autoheal=true label


Theodlz:
adding @petebachant  to the conversation, as he added the traefik network to that docker compose. I thought that a default network was auto-created anyways, but based on this its probably not true. Should be fine to just add a default network right?

@mcoughlin services/containers usually need to then see the network they should be on specified in the network(s) key. Does that need to be done here as well? I wonder if even though you added the default network, the containers aren't on it?

Theodlz:
> I need to do some reading to understand what this does. Note in the deployment procedure in the docs, there's a one-time setup step to create a Docker network on the target machine and set up Traefik, which has its own Docker Compose project. This is done so multiple deployments (e.g., one for prod and one for staging) can share that network and the routing is configured with labels. I'm not sure, but this change may not be necessary if you install that initial Traefik config/project first.

@petebachant I think the issue is we are somewhat forcing the hand when it comes to traefik, when on a dev setup or any deployment where you don't care about exposing the data publicly, you don't need / can't run it. I bet Michael just wants to process alerts with BOOM but not make that public in any way, is that right? In which case just a default network and no traefik is preferable. That said @mcoughlin this can be done just by using a custom dockerfile which is very normal as these need to change based on where things are deployed.

Also, I'm pretty sure the `driver: bridge` is not needed, isn't it the default config? I just just adding `default:` (with no value/empty) would do the trick, can you try that michael?

Issue #280, opened 2025-10-30
Decide on Babamul output schema


petebachant:
Probably something close to original.

Should we build the signup first so we can get feedback from the customers?

Theodlz:
I think we could, but by experience I'm not sure we would get anything useful.

Theodlz:
I want the frontend anyways, so I'll try to get that up and running throughout next week!

PR #293, opened 2025-11-19
Process ZTF alerts for Babamul streams

### LINKED ISSUE #290: Apply Babamul processing to ZTF alert stream
Currently we're only doing this for LSST, but we need to run against both.

petebachant:
Also merged `babamul_auth` into `auth` so at least the middleware functions are close to each other. Babamul authorization isn't used yet since there aren't any protected endpoints.

Resolves #290 

## TODO

- [ ] Can we implement the star/galaxy determination for ZTF here and now?
- [ ] Decide on appropriate key for Kafka messages?
- [x] Add tests if possible.
- [x] Do both ZTF and LSST alerts use the same output schema? Related to #280.
- [x] Try to refactor Babamul module to be even simpler and less verbose.

### DIFF
--- a/src/enrichment/babamul.rs
+++ b/src/enrichment/babamul.rs
@@ -13,7 +13,7 @@ use tracing::{info, instrument};
 
 // Wrapper around cutout bytes, so we can implement
 // AvroSchemaComponent for it, to serialize as bytes in Avro
-#[derive(Debug, serde::Deserialize, serde::Serialize)]
+#[derive(Debug, serde::Deserialize)]
 pub struct CutoutBytes(Vec<u8>);
 
 impl apache_avro::schema::derive::AvroSchemaComponent for CutoutBytes {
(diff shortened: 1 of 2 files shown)


src/enrichment/babamul.rs
```diff
@@ -0,0 +1,269 @@
...
+            {
+                // Skip this alert, it doesn't meet the criteria
+                continue;
+            }
+
+            // Determine which topic this alert should go to
+            // Is it a star, galaxy, or none, and does it have a ZTF crossmatch?
+            // TODO: Get this implemented
+            // For now, all LSST alerts go to "babamul.none"
+            let category: String = "none".to_string();
```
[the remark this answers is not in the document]
petebachant: Yeah, thinking about this more, maybe users would prefer both alerts in the same schema? Something like:

```yaml
source_survey: ztf | lsst # enum
source_alert: # The alert just as it came in from the survey, a union of the two schemas
boom_annotations: # All the stuff we add
# More
```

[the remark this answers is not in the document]
petebachant: Chatted about this a bit with Gemini 3 and was convinced splitting into multiple topics makes the most sense.

src/enrichment/babamul.rs
```diff
@@ -5,40 +5,185 @@ use crate::enrichment::lsst::EnrichedLsstAlert;
...
-        {"name": "fp_hists", "type": "string"},
-        {"name": "properties", "type": "string"}
-    ]
+// Generate a small Avro schema from a Rust type inspected via `schemars`.
+// We only include the fields used by Babamul (the fields that are put into
+// the Avro record). Complex nested Rust types (objects/arrays) are mapped
+// to Avro `string` since we serialize those nested values as JSON strings
+// when producing the Avro payload.
+// Convert a schemars JSON Schema node to an Avro JSON type.
+fn json_schema_node_to_avro(
```
Theodlz: Did you look at https://crates.io/crates/apache-avro-derive (struct -> avro schema), or https://crates.io/crates/avrotize (json schema -> avro schema)? Might be easier than re-implementing it yourself from scratch

petebachant: Switching over to that now. Thanks for the tip!

petebachant:
Alright, there are some tests for Babamul in here now. Ready for another look whenever you are @Theodlz.

src/enrichment/babamul.rs
```diff
@@ -186,6 +168,12 @@ impl Babamul {
                 .push(alert);
         }
 
+        // If there is nothing to send, return early
+        if alerts_by_topic.is_empty() {
+            info!("No LSST alerts to send to Babamul");
```
Theodlz: once we are happy with the code, we could move that info to trace or debug, so they don't pollute the logs

petebachant: True, but it also might be nice to see some confirmation that they were at least processed, even though none will make it to Kafka.

PR #307, opened 2025-12-05
Fix bug on add_filter

antoine-le-calloch:
Fix errors when adding a filter using the add_filter file:
- Add `created_at` and `updated_at` fields
- rename the `catalog` field to `survey`

### DIFF
--- a/docker-compose.override.yaml
+++ b/docker-compose.override.yaml
@@ -69,6 +69,7 @@ services:
     command: ["cargo", "watch", "-x", "run --bin api"]
     volumes:
       - ./src:/app/src
+      - ./apache-avro-macros:/app/apache-avro-macros
       - ./Cargo.toml:/app/Cargo.toml
       - ./Cargo.lock:/app/Cargo.lock
       - cargo-registry:/usr/local/cargo/registry


src/bin/add_filter.rs
```diff
@@ -16,6 +16,11 @@ struct Cli {
     filter_file: String,
 }
 
+fn now_jd() -> f64 {
```
Theodlz: You can use flare for this, it's just `flare::Time::now().to_jd()` or similar (plenty of examples of this in the code), so you don't have to deal with the jd conversion here again.

src/bin/add_filter.rs
```diff
@@ -49,17 +54,20 @@ async fn main() {
...
                 "fid": "v2e0fs",
-                "pipeline": filter_pipeline
+                "pipeline": filter_pipeline,
+                "created_at": now_jd(),
             }
         ],
         "active_fid": "v2e0fs",
+        "created_at": now_jd(),
+        "updated_at": now_jd(),
     };
```
Theodlz: In general I think this could all be done much more elegantly. This binary is very deprecated and existed in a world where we didn't have structs to define filters yet. An even better solution would be to check how things are now done in the API (caus that's what we now use to post filters, and therefore is up to date).

Theodlz (commented):
Thanks for the fix Antoine, that fell through the gaps because we don't use that binary anywhere. I left some comments for you. The JD calculation one needs to be addressed, but the other one requires more work so feel free to skip it for this PR, if you don't have the time for that

src/bin/add_filter.rs
Theodlz: did you ever have a chance to look at this once @antoine-le-calloch ?

PR #308, opened 2025-12-05
Babamul API: Getting web app ready

Theodlz:
WIP, DNM

Good ol' unit tests need to be added and the get_obj endpoint deserves a cleanup, but it all works great when paired with [boom-web](https://github.com/boom-astro/boom-web) :)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,41 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
+name = "aead"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
+dependencies = [
+ "crypto-common",
+ "generic-array",
+]
+
+[[package]]
+name = "aes"
+version = "0.8.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
+dependencies = [
+ "cfg-if",
+ "cipher",
+ "cpufeatures",
+]
+
+[[package]]
+name = "aes-gcm"
+version = "0.10.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "831010a0f742e1209b3bcea8fab6a8e149051ba6099432c8cb2cc117dec3ead1"
+dependencies = [
+ "aead",
+ "aes",
+ "cipher",
+ "ctr",
+ "ghash",
+ "subtle",
+]
+
+[[package]]
 name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -542,6 +577,7 @@ version = "0.1.0"
 dependencies = [
  "actix-rt",
  "actix-web",
+ "aes-gcm",
  "apache-avro",
  "apache-avro-derive",
  "apache-avro-macros",
@@ -576,6 +612,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
+ "sha2",
  "tempfile",
  "thiserror 2.0.17",
  "tokio",
@@ -956,7 +993,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1bfb12502f3fc46cca1bb51ac28df9d618d813cdc3d2f25b9fe775a34af26bb3"
 dependencies = [
  "generic-array",
+ "rand_core 0.6.4",
  "typenum",
+]
+
+[[package]]
+name = "ctr"
+version = "0.9.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0369ee1ad671834580515889b80f2ea915f23b8be8d0daa4bbaf2ac5c7590835"
+dependencies = [
+ "cipher",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


src/api/routes/babamul/users.rs
```diff
@@ -500,7 +576,7 @@ pub struct BabamulAuthResponse {
 #[utoipa::path(
     post,
     path = "/babamul/auth",
```
petebachant: I probably would have the module path mimic the URL path and not put this in `users.rs`; maybe just in `mod.rs` since it's a top level Babamul endpoint.

src/api/routes/kafka.rs
```diff
@@ -18,13 +18,13 @@ use std::str;
 #[get("/kafka/acls")]
 pub async fn get_kafka_acls(
     current_user: web::ReqData<User>,
-    kafka_producer_config: web::Data<crate::conf::KafkaProducerConfig>,
+    config: web::Data<crate::conf::AppConfig>,
```
petebachant: Nice 👍

src/api/email.rs
```diff
@@ -102,19 +103,39 @@ impl EmailService {
...
-             After activation, you'll receive a password that you can use to:\n\
-             1. Connect to Kafka streams (topics: babamul.*)\n\
-             2. Authenticate to the Babamul API\n\n\
-             This code will expire in 24 hours.\n\n\
-             If you did not request this, please ignore this email.",
-            activation_code, to_email, activation_code
-        );
+        // if webapp url exists then use it and give instructions to activate via web link
+        // else just give activation code and instructions to activate via code only
+        let email_body = if let Some(url) = webapp_url {
```
petebachant: Instead of defining a `webapp_url`, it might be better to define a `domain` for the entire BOOM application. Then the front end would be hosted at `domain`, the API at `api.domain`, etc. By default, or for testing, the domain would be `localhost`.

src/api/routes/babamul/surveys.rs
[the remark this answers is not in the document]
Theodlz: > Why a rename instead of a new module and set of routes? Don't we want an endpoint that returns all surveys in the system, not just the Babamul ones, so regular users can also see what surveys are in the system?

Not so sure what you meant by `Why a rename instead of a new module and set of routes?` since what that PR does is create a babamul module with its own set of routes. I moved this boom route to babamul because that's why it was added initially. The assumption was always that the "regular" API of BOOM gives full access to the DB anyways and is for advanced users with the right data access levels (at least that's what it is for ZTF), that will require specific queries and won't be satisfied by an endpoint like this one designed for babamul (or rather its frontend).

Not saying we can't also just add an endpoint like this one to easily retrieve object-level info in BOOM, just that it's not as important as for Babamul and if we want a similar/identical endpoint for both APIs, then it raises the following question: how do we best avoid duplication between these 2 distinct APIs and if they have sufficient overlap should they just be combined into one where user accounts have permission that condition which endpoints are accessible>

src/api/routes/babamul/users.rs
```diff
@@ -569,6 +650,31 @@ pub async fn post_babamul_auth(
...
+    path = "/babamul/profile",
+    responses(
+        (status = 200, description = "User profile retrieved successfully", body = BabamulUserPublic),
+        (status = 401, description = "Unauthorized"),
+        (status = 500, description = "Internal server error")
+    ),
+    tags=["Babamul"]
+)]
+#[get("/babamul/profile")]
+pub async fn get_babamul_profile(current_user: Option<web::ReqData<BabamulUser>>) -> HttpResponse {
```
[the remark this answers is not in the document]
Theodlz: unless I made a mistake somewhere after the line you highlighted for this comment, this endpoint was definitely meant to require auth. I will check

src/api/email.rs
Theodlz: I'm not a fan of tightly coupling a web app deployment from its backend, which is why I'd rather have a distinct parameter for a web app's domain and backend's.

Theodlz: I did add a api.domain parameter to the config (can be overwritten with BOOM_API__DOMAIN), that we can use anywhere we need to know what's the public URL for the API. But, I left the web app url parameter as is, so deployments can be done separately

petebachant: Fair enough--your call. However, consider the case where a change needs to be made to both the front end back end (an indication of tight coupling, which is likely to happen with an app like this anyway). That would take two separate coordinated PRs and deployments. It's a little more painful than a single deployment pipeline for the whole stack IMO. Again, your call. I'll shut up about it now, heh.

petebachant:
The Kafka auth changes make sense to me. Thanks for the detailed explanation!

src/api/email.rs
Theodlz: I'm super fine with coupling the deployment and even the dev. The domain name problem remains the same. the web app isn't hosted on the same domain as the rest (babamul.caltech.edu, vs api.kaboom.caltech.edu). For the babamul-specific web app it's important we run on a babamul-named domain. Question is do we just run everything under babamul.caltech.edu (including the boom specific stuff) or not. So, I'd rather not force everything under the same domain. + I don't have control over creating new domains/redirections (IT folks do), so I'd rather not couple stuff since I can't easily make these changes in prod.

The API itself does run under the same domain as everything else so yes that can be updated. Actually, we are already using the domain (check a few lines under). So the webapp is just an exception in the rust code, but we can easily just set these 2 values (webapp_url, and domain to the same value when we deploy with docker compose

.github/workflows/test.yaml
```diff
@@ -41,6 +43,24 @@ jobs:
...
+          echo "Attempt $i/60 - MongoDB not ready yet, waiting..."
+          sleep 2
+        done
+        # Final check to confirm MongoDB is ready
+        if ! docker compose exec -T mongo mongosh localhost:27017/test --quiet --eval "db.runCommand('ping').ok" >/dev/null 2>&1; then
+          echo "MongoDB failed to start within timeout" >&2
+          docker compose logs mongo || true
+          exit 1
+        fi
+        docker compose exec -T mongo mongosh localhost:27017/boom_test --quiet --username mongoadmin --password mongoadminsecret --authenticationDatabase admin --eval "db.createCollection('LSPSC'); db.LSPSC.createIndex({'coordinates.radec_geojson': '2dsphere'});"
```
petebachant: I believe there are some database init steps in the Rust code with a branch for Babamul. Could move this in there for simplicity's sake.

Theodlz:
Thanks a lot Pete!!! I will open issues and other PRs for all your minor comments, I definitely want to address these ASAP. I have to get a version of the API deployed extra ASAP, so rogue merging this right now 😓

PR #311, opened 2025-12-17
Babamul multi-survey support (bugfixes and improvements)

Theodlz:
- [x] revert some changes to avoid adding an extra addFields pipeline to make aux array fields into single docs
- [x] add get_array_dict_element where we use {} as the missing value placeholder for
- [x] revert change to fetch_timeseries_op since all its inputs are back to being array like (and therefore need a get_array_element fn call)
- [x] add ZtfMatch and LsstMatch structs, and edit the create_<survey>_alert_pipeline methods to format the survey matches fit with these new structs
- [x] newly added <Survey>Match structs include more information right lightcurves as well as position
- [x] since we now get multisurvey data in the enrichment worker processing we add a multisurvey_photstats field
- [x] updated babamul unit tests
- [ ] when creating the babamul output of LSST alerts, remove non-public ZTF data from the ZTF match lightcurves
- [ ] the babamul unit tests create aux table entries by writing documents manually. It's very prone to errors (for example here, the documents were a clear mismatch with the data model!!!). I know we have a document DB but we also have the data model clearly modeled as structs now, so we should use that and never create documents from scratch unless it's really a lot easier. Structs for everything is the way to go. Here the unit tests kept failing as I changed or fixed the code, not because of the changesd were incorrect but because the documents created in the tests were incorrect or not in sync.

### DIFF
--- a/apache-avro-macros/src/lib.rs
+++ b/apache-avro-macros/src/lib.rs
@@ -122,7 +122,14 @@ fn serdavro_impl(item: &ItemStruct) -> syn::Result<TokenStream> {
 
                             Schema::Record(record)
                         }
-                        _ => panic!("Only record schemas are supported")
+                        // Support Schema::Ref by cleaning up the name.
+                        Schema::Ref { name } => {
+                            let name_str = name.to_string();
+                            let clean_name_str = name_str.trim_start_matches("__SERDAVRO__").trim_end_matches("__");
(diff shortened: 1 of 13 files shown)


Theodlz:
Just putting this on your radar @petebachant. Feel free to commit to this PR directly or merge with yours (in which case the non-completed TODO items need to be moved to the other one)

PR #313, opened 2025-12-18
Cutout Storage: Mongo and S3 support

Theodlz:
This pull request introduces support for multiple cutout storage backends (shared MongoDB, dedicated MongoDB, and S3-compatible object storage) and updates the configuration, documentation, and development workflow to make switching between these modes straightforward.

* Added a new CutoutStorage concept, where at startup one can configure if the apps use mongo to storage cutouts or an S3 storage bucket. For mongo: we can pick between the same mongo as the alerts or a separate one; For S3: we can pick between a local (rustfs) or distant instance. S3 is not nearly as fast a mongo regardless of what one does because objects can't be retrieved in bulk i.e. an expensive loop with network overheads is needed in the enrichment and filter workers. So when using S3 we also spawn a dedicated Valkey instance. dedicated one because that way we can have a memory limit on it. We also use a TTL (default 30s) for the cutouts in the cache (when we hit the memory limit, cutouts that are closest to their expiration date are evicted first). This way the enrichment and filter workers stay quick for real time processing, while still allowing us to use S3.
* While ZTF cutouts are already GZIPED images, that's not true for LSST. So there, we compress the data when caching and storing to S3, but using zstd which is more usually efficient (in resource utilization, speed, and final file size). No need for that in Mongo, which compressed the data by default.
* Added configuration options for three cutout storage modes (shared MongoDB, dedicated MongoDB container, and S3-compatible storage both local and external) in `.env.example`, `config.yaml`, and the `README.md`, with detailed setup instructions and environment variable documentation.
* Added new Docker Compose overwrites for all the deployment modes described above, and corresponding Makefile targets (`dev-mongo`, `dev-s3`, `dev-s3-external`) to simplify local and production setup.
* Updated GitHub Actions workflows (`test.yaml` and `test-throughput.yaml`) to run tests and benchmarks against both MongoDB and S3 cutout storage backends using a matrix strategy, ensuring both modes are continuously tested. The throughput benchmarking workflow now aggregates and report results for each storage backend, compare against baselines, and fail the build if performance regressions exceed a set tolerance. BTW, it will say NaN for the baseline since the cache key has changed, but for the purpose of this PR please just compare to main manually if needed.

As the throughput test will show you, S3 is still slower than Mongo (even after this cache optimization). But, that isn't the cache on my machine. This indicates to me that the performance is just poorer on github action runners, perhaps because of the limited amount of threads, memory, and/or IO. Might be worth playing w/ the TTL time configured for the github CI.


tests/throughput/compose.yaml
```diff
@@ -42,6 +43,40 @@ services:
...
+  rustfs:
+    image: rustfs/rustfs:latest
+    hostname: rustfs
+    restart: unless-stopped
+    expose:
+      - "9000"
+      - "9001"
+    ports:
+      - "9000:9000" # For local testing; remove in production
+      - "9001:9001" # For local testing; remove in production
```
petebachant: Note this is the throughput config. I think you want this in `docker-compose.override.yaml`, which doesn't get used in the prod deployment, only local dev.

docker-compose.cutouts-s3.yaml
```diff
@@ -0,0 +1,107 @@
...
+
+  # Point all boom services at S3/rustfs for cutout storage.
+  # Credentials (BOOM_CUTOUTS_STORAGE__ACCESS_KEY / BOOM_CUTOUTS_STORAGE__SECRET_KEY) must be set in .env.
+  # All other S3 fields are infrastructure-determined and set here so that
+  # config.yaml does not need to be edited for Docker deployments.
+  api: &s3_cutouts_env
+    environment:
+      BOOM_CUTOUTS_STORAGE__TYPE: s3
+      BOOM_CUTOUTS_STORAGE__REGION: local
+      BOOM_CUTOUTS_STORAGE__ENDPOINT_URL: http://rustfs:9000
```
petebachant, line 67: Should the bucket name also be parameterized with a sane default?

Theodlz: At the moment, the bucket name is inferred from the survey (which also means we have one bucket for each survey), as `let bucket_name = format!("{}-cutouts", survey.to_string().to_lowercase());`. Would you recommend we take another route? Like using a single bucket (which name would be configurable here) and simply starting the file names with the survey names (e.g. `lsst-128439887612358123.json` or `lsst/28439887612358123.json`, where that ugly number is the candid)? Or perhaps storage buckets let you put things in some equivalent of "subdirectories", where we can have one directory per survey (but all in a single bucket)? @petebachant

Theodlz: Ok I already moved to a single bucket approach with survey-based prefixes. No reason I can come up with to have multiple buckets, since TTLs (the only thing I worried about, since for LSST we'll want to have some lifetime rule to keep images only up to N days) can have "prefix" filter rules in S3. Makes deploying to something other than rustfs a little easier!

src/alert/decam.rs
```diff
@@ -423,19 +426,22 @@ impl AlertWorker for DecamAlertWorker {
...
+        let status = self
+            .format_and_insert_alert(candid, &alert, &self.alert_collection)
             .await
             .inspect_err(as_error!())?;
 
-        if let ProcessAlertStatus::Exists(_) = cutout_status {
-            return Ok(cutout_status);
+        if let ProcessAlertStatus::Exists(_) = status {
+            return Ok(status);
         }
```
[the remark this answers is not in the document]
Theodlz: well that's not new. we had this issue before when we inserted cutouts first and skipped processing further if they already existing (in which case we had cutouts, but not the alerts). Personally, until a fail-proof alternative is selected, having alerts w/ no cutouts seems better than cutouts with no alerts.

src/api/routes/surveys/cutouts.rs
```diff
@@ -29,24 +30,26 @@ pub async fn get_cutouts(
...
     let survey = path.into_inner();
-    let cutout_collection: Collection<AlertCutout> =
-        db.collection(&format!("{}_alerts_cutouts", survey));
+    let cutout_storage = match config.build_cutout_storage(&survey).await {
+        Ok(storage) => storage,
+        Err(error) => {
+            tracing::error!("Error building cutout storage: {}", error);
+            return response::internal_error("error building cutout storage");
+        }
+    };
```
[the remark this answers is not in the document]
Theodlz: outdated, we now build it once at API startup

src/alert/lsst.rs
```diff
@@ -1148,19 +1151,23 @@ impl AlertWorker for LsstAlertWorker {
...
+        let status = self
+            .format_and_insert_alert(candid, &alert, &self.alert_collection)
             .await
             .inspect_err(as_error!())?;
 
-        if let ProcessAlertStatus::Exists(_) = cutout_status {
-            return Ok(cutout_status);
+        if let ProcessAlertStatus::Exists(_) = status {
+            return Ok(status);
         }
```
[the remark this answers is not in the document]
Theodlz: well that's not new. we had this issue before when we inserted cutouts first and skipped processing further if they already existing (in which case we had cutouts, but not the alerts). Personally, until a fail-proof alternative is selected, having alerts w/ no cutouts seems better than cutouts with no alerts.

src/utils/cutouts.rs
```diff
@@ -0,0 +1,668 @@
...
+        self.cache.set(&cutouts).await;
+        Ok(())
+    }
+
+    #[instrument(skip_all, err)]
+    pub async fn retrieve_cutouts(&self, candid: i64) -> Result<AlertCutout, CutoutStorageError> {
+        let raw = if let Some(cutout) = self.cache.get(candid).await {
+            cutout
+        } else {
+            retrieve_alert_cutouts(candid, &self.bucket_name, &self.s3_client).await?
```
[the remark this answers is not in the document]
Theodlz: fixed

src/conf.rs
```diff
@@ -148,6 +156,99 @@ async fn build_redis(
...
+}
+
+#[instrument(skip_all, err)]
+async fn build_cutout_storage(
+    survey: &Survey,
+    conf: &AppConfig,
+) -> Result<CutoutStorage, BoomConfigError> {
+    let storage = match &conf.cutouts_storage {
+        CutoutsStorage::S3(s3_conf) => {
+            let credentials_static_str = string_to_static_str(s3_conf.credentials_provider.clone());
```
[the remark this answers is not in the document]
Theodlz: outdated. the api now creates the cutout storage on startup, so the Box::leak is not invoked more than once per survey, which is just a few bytes.

tests/throughput/_run.sh
```diff
@@ -39,8 +39,16 @@ fi
 COMPOSE_CONFIG=("-f" "$BOOM_REPO_ROOT/tests/throughput/compose.yaml")
 BG_PIDS=()
 
-PLATFORM=$(uname -s | tr '[:upper:]' '[:lower:]')
+# Select the cutout storage overlay based on BOOM_CUTOUTS_STORAGE__TYPE (default: s3)
+CUTOUTS_TYPE="${BOOM_CUTOUTS_STORAGE__TYPE:-s3}"
+if [ "$CUTOUTS_TYPE" = "s3" ]; then
+    COMPOSE_CONFIG+=("-f" "$BOOM_REPO_ROOT/tests/throughput/compose.cutouts-s3.yaml")
+else
+    COMPOSE_CONFIG+=("-f" "$BOOM_REPO_ROOT/tests/throughput/compose.cutouts-mongo.yaml")
```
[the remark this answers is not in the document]
Theodlz: fixed

.env.example
```diff
@@ -8,6 +8,28 @@ STACK_NAME=boom
...
+#   Leave unset to use a Docker-managed named volume (mongodb_cutouts).
+# BOOM_CUTOUTS_MONGO_VOLUME=/mnt/large-disk/boom_cutouts
+#
+# S3/rustfs (make dev-s3):
+#   Required credentials for the rustfs container and boom services.
+BOOM_CUTOUTS_STORAGE__ACCESS_KEY=rustfsadmin
+BOOM_CUTOUTS_STORAGE__SECRET_KEY=rustfsadminsecret # Replace with something secure in production
+#   Optional: memory cap for the dedicated valkey-cutouts cache.
+#   When full, keys with the shortest remaining TTL are evicted first (volatile-ttl).
+# BOOM_CUTOUTS__CACHE__MAXMEMORY=1024mb
```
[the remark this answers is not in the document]
Theodlz: you missed the actual problem, but I fixed this variable naming situation nonetheless

src/alert/ztf.rs
```diff
@@ -978,20 +981,22 @@ impl AlertWorker for ZtfAlertWorker {
...
+
+        let status = self
+            .format_and_insert_alert(candid, &alert, &self.alert_collection)
             .await
             .inspect_err(as_error!())?;
 
-        if let ProcessAlertStatus::Exists(_) = cutout_status {
-            return Ok(cutout_status);
+        if let ProcessAlertStatus::Exists(_) = status {
+            return Ok(status);
```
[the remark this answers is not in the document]
Theodlz: well that's not new. we had this issue before when we inserted cutouts first and skipped processing further if they already existing (in which case we had cutouts, but not the alerts). Personally, until a fail-proof alternative is selected, having alerts w/ no cutouts seems better than cutouts with no alerts.

src/api/routes/babamul/surveys/cutouts.rs
```diff
@@ -39,22 +41,23 @@ pub async fn get_cutouts(
...
 
-    let cutout_collection: Collection<AlertCutout> =
-        db.collection(&format!("{}_alerts_cutouts", survey));
+    let cutout_storage = match config.build_cutout_storage(&survey).await {
+        Ok(storage) => storage,
+        Err(error) => {
+            tracing::error!("Error building cutout storage: {}", error);
+            return response::internal_error("error building cutout storage");
+        }
+    };
```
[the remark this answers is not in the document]
Theodlz: now built once at API startup

.github/workflows/test-throughput.yaml
frenbox, line 1: Seems to be working good and the CI (and local testing) is also showing similar times. Have you done a test on a S3 bucket itself? How much does the network read/write overhead add to the latency?

src/alert/base.rs
```diff
@@ -1135,33 +1102,38 @@ pub trait AlertWorker {
...
+            cutout_template,
+            cutout_difference,
+            cutout_storage
+        ),
+        err
+    )]
     async fn format_and_insert_cutouts(
         &self,
         candid: i64,
+        object_id: &str,
```
frenbox, line 1118: `object_id` is not really used later in the script.

src/conf.rs
```diff
@@ -84,9 +90,7 @@ pub fn load_raw_config(filepath: &str) -> Result<Config, BoomConfigError> {
 }
```
antoine-le-calloch, line 91: The custom `CutoutsStorage` deserializer round-trips through serde_json::Value, so env vars like BOOM_CUTOUTS_STORAGE__CACHE__PORT=6379 stay strings instead of being coerced to u16, breaking docker-compose overlays.

You can either enable `.try_parsing(true)` on the Environment source, or coerce numeric strings inside the custom deserializer before serde_json::from_value.

src/utils/cutouts.rs
```diff
@@ -0,0 +1,757 @@
...
+        create_bucket_if_not_exists(&bucket_name, &s3_client).await?;
+        Ok(Self {
+            s3_client,
+            bucket_name,
+            key_prefix,
+            concurrency_limit: concurrency_limit.unwrap_or(1),
+            cache,
+            compress_stamps,
+        })
+    }
```
antoine-le-calloch, line 463: `S3CutoutStorage.concurrency_limit` is hardcoded to 1. `S3CutoutsStorageConfig` doesn't expose the field, so build_cutout_storage (conf.rs:246) passes None to CutoutStorage::from_s3.

src/conf.rs
```diff
@@ -667,6 +853,20 @@ impl AppConfig {
...
+
+    #[instrument(skip_all, err)]
+    pub async fn build_cutout_storage(
+        &self,
+        survey: &Survey,
+    ) -> Result<CutoutStorage, BoomConfigError> {
+        match build_cutout_storage(survey, self).await {
+            Ok(storage) => Ok(storage),
+            Err(e) => {
+                println!("Failed to build cutout storage: {}", e);
```
antoine-le-calloch: Why a `println` here, and not something like `as_error` ?

src/conf.rs
```diff
@@ -148,6 +156,109 @@ async fn build_redis(
...
+            let s3_config = s3_config_builder.load().await;
+
+            let rustfs_client = aws_sdk_s3::Client::from_conf(
+                aws_sdk_s3::Config::from(&s3_config)
+                    .to_builder()
+                    .force_path_style(true)
+                    .build(),
+            );
+            let bucket_name = s3_conf.bucket_name.clone();
+            let key_prefix = survey.to_string().to_lowercase();
```
antoine-le-calloch, line 231: Is there a reason not to have the same name as we have in mongo `ZTF_alerts_cutouts` ?

Cargo.toml
```diff
@@ -39,8 +41,9 @@ redis = { version = "0.32.7", features = ["tokio-comp"] }
 regex = "1.12.2"
 reqwest = { version = "0.12.24", features = ["json", "stream"] }
 serde = { version = "1.0", features = ["derive"] }
+serde_bytes = "0.11.19"
```
antoine-le-calloch: this seems unused.

src/conf.rs
```diff
@@ -703,6 +903,19 @@ pub async fn get_test_db() -> Database {
     config.build_db().await.unwrap()
 }
 
+pub async fn get_test_redis_connection() -> redis::aio::MultiplexedConnection {
+    let config = AppConfig::from_test_config().expect("Failed to load test config");
+    config.build_redis().await.unwrap()
+}
+
```
antoine-le-calloch: Unused ?

src/bin/api.rs
```diff
@@ -27,6 +31,19 @@ async fn main() -> std::io::Result<()> {
     // Initialize email service
     let email_service = EmailService::new();
 
+    // Build cutout storage for each survey once at startup
+    let mut cutout_storage_map: HashMap<Survey, CutoutStorage> = HashMap::new();
+    for survey in [Survey::Ztf, Survey::Lsst, Survey::Decam] {
```
antoine-le-calloch, line 40: api.rs initializes cutout storage for ZTF, LSST, and DECam, so even deployments not serving DECam still connect to every survey’s S3/Redis backend. Any failure on an unused survey prevents the API from starting.

src/filter/lsst.rs
```diff
@@ -145,9 +143,9 @@ pub async fn build_lsst_alerts(
         );
     }
 
-    let mut candid_to_cutouts = fetch_alert_cutouts(&candids, &alert_cutout_collection)
-        .await
-        .map_err(|e| FilterWorkerError::FetchCutoutsError(e.to_string()))?;
+    let mut candid_to_cutouts = alert_cutout_storage
+        .retrieve_multiple_cutouts(&candids, true)
+        .await?;
```
antoine-le-calloch, line 149: `populate_cache: true` is wasted work here. `build_lsst_alerts` immediately calls `evict_from_cache(&candids)`, so cache misses trigger a SET EX followed right away by a DEL on the same keys, with no reader in between.

src/filter/ztf.rs
```diff
@@ -187,9 +188,9 @@ pub async fn build_ztf_alerts(
         );
     }
 
-    let mut candid_to_cutouts = fetch_alert_cutouts(&candids, &alert_cutout_collection)
-        .await
-        .map_err(|e| FilterWorkerError::FetchCutoutsError(e.to_string()))?;
+    let mut candid_to_cutouts = alert_cutout_storage
+        .retrieve_multiple_cutouts(&candids, true)
+        .await?;
```
antoine-le-calloch, line 194: Same here. `populate_cache: true` is wasted work.

src/bin/api.rs
```diff
@@ -27,6 +31,19 @@ async fn main() -> std::io::Result<()> {
...
+            .await
+            .unwrap_or_else(|e| {
+                panic!("Failed to initialize cutout storage for {}: {}", survey, e)
+            });
+        cutout_storage_map.insert(survey, storage);
+    }
+    let cutout_storages = web::Data::new(cutout_storage_map);
+
     // Initialize logging
     env_logger::init_from_env(env_logger::Env::default().default_filter_or("info"));
```
antoine-le-calloch: Maybe we should move this above the cutout block.

.github/workflows/test-throughput.yaml
Theodlz: Hey Sushant. Not sure what you mean, but you probably meant to ask if I tried with a remote (and not local) s3 bucket? If so, then the answer is no. I don't have access to one. If you do, that would be a nice thing to try. That said I don't think this will have an impact on the implementation anyways, right? There is already a caching layer that's there to help with how slow S3 can be. So the only direct write in the "hot-path" (for real time processing that needs to be super duper fast) happens in the alert worker, which can be scaled to compensate any slowness.

src/alert/base.rs
Theodlz: it's there for the instrumentation that's above. That way if it fails with an error, we know the object_id will show up. There's multiple functions in here that have the same pattern.

src/conf.rs
Theodlz: looks like a claude/chatgpt comment hahaha. But that's true. However, this is already true for all the non-string values in the config. We only saw this becoming a problem when we have no value in the config (no default), and used an env var to overwrite. When the value isn't optional and we have a default value in the config (which is what we have here), it seems to work just fine. I can add `.try_parsing(true)` if you want, but I don't think we actually need it here?

src/utils/cutouts.rs
Theodlz: yep, I had this concurrency mechanism built-in but my opinion on it is that it makes our workers use many more threads and it gets hard to track stuff. So I kept it here in case one day we really need concurrency at the S3 level to keep things fast enough (might become mandatory for the UMN application, when we reprocess data and need batches of cutouts that are not in cache), but decided not to make it a configurable thing just yet. If you think that deserves an in-code comment or should just be removed, let me know.

src/conf.rs
Theodlz: that's an error for sure! Let me fix it.

Theodlz: we can just return the error, but I felt like a nice log message to alert us was useful. I'll use `error!` instead of the print.

.github/workflows/test-throughput.yaml
frenbox: Understandable!

src/alert/base.rs
frenbox: Ahh! I didn't think about it like that.

src/conf.rs
Theodlz: ok so the bucket name is not necessarily something you choose. If you're using a distant S3 server, you may have access to an existing bucket and can't create one with a new name. So, we make it configurable with some reasonable default (boom-cutouts). Similarly, you may only have access to one bucket and not one per-survey. So in the current implementation, we use only one bucket for all our cutouts.

In S3, which is just a glorified file system with a network interface on top, you want keys to look like paths you'd use on disk. So we start our keys/paths by the survey name, then have a filename like <candid>.json. It just felt more appropriate to have things structured this way for S3. As a matter of fact, I just committed a change to the filenames where I now have `<survey>/cutouts/<candid>.json` to future-proof us a bit, if one day we want to store more than cutouts.

So I guess the answer is just: felt more s3-compliant to use a file-path-like structure and therefore stick with non-case sensitive names. I didn't see any advantage to have the same key-prefix as the mongo collection names.

src/bin/api.rs
Theodlz: there is a single bucket for all surveys, so just one s3 server we connect too. So this just means we maintain 3 connections, which doesn't come at any cost. Since all surveys anyway point to a single server, one of them failing to do so means something is wrong for the server and that's an issue for all surveys. I definitely don't think we want to start the API is the S3 server is unreachable.

src/filter/lsst.rs
Theodlz: ah good call! It's only useful for the enrichment so we have data ready for the filter worker, but yeah no need to populate the cache here. thx!

src/filter/ztf.rs
Theodlz: yep, removing

src/bin/api.rs
Theodlz: I don't understand. You mean we define the `web::Data` first then edit its value? Felt more natural to me to create the hashmap -> populate it -> then pass its ownership to the web::Data. I don't think I can edit it once its in that wrapper.

src/filter/lsst.rs
Theodlz: Since we have a dedicated cutout storage cache now (that doesn't harm other valkey use cases, that are running in the other container), I'm starting to wonder if we should just never evict from the cache at all and just let the TTL do that for us (when its reached, or when redis removes the older TTLed item when we reach the memory cap). That way for real time operations, the cutouts are still in cache when babamul people get alerts from Kafka and hit the API to grab the images. Maybe something to test and change in its own PR, but for now let's evict explicitely and not populate the cache again in the filter worker (which is useless as you pointed out since we evict right away).

src/alert/base.rs
Theodlz: no worries, I mean you are 100% right, as far as the function itself is concerned it's an unused variable. But the indicator is that the IDE probably doesn't flag it as unused (or at least mine doesn't), while it usually does that. That's often a good indicator that something is actually using it - even if not explicitely - like the tracing instrument macro here.

src/utils/cutouts.rs
```diff
@@ -0,0 +1,761 @@
...
+    s3_client: &aws_sdk_s3::Client,
+) -> Result<(), CutoutStorageError> {
+    let candid = cutouts.candid;
+    let key = cutout_s3_key(key_prefix, candid);
+
+    let encoded = serde_json::to_vec(&S3AlertCutout::from(cutouts))?;
+    let body = aws_sdk_s3::primitives::ByteStream::from(encoded);
+
+    match s3_client
+        .put_object()
```
frenbox, line 357: Hey Theo,

The MongoDB counterpart checks for existing cutout (though internal write error).  I was trying to look up if similar thing exist for S3. But overwriting might not be a big deal as well.

src/bin/api.rs
antoine-le-calloch: I was talking about `env_logger::init_from_env(env_logger::Env::default().default_filter_or("info"));`

Theodlz: Oh I see, obviously. My bad. Fixing it now

src/utils/cutouts.rs
Theodlz: unfortunately it doesn't exist for S3. that's why I changed the alert worker logic. Before we started by inserting cutouts and if they were already there skipped processing the alert. Now we start by inserting the alert and if its already there we skip processing. Both are equivalent anyways. A better logic would be "check if the alert exist and if so skip processing, only insert the alert at the very end of the processing loop so we know for sure its been fully processed if the alert is there", but that means you do one read before doing a write, and that's extra latency. Maybe that's an improvement for another PR, since it's unrelated to the current changes.

src/bin/api.rs
Theodlz: i moved the env logger init to the very beginning of the `main` function

antoine-le-calloch (approved):

Theodlz:
@frenbox I made some changes since your last review, minor but tagging you again in case you want to verify them.

@petebachant would be great if you could review this soon. This is a blocker to finish transferring data over from Kowalski, which I need to wrap up in just a few days.

docker-compose.cutouts-s3.yaml
petebachant: The only hangup I could see is using cloud providers, where bucket names typically need to be unique globally. But it probably makes sense to keep hard-coded for now and only add the bucket name parameter if a given instance needs it. Does UMN need it now?

docker-compose.cutouts-s3-external.yaml
```diff
@@ -0,0 +1,83 @@
...
+      retries: 3
+
+  # Point all boom services at the external S3 bucket.
+  api: &s3_external_cutouts_env
+    environment:
+      BOOM_CUTOUTS_STORAGE__TYPE: s3
+      BOOM_CUTOUTS_STORAGE__REGION: ${BOOM_CUTOUTS_STORAGE__REGION:?BOOM_CUTOUTS_STORAGE__REGION must be set (e.g. us-east-1)}
+      BOOM_CUTOUTS_STORAGE__ACCESS_KEY: ${BOOM_CUTOUTS_STORAGE__ACCESS_KEY:?BOOM_CUTOUTS_STORAGE__ACCESS_KEY must be set}
+      BOOM_CUTOUTS_STORAGE__SECRET_KEY: ${BOOM_CUTOUTS_STORAGE__SECRET_KEY:?BOOM_CUTOUTS_STORAGE__SECRET_KEY must be set}
+      BOOM_CUTOUTS_STORAGE__BUCKET_NAME: ${BOOM_CUTOUTS_STORAGE__BUCKET_NAME:-boom-cutouts}
```
petebachant, line 71: Ah, I see we have a bucket name parameter here. That negates the comment above.

petebachant (approved):
This is huge, so I didn't get into crazy fine detail. I'd say go for it unless you want specific feedback on any decisions that are very hard to reverse!

Theodlz:
> This is huge, so I didn't get into crazy fine detail. I'd say go for it unless you want specific feedback on any decisions that are very hard to reverse!

@petebachant actually I'd love feedback like that! I wouldn't want to make an irreversible mistake 😉. No worries if an in-depth review needs one or 2 more days. I need to finish reviving the existing "kowalski data import script", and won't get to that before my Friday morning.

petebachant:
Something to think about for the future, but requires client-side changes, is using presigned URLs for sending cutouts to users. This takes a load off the BOOM API server since all data does not need to flow through it--users can download directly from object storage.

petebachant:
> I wouldn't want to make an irreversible mistake

Is there anything in here that you're worried might be hard to change later if it turned out to be a suboptimal choice?

Theodlz:
> > I wouldn't want to make an irreversible mistake
> 
> Is there anything in here that you're worried might be hard to change later if it turned out to be a suboptimal choice?

Nothing specific! I was just agreeing with your comment :)

frenbox (approved):
It would be great if @antoine-le-calloch could test this (we have a S3 bucket).

I ran the benchmark locally in my computer, everything works fine and I looked at some finer details of some files (but not all of it).

docker-compose.cutouts-s3.yaml
petebachant: And on single versus multiple buckets: one bucket per application is a good place to start. Splitting one bucket per survey would be unnecessary complexity at this stage IMO.

docker-compose.cutouts-s3-external.yaml
petebachant: In the past I have regretted being this specific about bucket names/purposes up front. May want to consider using `boom` as the bucket name and use `cutouts` as a prefix inside. So the purpose of the bucket then becomes BOOM object (not in the astronomical sense) storage rather than BOOM cutout storage.

ChatGPT discussion on the topic also leans towards starting with one bucket per app and expanding later: https://chatgpt.com/share/6a072a4b-4930-83e8-bf93-c35d03c5efb2

antoine-le-calloch:
> It would be great if @antoine-le-calloch could test this (we have a S3 bucket).
> 
> I ran the benchmark locally in my computer, everything works fine and I looked at some finer details of some files (but not all of it).

@Theodlz It will take some time for me to run it, so don't hesitate to merge meanwhile.

Issue #292, opened 2025-11-17
Properly split Babamul alerts into multiple streams


petebachant:
Currently they all go to `babamul.none`, but we need to put them into different categories.

## Some notes from @Theodlz:

when babamul is enabled, we want to ensure that:
- there are crossmatches with the LSSG catalog configured in the config
- that the catalog name specified for it (the name of it in Mongo) does point to an existing catalog. That validation step is important to make the rest of the process easier to debug

retrieve the cross_matches for LSST in the enrichment worker. You can assume that there will be a list of LSSG crossmatches in there, now that it's configured. What ZTF has but LSST doesn't is that star property. So one thing you can add is a star property to the LSSTProperties we compute. For the babamul stuff you'll need to know a bit more than that. My suggestion is to make LSST match how things are handled in ZTF, i.e. there we have the sgscore and distpsnr (1, 2, 3 for the 3 nearest) directly in the candidate. That makes it super useful to make searches directly on the alert catalog and not have to depend on the aux table that has the crossmatches. So, long term we may want to have the enrichment worker add a candidate.sgscore  and candidate.distpsnr like fields that have the score and distance to the nearest object in LSSG , when that crossmatch info is available. Issue with that is that then our LsstCandidate schema is essentially outdated and such so I'd say don't go that route for the sake of babamul, it's not needed. For now, all we need for babamul is the data from these up to 3 nearest crossmatches (score+distance) to be propagated to the babamul stuff somehow. So, why not just add them to the EnrichedLSST struct that we pass to babamul, or as a separate parameter.

there, use that property to split into topics with the following logic:
- if you have matches, go through these if statements in order:
- if the nearest one is within say 1 arcsec and score > 0.5, it goes to the stellar topic
- if there is at least one match where score < 0.5, it goes to the hosted topic
- if none of the above was true, then it goes to hostless
- else (you don't have matches):
    - if it is in the footprint (I'll add that part, for now let's just fake that value. Feel free to set it to something static or randomize it), then it means there are no nearby objects and its just hostless
    - else (not in the footprint), then we don't know what that could be! So, we can assign it to some unknown topic.

PR #315, opened 2025-12-23
Fetch cross-matches in LSST enrichment worker to compute star/galaxy category from LSSG

petebachant:
Towards #292 

This also adds a part to the topic name denoting if an alert has a ZTF match or not, e.g., `babamul.ztf-match.stellar`.

## TODO

- [x] Fail if LSSG crossmatches are not configured or the catalog doesn't exist, so the app simply doesn't start with an invalid config for Babamul.
- [x] Properly compute `is_star`, or do the star logic in Babamul instead of LSST enrichment.
- [x] Test this, perhaps ingesting or creating some fake LSSG objects.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -56,7 +56,7 @@ services:
       timeout: 5s
       retries: 3
   broker:
-    image: apache/kafka:latest
+    image: apache/kafka:4.1.1
     hostname: broker
     networks:
       - boom
@@ -87,7 +87,7 @@ services:
       # controller metadata or internal admin ops can be unauthorized and cause the process to exit.
       # (Alternative: secure internal listener with SASL and drop ANONYMOUS.)
       KAFKA_SUPER_USERS: "User:admin;User:ANONYMOUS"
-      KAFKA_OPTS: -Djava.security.auth.login.config=/etc/kafka/kafka_server_jaas.conf
+      KAFKA_OPTS: "-Djava.security.auth.login.config=/etc/kafka/kafka_server_jaas.conf -Djava.security.egd=file:/dev/./urandom"
     volumes:
       - kafka_data:/var/lib/kafka/data
       - ./config/kafka_server_jaas.conf:/etc/kafka/kafka_server_jaas.conf:ro
@@ -99,9 +99,9 @@ services:
         timeout 5s /opt/kafka/bin/kafka-configs.sh --bootstrap-server broker:29092 --entity-type users --describe >/dev/null 2>&1 && \
         timeout 5s /opt/kafka/bin/kafka-acls.sh --bootstrap-server broker:29092 --list >/dev/null 2>&1"]
       interval: 10s
-      timeout: 25s
-      retries: 6
-      start_period: 30s
+      timeout: 30s
+      retries: 15
+      start_period: 60s
     labels:
       - autoheal=true
       - traefik.enable=true
@@ -113,7 +113,7 @@ services:
       # Uncomment to enable TLS passthrough (requires EXTERNAL listener to use SSL)
       # - traefik.tcp.routers.${STACK_NAME?Variable not set}-kafka.tls.passthrough=true
   kafka-acl-init:
-    image: apache/kafka:latest
+    image: apache/kafka:4.1.1
     hostname: kafka-acl-init
     depends_on:
       broker:
@@ -126,7 +126,12 @@ services:
     volumes:
       - ./scripts/init_kafka_acls.sh:/init_kafka_acls.sh:ro
     entrypoint: ["bash","/init_kafka_acls.sh"]
-    restart: "no"
+    restart: on-failure
+    deploy:
+      restart_policy:
+        condition: on-failure
+        delay: 5s
+        max_attempts: 3
   api:
     image: boom-astro/boom:latest
     hostname: api
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -59,7 +59,10 @@ jobs:
         EXIT_CODE=$(docker wait "$CID")
         echo "kafka-acl-init exited with code $EXIT_CODE"
         if [ "$EXIT_CODE" != "0" ]; then
+          echo "--- kafka-acl-init logs: ---"
           docker compose logs kafka-acl-init || true
+          echo "--- broker logs: ---"
+          docker compose logs broker || true
           exit 1
         fi
         echo "kafka-acl-init completed successfully."
--- a/docker-compose.override.yaml
+++ b/docker-compose.override.yaml
@@ -58,6 +58,12 @@ services:
     ports:
       - "9092:9092"
       - "9093:9093"
+    environment:
+      # Limit heap + use G1GC with fewer threads (auto-tuned based on heap size)
+      KAFKA_HEAP_OPTS: "-Xmx512M -Xms512M -XX:+UseG1GC -XX:MaxGCPauseMillis=200"
+      # Increase timeouts for slower CI environments
+      KAFKA_SOCKET_CONNECTION_SETUP_TIMEOUT_MS: 30000
+      KAFKA_SOCKET_CONNECTION_SETUP_TIMEOUT_MAX_MS: 60000
   api:
     env_file:
       - .env
--- a/scripts/init_kafka_acls.sh
+++ b/scripts/init_kafka_acls.sh
@@ -26,6 +26,36 @@ TIMEOUT=60 # seconds
 
 kafka_log() { echo "[init-kafka] $*"; }
 
+# Wait for broker hostname to be resolvable and port to be reachable
+# This handles Docker DNS/networking race conditions in CI
+wait_for_broker_network() {
+  local host="${BROKER%%:*}"  # Extract hostname from broker:port
+  local port="${BROKER##*:}"  # Extract port
+  local start=$(date +%s)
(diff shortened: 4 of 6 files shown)


Theodlz:
> @Theodlz this is still a draft, but requesting a review so you can make sure I'm going in the right direction with this.

Thanks Pete. I'll do my best to get to this soon!

petebachant:
This PR makes me think we should build a way to idempotently load in archival catalogs on startup. This way, we could declare that LSSG, NED, etc., should exist in the system. If they're missing, we load them in. If not, we skip that step.

petebachant:
I'm currently stuck trying to figure out where `score` might come from. I don't see it being added in the cross-matching, so I assume it's in the catalog. Can I download part of that somewhere?

Theodlz:
> I'm currently stuck trying to figure out where `score` might come from. I don't see it being added in the cross-matching, so I assume it's in the catalog. Can I download part of that somewhere?

we should require the following:
- LSST crossmatch config to be configured to crossmatch against the `LSSG` catalog (rip the naming scheme here, literally couldn't be more confusing)
- when we configure the crossmatch in the config, we have a `projection`, that will contain the schema. For LSSG, that's just called "score". BOOM adds a `distance_arcsec` field to crossmatch results so you can use that as the distance between the LSST object and the LSSG xmatch results.

Theodlz:
> This PR makes me think we should build a way to idempotently load in archival catalogs on startup. This way, we could declare that LSSG, NED, etc., should exist in the system. If they're missing, we load them in. If not, we skip that step.

I agree that we could make crossmatching easier to setup up, even at the cost of flexibility. But, do keep in mind that a catalog like LSSG has 3B+ entries and that can take hours to days to load (without even accounting for the download time, and availability of the data!). So, I don't think we should prioritize that until we figure out a little better which catalogs we care about.

src/enrichment/babamul.rs
```diff
@@ -62,6 +63,7 @@ impl EnrichedLsstAlert {
         cutout_template: Option<Vec<u8>>,
         cutout_difference: Option<Vec<u8>>,
         properties: LsstAlertProperties,
+        cross_matches: Option<std::collections::HashMap<String, Vec<CrossMatch>>>,
```
Theodlz: That doesn't make sense to me. Crossmatches is a document (here hashmap) with the name of a catalog as key, and an array of matches as value. Each catalog as a different schema. So `HashMap<String, Vec<CrossMatch>>` is incorrect because it says: "all catalogs have the same schema".

Theodlz: Until we replace our flexible xmatch system by some strongly typed one, I'd suggest just sticking to that (hashmap of values or documents)

src/enrichment/babamul.rs
```diff
@@ -74,6 +76,53 @@ impl EnrichedLsstAlert {
...
+                    // Rule 1: Check if nearest match is within distance threshold and score > 0.5
+                    if let Some(nearest) = lssg_matches.first() {
+                        // Convert distance from radians to arcseconds if needed
+                        let distance_arcsec = if nearest.distance > 1.0 {
+                            // Assume already in arcseconds
+                            nearest.distance
+                        } else {
+                            // Assume in radians, convert to arcseconds
+                            nearest.distance * 206265.0
+                        };
```
Theodlz: I really don't get this piece of code? If distance > 1 you use it as is and otherwise convert from radians to arcsec? 

All crossmatches we compute (and you can see that in spatial.rs) gets a `distance_arcsec` field added. So we should use that, no flexible conversion thing should be needed

src/enrichment/babamul.rs
```diff
@@ -74,6 +76,53 @@ impl EnrichedLsstAlert {
             cutout_template: cutout_template.map(CutoutBytes),
             cutout_difference: cutout_difference.map(CutoutBytes),
             survey_matches: alert.survey_matches,
+            cross_matches,
+        }
+    }
+
+    pub fn compute_babamul_category(&self) -> String {
+        const DISTANCE_THRESHOLD_ARCSEC: f64 = 1.0;
+        const SCORE_THRESHOLD: f64 = 0.5;
```
Theodlz: best to put constants at the top of the file

tests/config.test.yaml
```diff
@@ -63,7 +63,14 @@ workers:
...
 crossmatch:
-  lsst: []
+  lsst:
+    - catalog: LSSG
+      radius: 2.0 # 2 arcseconds
+      use_distance: false
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
```
Theodlz: Perhaps this PR isn't synced with main (where LSSG is already in config) but it should be configured like this instead:
```yaml
  lsst:
    - catalog: LSSG
      radius: 30.0 # 30 arcseconds
      use_distance: false
      projection:
        _id: 1
        ra: 1
        dec: 1
        score: 1
        magwhite: 1
      max_results: 3
```

Theodlz (changes requested):
I left some comments Pete, just to help guide the development a bit

src/enrichment/babamul.rs
```diff
@@ -8,14 +8,87 @@ use crate::enrichment::{EnrichmentWorkerError, LsstPhotometry, ZtfPhotometry};
...
+            None
+        }
+        Err(e) => {
+            tracing::warn!("Failed to parse footprint MOC: {}", e);
+            None
+        }
+    }
+}
+
+fn is_in_footprint(ra_deg: f64, dec_deg: f64) -> bool {
```
petebachant: This might be a little too robust. For example, f we fail to load the FITS file, we simply always get `false`.

src/enrichment/babamul.rs
```diff
@@ -8,14 +8,75 @@ use crate::enrichment::{EnrichmentWorkerError, LsstPhotometry, ZtfPhotometry};
...
+impl apache_avro::schema::derive::AvroSchemaComponent for CrossMatchesWrapper {
+    fn get_schema_in_ctxt(
+        _named_schemas: &mut HashMap<apache_avro::schema::Name, apache_avro::schema::Schema>,
+        _enclosing_namespace: &apache_avro::schema::Namespace,
+    ) -> apache_avro::Schema {
+        // Return null schema since this field is never serialized
+        apache_avro::Schema::Null
+    }
+}
+
```
Theodlz: Nice

src/enrichment/lsst.rs
```diff
@@ -150,6 +155,7 @@ pub struct LsstAlertForEnrichment {
 pub struct LsstAlertProperties {
     pub rock: bool,
     pub stationary: bool,
+    pub star: bool,
```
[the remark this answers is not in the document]
Theodlz: Also, we can't make that decision if its outside of the footprint

src/enrichment/lsst.rs
```diff
@@ -16,6 +17,8 @@ use schemars::JsonSchema;
 use std::collections::HashMap;
 use tracing::{error, instrument, warn};
 
+pub const IS_STELLAR_DISTANCE_THRESH_ARCSEC: f64 = 1.0;
```
Theodlz: I will bet that we'll need to change that value later, to account for the fine spatial resolution of both LSST and Legacy Survey (that was used to build that catalog)

src/enrichment/lsst.rs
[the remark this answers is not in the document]
Theodlz: the one we may chance quicker is the score threshold. We should read the paper of that catalog and see what cut they recommend. Let me see if I can figure that out now...

the key is not labeling something as stellar if it isn't (most folk using the alerts will do non-stellar science I bet, so they'd rather get a star incorrectly than missing a non-star transient we labelled as stellar)

src/enrichment/lsst.rs
```diff
@@ -5,17 +5,63 @@ use crate::enrichment::{
...
+            panic!("Unexpected MOC type in footprint MOC file");
+        }
+        Err(e) => {
+            panic!("Failed to parse footprint MOC: {}", e);
+        }
+    }
+}
+
+pub fn is_in_footprint(ra_deg: f64, dec_deg: f64) -> bool {
+    let moc = FOOTPRINT_MOC.get_or_init(load_footprint_moc);
```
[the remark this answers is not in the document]
Theodlz: we do want the app to crash instead of sending "false", which could lead to incorrect states

tests/test_babamul.rs
```diff
@@ -188,6 +267,188 @@ async fn delete_kafka_topic(topic: &str, config: &AppConfig) {
...
+        false,
+        None, // No LSPSC cross-matches
+        survey_matches,
+    );
+    let category = alert_ztf_unknown.compute_babamul_category();
+    assert_eq!(
+        category, "ztf-match.hostless",
+        "Alert with ZTF match but no LSPSC and in footprint should be ztf-match.hostless"
+    );
+}
```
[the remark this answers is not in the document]
Theodlz: that's a great point, will add

docker-compose.yaml
```diff
@@ -241,6 +241,7 @@ services:
     volumes:
       - ${PWD}/config.yaml:/app/config.yaml
       - ${PWD}/data/models:/app/data/models
+      - ${PWD}/data/ls_footprint_moc.fits:/app/data/ls_footprint_moc.fits:ro
```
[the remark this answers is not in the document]
Theodlz: done

PR #317, opened 2026-01-08
Tune Kafka options for CI checks

petebachant:
Trying to eliminate some of the flakiness we have in the Kafka initialization on the GitHub Actions runners.


petebachant:
Welp, first try failed...

PR #319, opened 2026-01-09
Lightcurve linear fitting (for rise and fade rates): weighted and centered

Theodlz:
In this PR, we:
- replace the simple linear fitting we have been doing so far, by a weighted least square method that makes use of the error bars (with centering for numerical stability).
- compute a reduced chi2 and slope error, in place of the r2

Performance is naturally slower (half as fast as the previous version I believe), but on 1k points we are still only talking about 3.4 micro seconds, which is absolutely nothing compared to the rest of the compute done in the enrichment worker, making it negligible anyways. But, the CI may confirm or deny that claim!

For context, we looked at what fancier (and well established) fast transient pipelines are doing, namely ZTFReST. They do a least square with errors accounted for as well. That's what we were missing so far (accounting for the errors) and that this PR is adding. Also, they were using scipy's fitting method for their linear equation, which uses a Levenberg–Marquardt algorithm. That algorithm uses an iterative process for fitting, which only makes sense for a non-linear equation were you don't have closed-form solution. But for the linear case, you do, and therefore it shouldn't add anything other than computational complexity. So, we do not use this here.

### DIFF
--- a/src/utils/lightcurves.rs
+++ b/src/utils/lightcurves.rs
@@ -436,6 +436,86 @@ mod tests {
     }
 
     #[test]
+    fn test_weighted_least_squares_centered() {
+        // Test case 1: simple linear data (approximately y = 2x) with small errors
+        // => expecting a good fit
+        let x = vec![1.0, 2.0, 3.0, 4.0, 5.0];
+        let y = vec![2.0, 4.1, 6.0, 8.1, 10.2];
+        let sigma = vec![0.1, 0.1, 0.1, 0.1, 0.1];
+
+        let result = weighted_least_squares_centered(&x, &y, &sigma).unwrap();
+        assert!((result.rate - 2.04).abs() < 1e-2);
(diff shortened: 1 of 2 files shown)


src/enrichment/ztf.rs
```diff
@@ -177,7 +178,7 @@ where
                 .filter_map(|p| {
                     ZtfPhotometry::try_from(p)
                         .map_err(|e| {
-                            warn!(
+                            debug!(
                                 "Failed to convert ZtfForcedPhotometry to ZtfPhotometry: {}",
                                 e
                             );
```
antoine-le-calloch: Thanks 😆

antoine-le-calloch (approved):
LGTM

src/enrichment/ztf.rs
Theodlz: Just made that change, we now skip bad procstatus values (as intended) but show a warning for any other - unexpected - errors

PR #322, opened 2026-01-14
LSST enrichment: add multisurvey_photstats

Theodlz:
- add multisurvey_photstats when enriching lsst alerts, using ZTF photometry if there is a match (combining LCs from both, as we already do when we enrich ZTF alerts)
- always set a multisurvey_photstats, whether we have matches with other surveys or not (if we do, we combine lightcurves and compute new stats. if we don't, we just copy the single-survey photstats). That way, a user that is interested in using multisurvey data can use that field for a filter and does not need to use the regular OR multisurvey photstats. This should help keep the filters simple. In the future, we may want to add some field that indicates what surveys were involved in computing the provided multisurvey photstats

### DIFF
--- a/src/enrichment/lsst.rs
+++ b/src/enrichment/lsst.rs
@@ -434,8 +434,9 @@ impl LsstEnrichmentWorker {
         prepare_photometry(&mut lightcurve);
         let (photstats, _, stationary) = analyze_photometry(&lightcurve);
 
-        // make a multisurvey lightcurve if we have matches with other surveys (ZTF for now)
-        let multisurvey_photstats = if let Some(survey_matches) = &alert.survey_matches {
+        // Compute multisurvey photstats (including ZTF if available, other surveys can be added later)
+        let mut has_matches = false;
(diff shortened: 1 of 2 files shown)


src/enrichment/lsst.rs
```diff
@@ -433,11 +434,38 @@ impl LsstEnrichmentWorker {
         prepare_photometry(&mut lightcurve);
         let (photstats, _, stationary) = analyze_photometry(&lightcurve);
 
+        // Compute multisurvey photstats (including ZTF if available, other surveys can be added later)
```
petebachant: If this is extended later, will the schema need to change? Maybe it would be better to have something like this so it's purely an extension and not a change down the road:

```rust
pub struct MultiSurveyPhotStats {
    pub ztf: PerBandProperties
}
```

petebachant: Gotcha. So the users will know what surveys' data made it into these combined stats by looking at `survey_matches`.

PR #323, opened 2026-01-14
Split ZTF Babamul topics

### LINKED ISSUE #292: Properly split Babamul alerts into multiple streams

petebachant:
Closes #292 

## TODO

- [x] Implement hosted/hostless categorization
- [x] Ensure this is properly tested

### DIFF
--- a/tests/test_babamul.rs
+++ b/tests/test_babamul.rs
@@ -623,7 +623,10 @@ async fn test_babamul_process_ztf_alerts() {
     let alert2 = create_mock_enriched_ztf_alert(1234567891, &ztf_obj2, false);
 
     // Process the alerts
-    let _result = babamul.process_ztf_alerts(vec![alert1, alert2]).await;
+    let _result = babamul
+        .process_ztf_alerts(vec![alert1, alert2])
+        .await
+        .unwrap();
 
     // Consume messages from Kafka topic and verify our specific alerts are present
     let messages = consume_kafka_messages(topic, &config).await;


src/enrichment/babamul.rs
```diff
@@ -202,6 +202,33 @@ impl EnrichedZtfAlert {
...
+
+        // Use star-galaxy score to determine if hosted
+        // sgscore1 <= 0.5 suggests galaxy-like, indicating the alert is likely hosted
+        let sgscore1 = self.candidate.candidate.sgscore1.unwrap_or(1.0);
+        if sgscore1 <= 0.5 {
+            return category + "hosted";
+        }
+
+        // Otherwise, we can't definitively classify it
+        category + "unknown"
```
petebachant: @Theodlz can you check this logic for me?

Theodlz: the catalog that the SGSCORE comes from, has - to my knowledge - full coverage of the ZTF footprint. There may be some small exceptions when we get closer to dec of -30 (the southern bound of the footprint), but overall we should have coverage for all this. So, I don't think we should have an `unknown` category here. 

It should be:
- if the enrichment worker flagged it as star-like, report as `stellar`
- otherwise, if we have sgscore1, or sgscore2, or sgscore3 (the 3 nearest objects in the sgscore catalog up to 30 arcsec, see [here](https://zwickytransientfacility.github.io/ztf-avro-alert/schema.html)) < 0.5, let's flag it as `hosted`. **CAREFUL** about the values here. Let's also make sure the scores are >= 0. I believe ZTF's pipeline may use -99 or some other crap as the "there is no value caus no match" placeholder.
- and otherwise, flag as `hostless`.

Notes:
- Let's do a bit of research to check if there is some footprint definition we need to worry about ala LSPSC, which means there would be an unknown category to determine. But as is, the default category for ZTF should be hostless.
- In the future, we will want to revisit this to add LSPSC to the mix also for ZTF. So if we have no sgscore values, we check if LSPSC may have some info.

PS: None of these 2 notes need to be implemented/considered in this first PR.

tests/test_babamul.rs
```diff
@@ -996,7 +1103,12 @@ async fn test_babamul_ztf_with_lsst_match() {
     let db = boom::conf::get_test_db().await;
     let config = AppConfig::from_path(TEST_CONFIG_FILE).unwrap();
     let mut ztf_alert_worker = boom::utils::testing::ztf_alert_worker().await;
-    delete_kafka_topic("babamul.ztf.none", &config).await;
+
+    // ZTF alert with LSST match could go to any lsst-match.* topic depending on properties
+    // Clean up potential topics before test
+    delete_kafka_topic("babamul.ztf.lsst-match.stellar", &config).await;
```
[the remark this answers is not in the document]
Theodlz: you can leave this here but just a heads up

src/enrichment/babamul.rs
```diff
@@ -219,15 +219,25 @@ impl EnrichedZtfAlert {
...
+            self.candidate.candidate.sgscore1,
+            self.candidate.candidate.sgscore2,
+            self.candidate.candidate.sgscore3,
+        ];
+
+        for score in sgscores.iter().flatten() {
+            // Only consider valid scores (>= 0)
+            if *score >= 0.0 && *score < 0.5 {
+                return category + "hosted";
+            }
```
Theodlz: Nice, I see that the assumption here - which is correct - is that the enrichment worker already flagged something as a star, if it is one. Makes me think that the babamul lsst would benefit from that too (i.e. if enrichment worker flagged as star then no need to check that again), to avoid duplicating the star-association logic. anyways, that's its own PR I think.

Just one comment: better make that score threshold a const or something, so we have it defined in one  - immutable - place and we can easily track that. Also wondering if 0.5 is the best cut here, maybe something we should ask the more science-included members of the team, or even some users! But, let's stick with that for now

PR #333, opened 2026-01-22
Babamul API: objectId partial search endpoint

Theodlz:
This PR adds an endpoint to search for objects (simply returning id, ra, dec) by - partial - objectId. Based on the objectId passed to the endpoint, we do our best to:
- check if it is "valid", with some regex logic to see if it could match LSST or ZTF
- infer what survey it aims to search for
- format it properly (so we don't need to do a case insensitive search on the DB which is not possible performance wise)
- find the first `limit` objects that match that partial id, sorted (limit can be between 1 and 100).

We also add unit tests for that endpoint.

Note: For context, this is super useful to support the web app, and let folks search for objects as they type! It's mostly for ZTF since LSST ids are absolutely unreadable and meaningless... But for ZTF or any other survey we may support in the future that uses human readable names, we know from experience on Fritz that this is something that people love to have! It's a very big UX improvement over an exact text search.

### DIFF
--- a/src/api/routes/babamul/surveys/objects.rs
+++ b/src/api/routes/babamul/surveys/objects.rs
@@ -479,6 +479,10 @@ fn infer_survey_from_objectid(value: &str) -> Result<(Survey, String), String> {
 
     // ZTF with complete prefix: only accept full "ZTF" when followed by digits/letters
     let ztf_prefix_re = get_ztf_prefix_regex();
+    println!(
+        "Trying to match ZTF prefix regex against: {} (regex: {:?})",
+        upper, ztf_prefix_re
+    );
     if let Some(caps) = ztf_prefix_re.captures(&upper) {
         let digits = caps.get(1).unwrap().as_str();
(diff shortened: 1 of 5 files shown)


src/api/routes/babamul/surveys/objects.rs
```diff
@@ -440,3 +442,187 @@ pub async fn get_object(
...
+        ("limit" = Option<u32>, Query, description = "Maximum number of results to return (1-100, default 10)"),
+    ),
+    responses(
+        (status = 200, description = "Search results", body = Vec<SearchObjectResult>),
+        (status = 400, description = "Invalid objectId format"),
+        (status = 500, description = "Internal server error")
+    ),
+    tags=["Surveys"]
+)]
+#[get("/babamul/surveys/objects/search/{value}")]
```
petebachant: This endpoint could be simply:

```suggestion
#[get("/babamul/objects")]
```

and use a query param, e.g., `id_like`, instead of the path param `value`. I forget if we already have an objects endpoint, but if so, this functionality could be merged into that.

PR #347, opened 2026-01-28
ZTF Enrichment Worker: skip MLing if features are missing

Theodlz:
If some ML features can't be computed, skip MLing ZTF alerts in the enrichment worker (and only add properties), instead of panicking which right now kills the worker. Besides, it's actually normal for these features to be missing for old alerts (say pre 2020), when a bunch of fields weren't in the schema yet.

For context: this created issues when I imported a bunch of old alerts from Kowalski, for which these models simply cannot run on by design, and that's ok.

EDIT: looks like we aren't updating the `updated_at` field of the alert when we enrich it, which I think we should be. This PR also adds that.

### DIFF
--- a/src/enrichment/decam.rs
+++ b/src/enrichment/decam.rs
@@ -149,6 +149,8 @@ impl EnrichmentWorker for DecamEnrichmentWorker {
             return Ok(vec![]);
         }
 
+        let now = flare::Time::now().to_jd();
+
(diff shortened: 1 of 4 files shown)


petebachant (approved):
LGTM. Do we do any ML unit testing? If so, you could add this case to it.

PR #349, opened 2026-01-28
Expand Babamul stream documentation

### LINKED ISSUE #334: Show how alerts from both LSST and ZTF will only go into one topic, i.e., they don't overlap

petebachant:
- Reword initial discovery by LSST
- Add diagram for ZTF topics to show all 14, and how any one alert only ends up in one topic (no overlap)

Resolves #334

### DIFF
--- a/docs/babamul.md
+++ b/docs/babamul.md
@@ -135,12 +135,12 @@ flowchart TD
     LSST[New LSST Alert] --> CheckLSPSC{Has matches<br/>in LSPSC?}
 
     CheckLSPSC -->|No| CheckFootprint{In LSPSC<br/>footprint?}
-    CheckFootprint -->|Yes| LSST_Hostless[LSST Hostless]
     CheckFootprint -->|No| LSST_Unknown[LSST Unknown]
+    CheckFootprint -->|Yes| LSST_Hostless[LSST Hostless]
 
-    CheckLSPSC -->|Yes| CheckLSSTStellar{Any match with<br/>distance ≤ 1.0″<br/>and score > 0.5?}
+    CheckLSPSC -->|Yes| CheckLSSTStellar{Any stellar match?}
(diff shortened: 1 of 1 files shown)


Theodlz (commented):
I have 4 comments, all about the  diagram(s) that show the suffix assignment (stellar, ...):
- There is some unbalance here, between the LSST and ZTF plots. For LSST you explain precisely how the flags are set, whereas for ZTF you just say "flag from enrichment worker". It would be best to either have the exact cuts in the diagram or omit them and then list them under the diagram for reference (like some kind of footnote).
- The Stellar logic is outdated for LSST I think? That's on me, I believe I added the `is_near_brightstar` property after your initial docs PR, without updating the diagram. I'm happy to take on that bit of the documentation update, if you want me to.
- We now have 3 diagrams. There is significant overlap between the first and the second (both are LSST specific, but the second has more detail to it. Wondering if the first one should be replaced by some survey agnostic thing that's oversimplified, to give an idea of what the flow looks like when we don't get into the survey-specific implementation details.
- for the `LSST alerts` plot, I think that the `unknown` branch should swap spots with the `hostless` one. That would avoid some arrows from overlapping

Theodlz:
> I made the bottom two diagrams more generic w.r.t. how exactly stellar, hosted, hostless are computed, since it will be a pain to keep these in sync with the code. It's open source, so if people really want to know the exact logic, they can go there. The purpose of this diagram is to help people decide what streams to listen to.
> 
> Re the top diagram, I could call them "survey 1" and "survey 2", but that might be a bit too abstract for some. I'm not sure.

Unfortunately I really doubt folks will even think about looking in the code, but will want to know about the exact cuts. I say for now we ignore them since we are refining these, and suggest (should be somewhere in plaintext) looking into the code. When we settle on something more stable, let's put that back in the diagrams.

PR #350, opened 2026-01-29
Add Babamul personal access tokens (PATs)

### LINKED ISSUE #346: Add Babamul personal access tokens so users can securely interact with the API
Will be nice for the Python client to be able to do typical API stuff without needing plaintext usernames and passwords sitting around.

petebachant:
Resolves #346

Users can then use these to make requests with the API in lieu of their username and password.

## TODO

- [x] Add delete and/or deactivate functionality
- [x] Test auth with PAT

### DIFF
--- a/src/api/auth.rs
+++ b/src/api/auth.rs
@@ -290,58 +290,32 @@ pub async fn babamul_auth_middleware(
                 // Look up the token and update last_used_at in a single atomic operation
                 // Use aggregation pipeline to update token and join with user in one operation
                 let now = flare::Time::now().to_utc().timestamp();
-                let tokens_collection: mongodb::Collection<
-                    crate::api::routes::babamul::tokens::BabamulUserToken,
-                > = db_app_data.collection("babamul_user_tokens");
 
-                match tokens_collection
-                    .find_one(doc! { "token_hash": &token_hash })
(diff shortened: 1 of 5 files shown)


Theodlz:
> > I say we keep it for now and can decide to revisit later (easily dropped, as you mentioned), how does that sound?
> 
> Might not be so straightforward to update tokens if they are an array in the `babamul_user` model though, just something to keep in mind. I'm sure it's doable but I just wonder what the query looks like!

If it turns out that it ain't easy to do both, I say we drop the `last_used_at` field then. Just not worth it to do 2 reads + 1 write vs 1 read, for a feature I am not even sure we will make use of.

Theodlz:
> Ah one last thing! @petebachant. You may be able to combine the 2 reads in one (token collection and user collection) and get around these issues just by running an aggregation pipeline. I even wonder if you could do 1 query that does it all! (match + update + return the user)

Here's what I had in mind: https://www.mongodb.com/docs/manual/reference/method/db.collection.findoneandupdate/

petebachant:
If this ends up being too slow, we could add an in-memory cache for Babamul user structs keyed by hashed PAT with a short-ish TTL like 10 minutes.

PR #352, opened 2026-02-03
PATs on Babamul user collection

Theodlz:
This PR is an attempt at storing the PATs as an embedded array of the user collection, rather than as a separate one. It gives us these advantages:
- since we aren't in a relational DB, deleting a user in the current approach would leave the tokens in the other collection
- now when authenticating, instead of using 2 to 3 queries (reads/writes to the babamul_users and babamul_user_tokens collections) we only need one: a findOneAndUpdate that finds the user with that token hash if any , updates the last_used_at field and returns the user all in one DB round-trip
- Handlers like the get_tokens become very easy to implement, since we already get the `current_user` which contains the tokens.

BTW this is just my take on it! If the reviewer sees something wrong about that approach, happy to not take that road.

### DIFF
--- a/src/api/auth.rs
+++ b/src/api/auth.rs
@@ -290,58 +290,32 @@ pub async fn babamul_auth_middleware(
                 // Look up the token and update last_used_at in a single atomic operation
                 // Use aggregation pipeline to update token and join with user in one operation
                 let now = flare::Time::now().to_utc().timestamp();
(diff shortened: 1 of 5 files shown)


petebachant (approved):
This looks good to me. You could drop the ID field if you want, though it's not going to break the bank storage or network-wise IMO.

PR #359, opened 2026-02-08
Fix issue 111

ThatSoulyGuy:
This PR introduces a complete GCN (Gamma-ray Coordinates Network) event crossmatching subsystem into BOOM. It enables
  real-time spatial matching between incoming astronomical alerts (ZTF, LSST) and active transient events from missions
  such as LVK, Swift, Fermi, SVOM, Einstein Probe, and IceCube, as well as user-defined watchlist positions.

  Architecture

  The implementation is organized as a self-contained gcn module (src/gcn/) with four submodules:

  - models.rs — Core data types: GcnEvent, EventMatch, EventGeometry (tagged enum supporting both circular regions and
  HEALPix skymaps), GcnSource (7 mission variants), GcnEventType (5 classification variants), and WatchlistError.
  HEALPix geometry uses cdshealpix for nested pixel indexing with O(1) containment checks via HashSet. Constructors
  enforce invariants at creation time (nside must be a power of 2, pixel/probability arrays must be equal length).
  - cache.rs — ActiveEventsCache: a thread-safe in-memory cache of non-expired active events, backed by MongoDB. Uses a
  combined RwLock<CacheState> for atomic reads and a separate Mutex to prevent concurrent refresh operations (TOCTOU
  protection). The cache is shared across async alert workers and auto-refreshes on a configurable interval.
  - xmatch.rs — event_xmatch() (async) and event_xmatch_sync(): the crossmatch functions that take an alert's (RA, Dec,
  JD) and return all matching EventMatch records. Each match includes angular separation (for circular events) or pixel
  probability (for HEALPix events), plus the trigger offset in days.
  - tests.rs — 127 unit tests organized across four test modules covering models, watchlist validation, crossmatch
  logic, and configuration parsing.

  Input Validation and Security

  - Watchlist names are trimmed and capped at 256 characters; descriptions at 2048
  - Coordinate validation: RA in [0, 360], Dec in [-90, 90], NaN/Infinity rejected
  - Radius bounds enforced against configurable maximum (default 10 degrees)
  - Expiration capped at 365 days
  - All database errors return generic messages to the client; actual errors are logged server-side via tracing::error!
  to prevent information leakage
  - Watchlist ownership is enforced on all mutation endpoints

  Alert Worker Integration

  Both ZtfAlertWorker and LsstAlertWorker now import EventMatch, preparing the pipeline for event match storage on alert
   auxiliary documents. The event_matches field on aux objects will hold a Vec<EventMatch> populated during alert
  ingestion.

  Database Changes

  New MongoDB indexes on the gcn_events collection:
  - { "coordinates.radec_geojson": "2dsphere" } — geospatial queries
  - { "source": 1, "is_active": 1, "expires_at": 1 } — active event lookups
  - { "trigger_time": 1 } — temporal queries
  - { "user_id": 1 } (sparse) — watchlist ownership lookups

  Configuration

  Full configuration hierarchy under gcn: in the application config, with sensible defaults:
  - Consumer settings (client ID/secret, Kafka topics)
  - Per-source expiration periods (e.g., LVK: 30 days, Swift: 7 days)
  - Backfill lookback/lookahead windows
  - Cache refresh interval
  - Watchlist limits (max entries per user, max radius)

  All fields support environment variable overrides via the existing BOOM_ prefix convention.

  Test Coverage

  127 tests covering:
  - Serialization roundtrips for all model types (serde snake_case, MongoDB _id mapping)
  - Spherical geometry: containment at poles, RA wraparound, zero/sub-arcsecond/large radii
  - HEALPix: pixel containment, probability lookup, invalid nside rejection, empty maps
  - Event matching: active/inactive/expired states, exact boundary conditions, trigger offsets
  - Crossmatch: empty inputs, single/multiple matches, concentric regions, ordering, scaling to many events
  - Watchlist validation: coordinate bounds, NaN/Infinity rejection, radius limits
  - Configuration: defaults, partial overrides, full deserialization for all config structs

### DIFF
--- a/src/api/routes/events.rs
+++ b/src/api/routes/events.rs
@@ -1,6 +1,6 @@
 use crate::api::models::response;
 use crate::api::routes::users::User;
-use crate::gcn::{EventGeometry, GcnEvent};
+use crate::gcn::{EventGeometry, GcnEvent, GcnSource};
 use actix_web::{get, web, HttpResponse};
 use futures::stream::StreamExt;
 use mongodb::bson::{doc, Document};
@@ -12,11 +12,11 @@ use utoipa::{IntoParams, ToSchema};
 /// Query parameters for listing events
 #[derive(Debug, Deserialize, IntoParams)]
 pub struct EventListQuery {
-    /// Filter by source (e.g., "lvk", "swift", "custom")
+    /// Filter by source (e.g., "lvk", "swift") — excludes "custom" watchlist events
     pub source: Option<String>,
     /// Only show active events (default: true)
     pub active_only: Option<bool>,
-    /// Maximum number of results (default: 100)
+    /// Maximum number of results (default: 100, max: 1000)
     pub limit: Option<i64>,
     /// Skip this many results (for pagination)
     pub skip: Option<u64>,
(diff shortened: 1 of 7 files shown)


ThatSoulyGuy:
> @ThatSoulyGuy can we split the PR between the user defined list code and what we get from GCN? We would like to see the user list go in first and should be simpler code.
> 
> For GCN, I think we will want to think about which missions (with tiny localizations) we want to ingest and how. I don't imagine we will do LVK and Fermi here, but instead EP and Swift and the like.

Sure, I'll make it happen

Theodlz:
Could we perhaps get on zoom one of these days and discuss both of these features (watchlists and GCN matching)? @ThatSoulyGuy @mcoughlin

PR #367, opened 2026-02-11
Babamul: Lighter alerts

Theodlz:
This PR:
- removes the cutouts from the Babamul alerts, which on average reduces ZTF alert size by 50% (we expect a more significant improvement for LSST, as its alerts' sizes aren't dominated by a long history of detections/non-detections/fp)
- we make the schema of the lightcurves as lightweight as possible. We now have 3 schemas (one for the detection history, one for the non detection history, and one for the forced phot history) that have the very bare minimum number of fields neeed. Before, we had duplication (both mag and flux space data, when we can easily recuperate the mag using the flux and without needing one zero point per data point, once PR #362 is merged), and useless fields (like programid for ZTF, when it's litteraly always 1; or ra and dec for non detections and forced phot where they simply don't exist, ...).
- move from snappy to deflate as the compression codec. While testing locally I noticed that the serialize/deserialize time was identical (at least, in Python) but the compression was ~5% better with deflate (tested over thousands of alerts). We may need to tune the `DeflateSettings` to get the "right" compression level, but it's clear that it's better than snappy.
- update unit tests to reflect these changes.
- only send ZTF alerts where drb > 0.2; set the LSST reliability cut to 0.0  (i.e. send all alerts) until we have a better understanding of what's a good cut, with the new model that is used now.

~Caution: This PR depends on #362, so let's not merge it until that other PR is done and merged to main~

### DIFF
--- a/tests/test_babamul.rs
+++ b/tests/test_babamul.rs
@@ -922,7 +922,7 @@ async fn test_babamul_filters_low_drb() {
     alert_no_drb.candidate.candidate.drb = None;
 
     let result = babamul
-        .process_ztf_alerts(vec![alert_low_drb, alert_zero_drb, alert_no_drb])
+        .process_ztf_alerts(vec![alert_low_drb, alert_no_drb])
         .await;
     assert!(
         result.is_ok(),


src/enrichment/babamul.rs
```diff
@@ -21,88 +21,89 @@ use tracing::{info, instrument};
...
+pub struct AlertPhotometry {
+    pub jd: f64,
+    #[serde(rename = "psfFlux")]
+    pub flux: f64, // in nJy
+    #[serde(rename = "psfFluxErr")]
+    pub flux_err: f64, // in nJy
+    pub band: Band,
+    pub ra: f64,
+    pub dec: f64,
 }
```
antoine-le-calloch: Here, `flux: f64`, `ra: f64`, and `dec: f64` are non-optional, but the base structs ZtfPhotometry and LsstPhotometry define these fields as optional.

petebachant (approved):

src/enrichment/babamul.rs
Theodlz: Will take a look. But some of the structs are used to read both detections and non detections from the DB, whereas `AlertPhotometry` is only detections hence the non-optional flux and positions. There is another struct (NonDetectionPhotometry) for the optional case.

Theodlz:
@petebachant @antoine-le-calloch I just rerequested a review since I made a couple of changes since the last review. Let me know what you think :)

src/enrichment/babamul.rs
```diff
@@ -17,93 +17,231 @@ use rdkafka::client::DefaultClientContext;
...
+    #[serde(rename = "psfFlux")]
+    pub flux: Option<f64>, // in nJy
+    #[serde(rename = "psfFluxErr")]
+    pub flux_err: f64, // in nJy
+    pub band: Band,
+}
+
+#[serdavro]
+#[derive(serde::Deserialize, serde::Serialize, Debug, Clone)]
+pub struct BabamulSurveyMatch {
```
petebachant: Nit: Instead of naming these with a Babamul prefix, you could import and use them with their namespace, e.g., `babamul::SurveyMatch`.

petebachant (approved):
Minor stuff, nothing blocking a merge.

src/enrichment/babamul.rs
Theodlz: Good point. I think I just wanted to differentiate between these and the structs from the enrichment worker. Some of our structs end up being exposed through let's say the API, so having unique names seemed a little better to me.

src/enrichment/babamul.rs
```diff
@@ -164,56 +300,77 @@ impl BabamulEnrichedLsstAlert {
 /// Enriched ZTF alert
 #[serdavro]
 #[derive(Debug, serde::Deserialize, serde::Serialize)]
-pub struct BabamulEnrichedZtfAlert {
+pub struct BabamulZtfAlert {
```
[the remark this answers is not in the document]
Theodlz: Ultimately, these don't represent the enriched alerts. We only add some of the data products in here  modified lightweight lightcurve schema, no images, no classifications, ... Earlier the Babamul schema was basically the alert + all the enrichmrnt results, but this isn't the case anymore. 

For the end user, this struct becomes the alert schema and it's just noise to call our babamul alert schema "Babamul Enriched [survey] Alert" and not just "Babamul [survey] Alert". Calling things enriched isn't very relevant for the Kafka stuff which is what this feeds. 

+ Babamul users don't need and don't care about knowing that this is "the result of the enrichment worker" (which anyway it isn't anymore in this PR, it's really just a babamul-only struct), which is what this enriched keyword was meant to represent originally.

PR #371, opened 2026-02-13
Fix benchmark

antoine-le-calloch:
- Fix benchmark
- Improve Readme with benchmark instructions.


Theodlz:
> What's the motivation behind changing where the NED catalog is downloaded? Was that breaking something that wasn't an issue on GitHub Actions?

same question

PR #376, opened 2026-02-19
Migrate ZTF forced photometry flux values to a fixed zeropoint.

antoine-le-calloch:
Migration binary that rescales ZTF forced photometry psfFlux/psfFluxErr from per-datapoint magzpsci zeropoints to the fixed ZTF_ZP=23.9, using batched update_many. Safe to re-run (old values saved as _old_* fields, already-migrated docs are skipped), with --cleanup to remove the _old fields after verification.

### DIFF
--- a/src/bin/migrate_fp_flux.rs
+++ b/src/bin/migrate_fp_flux.rs
@@ -11,23 +11,19 @@ const ZTF_ZP: f64 = 23.9;
 
 /// Migrate ZTF forced photometry flux values to a fixed zeropoint.
 ///
-/// Rescales `psfFlux` and `psfFluxErr` in `fp_hists` from per-datapoint
-/// `magzpsci` zeropoints to the fixed ZTF_ZP = 23.9 zeropoint.
+/// Recomputes `psfFlux` and `psfFluxErr` in `fp_hists` from the raw IPAC
+/// `forcediffimflux` and `forcediffimfluxunc` fields, converting to nJy at
+/// the fixed ZTF_ZP = 23.9 zeropoint.
 ///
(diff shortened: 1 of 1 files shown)


Theodlz (changes requested):
Thanks a lot Antoine, I really appreciate the help.

I do see one issue with this script. I can't run it twice, because basically psfFlux is computed as old pstFlux * factor. So it would keep increasing that value.

It would be a lot better if it was computed from the raw flux and fluxerr reported by IPAC, like we do in the rust code. That way we can run this update as many times as we want and it won't cascade into increasing that value more and more.

Theodlz:
> Thanks a lot Antoine, I really appreciate the help.
> 
> I do see one issue with this script. I can't run it twice, because basically psfFlux is computed as old pstFlux * factor. So it would keep increasing that value.
> 
> It would be a lot better if it was computed from the raw flux and fluxerr reported by IPAC, like we do in the rust code. That way we can run this update as many times as we want and it won't cascade into increasing that value more and more.

In pratice, that just means using the same logic (for conversion and validation) as what's in here: https://github.com/boom-astro/boom/blob/main/src/alert/ztf.rs#L255-L295

antoine-le-calloch:
> Thanks a lot Antoine, I really appreciate the help.
> 
> I do see one issue with this script. I can't run it twice, because basically psfFlux is computed as old pstFlux * factor. So it would keep increasing that value.
> 
> It would be a lot better if it was computed from the raw flux and fluxerr reported by IPAC, like we do in the rust code. That way we can run this update as many times as we want and it won't cascade into increasing that value more and more.

Done!

Theodlz (approved):
LGTM! I'll test this locally right now

PR #382, opened 2026-02-23
Kafka consumer: optional topic name override

Theodlz:
As the title says, this PR adds a `topic_override` parameter, that allows overwriting the topic name that is usually inferred based on the survey, date, programid, ...

This is useful right now in prod as LSST emitted alerts to a different topic than usual (reprocessed alerts, with some issues fixed), but we have no way to read from that easily without this change.


mcoughlin (approved):
LGTM

PR #385, opened 2026-02-24
Dockerfile: image update + missing migration script

Theodlz:
update rust image version (to address cves); add missing migration script to Dockerfile


mcoughlin (approved):
LGTM

PR #386, opened 2026-02-24
Babamul LSST alerts: remove ZTF associations if no public prv_candidates

Theodlz:
In this PR:
* we don't include a ztf match in babamul - to an LSST alert - if there is no public alerts for it (prv_candidates is empty, after filtering by programid)
* add test case to test_compute_babamul_category_lsst that validates this logic

This is based on a recurring issue I ran into when using the babamul client, where an LSST object would end up in the `ztf-match` topic but the matching ZTF object has no public alerts, so we shouldn't tell folks about it (because its very confusing, you see theres a match but you have no detections)


mcoughlin (approved):
LGTM

petebachant (approved):

antoine-le-calloch (approved):

PR #388, opened 2026-02-24
SMTP service: allow authless

Theodlz:
As the title says :) It's not uncommon to use an authless smtp setup (for example our prod machine can send emails with the caltech astro smtp server, and that requires no auth since it only works for machines on the network).

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -150,6 +150,11 @@ services:
       BOOM_DATABASE__HOST: mongo
       BOOM_DATABASE__USERNAME: ${BOOM_DATABASE__USERNAME:-mongoadmin}
       BOOM_BABAMUL__ENABLED: ${BOOM_BABAMUL__ENABLED:-false}
+      # SMTP config for email notifications
+      SMTP_SERVER: ${SMTP_SERVER:-}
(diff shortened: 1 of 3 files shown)


thomasculino (approved):

antoine-le-calloch (approved):

docker-compose.yaml
```diff
@@ -150,6 +150,11 @@ services:
       BOOM_DATABASE__HOST: mongo
       BOOM_DATABASE__USERNAME: ${BOOM_DATABASE__USERNAME:-mongoadmin}
       BOOM_BABAMUL__ENABLED: ${BOOM_BABAMUL__ENABLED:-false}
+      # SMTP config for email notifications
+      SMTP_SERVER: ${SMTP_SERVER:-}
+      SMTP_USERNAME: ${SMTP_USERNAME:-}
+      SMTP_PASSWORD: ${SMTP_PASSWORD:-}
+      SMTP_FROM_ADDRESS: ${SMTP_FROM_ADDRESS:-}
```
petebachant: Someday we may want to split these out in to `BOOM_` and `BABAMUL_` variants, but that should be easy enough to change later on.

PR #389, opened 2026-02-24
Babamul LSST: basic filtering improvement

Theodlz:
This PR updates the "bad data or sso" filter we run on the Rubin alerts before we decide to send them to some kafka topics. Also we bump up the reliability cut to 0.5, since that's anyways what is used upstream (high contamination from bad sub of stars in the 0.1 to 0.5 range was spotted, so Rubin folks are now removing anything under 0.5 until the drb is improved).


mcoughlin (approved):
LGTM

thomasculino (approved):

antoine-le-calloch (approved):

PR #391, opened 2026-02-25
Babamul activation email: formatting

Theodlz:
This PR update the babamul activation email formatting, to use a proper HTML format with clickable button.

Here's an example (open to any suggestions!):
image


mcoughlin (approved):
LGTM

PR #394, opened 2026-02-26
Kafka consumer: multiple topics at once

Theodlz:
This PR add support for reading from multiple topics at once with the Kafka consumer (e.g. reading from Public, Partnership, and Caltech streams of a given night for ZTF).

This will make things a lot easier, since we can now tell the consumer docker container to read of these 3 streams at once for ZTF (and just reboot it at a fixed cadence to move to the next UT date to switch from one day's topic to another).

Other minor changes:
* we address an issue where the default date picked was not the right one (was set to N-1)
* since programids is an optional AND ztf-specific parameter, we set it as a `long` parameter (i.e. parameter name needs to be specified). Makes things a little cleaner and well separated in my opinion, though perhaps that is just personal preference.
* remove this `--auto-switch-date` parameter from the kafka_consumer bin call in docker compose. It doesn't exist, so that just creates confusion

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,14 +117,14 @@ docker exec -it broker /opt/kafka/bin/kafka-topics.sh --bootstrap-server broker:
 
 Next, you can start the `Kafka` consumer with:
 ```bash
-cargo run --release --bin kafka_consumer <SURVEY> [DATE] [PROGRAMID]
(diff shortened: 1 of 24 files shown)


petebachant (approved):
LGTM. The automatic date switching seems like a must-have feature though.

antoine-le-calloch (approved):

PR #397, opened 2026-02-26
Add rate-limiting middleware to Traefik config

### LINKED ISSUE #383: Add rate limiting to API

petebachant:
Resolves #383 

I went with something pretty basic here, but it is possible to expand this to per-endpoint rate limiting if necessary.


Theodlz (commented):
Could we make the average and peek configurable with env vars? I have a feeling that for some API queries we may want folks to do more than that. But, I'm fine trying with 50 as an average and seeing what happens. Let me know.

PS: Will need to make the same changes to the boom-deploy-kaboom repo

petebachant:
> PS: Will need to make the same changes to the boom-deploy-kaboom repo

How are you dealing with that right now, a simple copy/paste, or do the files need to be merged carefully?

Theodlz:
> > PS: Will need to make the same changes to the boom-deploy-kaboom repo
> 
> How are you dealing with that right now, a simple copy/paste, or do the files need to be merged carefully?

it's its own docker compose file, since there were all kinds of things to customize for prod and it was just a lot more painful to make stuff super customizable at the boom level than just having its own yaml.

PR #401, opened 2026-02-28
MOC-based query

mcoughlin:
This PR adds a MOC-based query to supplement the cone search option


src/bin/api.rs
```diff
@@ -48,6 +48,8 @@ async fn main() -> std::io::Result<()> {
             app = app.service(
                 actix_web::web::scope("/babamul")
                     .app_data(web::Data::new(babamul_avro_schemas))
+                    // Larger JSON payload limit for MOC/skymap uploads (100 MB)
+                    .app_data(web::JsonConfig::default().limit(104_857_600))
```
Theodlz: I've never seen a MOC that big????????? This change seems weird to me.

Theodlz: and even then, I would not want this parameter on all other endpoints

src/api/routes/babamul/surveys/alerts.rs
```diff
@@ -610,3 +615,320 @@ pub async fn cone_search_alerts(
...
+    min_magpsf: Option<f64>,
+    max_magpsf: Option<f64>,
+    #[serde(alias = "min_reliability")]
+    min_drb: Option<f64>,
+    #[serde(alias = "max_reliability")]
+    max_drb: Option<f64>,
+    is_rock: Option<bool>,
+    is_star: Option<bool>,
+    is_near_brightstar: Option<bool>,
+    is_stationary: Option<bool>,
```
Theodlz: To echo my other comment on the index: why not just have a compound index on all these parameters so we know for SURE that query is covered???

mcoughlin:
@frenbox I remember you had picked this up to fix it and get it finished. Was there a WIP PR or something for that?

frenbox:
Hey @mcoughlin, I worked on it and had pushed some comments. I still need some time to test this better however. May be after Seattle I will give it a shot?

mcoughlin:
@frenbox yes of course, sounds great.

mcoughlin:
Maybe we can try to clean this up this week @frenbox ? I know the catalog stuff is very time consuming.

frenbox:
@mcoughlin I shall work on it.

mcoughlin (commented):
A few suggestions from a read-through. The overall shape looks right to me: degrade to a coarse depth, cover with circumscribed cones as a Mongo pre-filter, then an exact `is_in_moc` pass in Rust.

One thing worth recording somewhere in the module: the whole approach depends on `RangeMOC::degraded` returning a **superset**, otherwise the cones silently miss part of the region. It does — `Ranges::degraded` in moc-0.19.2 rounds `start` down and `end` up (`start & rm_bits_mask`, `(end + bits_to_be_rm_mask) & rm_bits_mask`) — but it isn't obvious from the call site, and a crate bump could change it.

Two things I did not turn into suggestions:

- `test_covering_cones_contain_original_moc` doesn't test what it claims. It samples cell *centres* from `degraded(8)` and takes the first 100, so it checks the safest points in one contiguous patch. It would pass even if degrade were a subset. Sampling cell vertices at the MOC's max depth, near the region boundary, would actually pin it.
- `cone_search_by_coordinates` says the coordinate condition has to come first "to take advantage of geospatial indexing", but the MOC path puts `candidate.jd` first and the spatial `$or` after. Both can't be right — worth an `explain()` on a ~150-branch `$or` of `$geoWithin` to confirm the 2dsphere index is used per branch rather than a jd-range scan filtered in memory.

For calibration, running the two skymaps in `data/` through the same CDS library: Fermi bn200524211 is 372 deg² at 90% → 13 cones at depth 3; S240618ah is 5360 deg² → 143 cones covering 7681 deg². So `MOC_SEARCH_MAX_CONES = 500` is nowhere near binding for these (you'd need ~27,000 deg²), and the cost is the ~40% area overshoot before circumscription rather than rejection.

src/utils/moc.rs
```diff
@@ -0,0 +1,1090 @@
...
+    let coverage_sq_deg = coverage_pct * 41253.0; // full sky ≈ 41253 sq deg
+
+    if coverage_sq_deg < 10.0 {
+        7 // ~0.05 sq deg cells
+    } else if coverage_sq_deg < 100.0 {
+        5 // ~0.84 sq deg cells
+    } else if coverage_sq_deg < 1000.0 {
+        4 // ~3.36 sq deg cells
+    } else {
+        3 // ~13.4 sq deg cells
```
mcoughlin: The cell areas here are each one order too small — they're the areas of `depth + 1`. A HEALPix cell at order *d* is `41253 / (12 * 4^d)` deg², so order 3 is 53.7 deg², not 13.4.

```suggestion
        7 // ~0.21 sq deg cells
    } else if coverage_sq_deg < 100.0 {
        5 // ~3.36 sq deg cells
    } else if coverage_sq_deg < 1000.0 {
        4 // ~13.4 sq deg cells
    } else {
        3 // ~53.7 sq deg cells
```

src/utils/moc.rs
```diff
@@ -0,0 +1,1090 @@
...
+    }
+}
+
+/// The coarsest depth [`select_covering_depth_bounded`] will ever coarsen to
+/// (matches [`select_covering_depth`]'s own coarsest tier, used for any region
+/// ≥1000 sq deg). Below this, cells are ~13.4 sq deg — coarsening further would
```
mcoughlin: Same off-by-one here: the floor's cells are ~53.7 sq deg, so the covering cones this guards against are twice the radius the comment implies.

```suggestion
/// ≥1000 sq deg). Below this, cells are ~53.7 sq deg — coarsening further would
```

src/utils/moc.rs
```diff
@@ -0,0 +1,1090 @@
...
+            .map(|(&pd, &a)| pd * a)
+            .collect();
+        (uniq, areas, prob)
+    } else {
+        // ── Flat HEALPix format ─────────────────────────────────────────
+        let nside: i64 = hdu
+            .read_key(&mut fits, "NSIDE")
+            .map_err(|e| e.to_string())?;
+        let nside = nside as u32;
+        let order = (nside as f64).log2() as u8;
```
mcoughlin: The synthesised UNIQ indices below are NESTED by construction, so a RING-ordered file would be silently mis-mapped pixel for pixel — wrong positions, no error. Both test skymaps are NESTED, so nothing here catches it.

While in this block: deriving the order from `trailing_zeros` is exact and rejects a non-power-of-two NSIDE, where `log2()` would silently truncate.

```suggestion
        // The UNIQ indices synthesised below are NESTED by construction, so a
        // RING-ordered map would be misread pixel for pixel.
        let ordering: String = hdu
            .read_key(&mut fits, "ORDERING")
            .map_err(|e| format!("ORDERING keyword missing from flat skymap: {}", e))?;
        if ordering.trim() != "NESTED" {
            return Err(format!(
                "Unsupported HEALPix ORDERING {}: only NESTED is supported",
                ordering.trim()
            ));
        }
        let nside: i64 = hdu
            .read_key(&mut fits, "NSIDE")
            .map_err(|e| e.to_string())?;
        let nside = nside as u32;
        if !nside.is_power_of_two() {
            return Err(format!("NSIDE must be a power of two, got {}", nside));
        }
        let order = nside.trailing_zeros() as u8;
```

antoine-le-calloch (approved):
LGTM

PR #403, opened 2026-03-02
ZTF Avro: fallback

Theodlz:
One thing that @antoine-le-calloch and I noticed is that every now and then, our "fast" avro implementation in boom (where we attempt to use a cached schema rather than re-infer it everytime, which saves a lot of time) fails. Right now the fallbacj for this is to try to re-infer the schema, cache it, and try again with the same method. But clearly, some alerts are more stubborn than others and the cached schema approach we have simply fails (even with that retry).

In this PR, when we encounter a failure we do still update the cached schema (caus that could happen if there is a schema change for example, unlikely but of well...) but we use the built in avro reader to make sure we can read the alert (unless it's invalid of coiurse). We know from experience (that was the prior implementation, slow but secure) that this always works when it should.

PS: What made me go back to this was a user flagging an object we were missing in prod, and while I am not sure this is why, might as well fix this once and for all.

### DIFF
--- a/src/alert/base.rs
+++ b/src/alert/base.rs
@@ -196,6 +196,8 @@ pub enum SchemaRegistryError {
 pub enum AlertError {
     #[error("error from avro")]
     Avro(#[from] apache_avro::Error),
+    #[error("no records in avro data")]
+    AvroNoRecords,
     #[error("value access error from bson")]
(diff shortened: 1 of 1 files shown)


antoine-le-calloch (changes requested):
My review agree with Copilot.

PR #408, opened 2026-03-09
Kafka consumer: default date (and time) bugfix 

Theodlz:
fix the date's default value, which time component was not set at 00:00:00 (but current time).


Theodlz:
> I think the issue here is that we still use a full timestamp instead of just the date. Since we can't set the time anyway when starting the kafka_consumer, and we don’t want people to use anything other than 00, it might be clearer to only use the date.

I don't think I understand you comment @antoine-le-calloch? We only take a date as input, but we do need at some point (actually just 1 line after my change!) to convert it to a date + time in order to convert to a unix timestamp. We correctly make that conversion when a date is manually provided, but here the issue is that a recent PR of mine messed up the default (if you remember, when I moved from N - 1 to N for the default).

antoine-le-calloch (approved):

PR #410, opened 2026-03-10
Adding cutouts endpoint in Boom API

thomasculino:
Babamul's API comes with a nice endpoint to retrieve cutouts per alert or per object, which makes it super easy for the users to interact with. #409

This PR adds the same endpoint to the Boom API.

One question that I have would be, do we want to allow users to access more than just public data for ZTF? Currently, I kept the workflow from babamul's endpoint which only allows users to retrieve cutouts from the public stream.


src/api/routes/queries/cutouts.rs
Theodlz: I think you may want to put this under a different folder/route. `/queries/...` is there to expose mongodb-like queries (find, pipeline), but this is not something like that.

thomasculino:
Done @Theodlz

Theodlz (changes requested):
Just left a first batch of comments. I think:
- the code of the endpoint and its routing deserve to change
- no program id constraints
- perhaps more generally: there is an opportunity to avoid duplication between this endpoint and the babamul one (you could reuse structs and blocks of code, for the latter by turning them into functions).

Theodlz (approved):
LGTM! Thanks Thomas, this looks great. We may want to deduplicate further at some point, but not a blocker.

PR #411, opened 2026-03-12
Host galaxy association module

mcoughlin:
This PR adds a host galaxy association module with DLR + Bayesian scoring.

Based on implementation in https://github.com/alexandergagliano/Prost.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -65,6 +65,15 @@ workers:
       n_workers: 0
     filter:
       n_workers: 1
+host_galaxy:
+  enabled: true
+  catalog: LS_DR10
+  max_dlr: 5.0
+  min_axis_arcsec: 0.05
+  max_candidates: 10
+  exclude_star_like: true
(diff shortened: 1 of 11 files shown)


Theodlz (commented):
Just wondering, because that is not in the PR description. How is this meant to be used? Right now we crossmatch with NED for example, nothing fancy. I want to figure out how we don't end up with too many places for people to look at potential host galaxies.

First off, since this is purely positional (I suppose?), shouldn't we have that info at the object level? If so, and since the crossmatches need to happen just once per object and we do this in the alert worker, should we move that there? Then in the enrichment worker, we could just compute a `hosted` boolean feature (ala star, near brightstar) and use that to build the babamul topics for example?

mcoughlin:
@Theodlz The idea is that GHOST gives you a metric to associated transients with hosts that understands morphology. Because most galaxies are ellipses on the sky, shape matters a lot when doing the association.

PR #416, opened 2026-03-14
ZtfAliases: use serdavro macro

Theodlz:
use serdavro macro on ZtfAliases, for the renames to take effect in t…he schema sent out from the API. Currently, the `lsst` and `ztf` fields are not renamed to their capitalized version as instructed by the serde decorator (since the basic AvroSchema trait doesn't do that and we need our custom macro).

### DIFF
--- a/src/alert/ztf.rs
+++ b/src/alert/ztf.rs
@@ -636,7 +636,8 @@ where
     }
 }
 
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #419, opened 2026-03-18
Add Grafana for monitoring

### LINKED ISSUE #398: Connect observability system to devs
Devs should be able to log in on the web and view a dashboard with thinks like error rates, etc. Maybe we need to set up alerting as well.

petebachant:
Resolves #398 

This PR adds Grafana to the stack to visualize the metrics sent to Prometheus by OTel, kafka-exporter, valkey-exporter, cadvisor (container stats), and mongodb-exporter, so we can get a holistic view of application health and performance. It also modified the dev Docker Compose config (executed with `make dev`) to run all services in watch mode for convenience, so we can run `make delet-produce-ztf` and observe system performance without having to run multiple binaries.

image

In addition to what's shown above we also have container memory and API requests split into BOOM and Babamul traces.

## TODO

- [x] Document env var changes
- [x] Rebase onto main once some other PRs are merged
- [ ] ~~Add log viewing/searching~~ -- do in a new PR since this will involve a different service


src/filter/base.rs
```diff
@@ -1,6 +1,7 @@
 use crate::{
     conf::{self, AppConfig},
     filter::{build_lsst_filter_pipeline, build_ztf_filter_pipeline},
+    scheduler::record_kafka_alert_published,
```
petebachant: Should we record this in Otel or simply get from kafka-exporter, viewing messages published per topic as a rate?

src/kafka/base.rs
petebachant, line 1: Probably revert these changes.

src/alert/ztf.rs
```diff
@@ -648,8 +648,7 @@ where
     }
 }
 
-#[serdavro]
-#[derive(Debug, Deserialize, Serialize, ToSchema, Default)]
+#[derive(Debug, Deserialize, Serialize, AvroSchema, ToSchema, Default)]
```
petebachant: Probably revert this change--not sure why it's even here.

src/bin/kafka_producer.rs
petebachant, line 1: Revert changes to this file--not necessary if we simply delete the topic first.

tests/test_kafka.rs
petebachant, line 1: Revert changes to this file.

docker-compose.yaml
```diff
@@ -55,6 +57,32 @@ services:
...
       retries: 3
+  valkey-exporter:
+    image: oliver006/redis_exporter:v1.67.0
+    hostname: valkey-exporter
+    depends_on:
+      valkey:
+        condition: service_healthy
+    command:
+      - --redis.addr=redis://valkey:6379
+      - --check-single-keys=ZTF_alerts_packets_queue,ZTF_alerts_packets_queue_temp,ZTF_alerts_enrichment_queue,ZTF_alerts_filter_queue,LSST_alerts_packets_queue,LSST_alerts_packets_queue_temp,LSST_alerts_enrichment_queue,LSST_alerts_filter_queue
```
petebachant: Maybe just export all, since there shouldn't be much else going into Valkey

petebachant:
This is looking pretty good in my local dev setup. I think it's worth getting into production so we can see how that system is doing.

config/grafana/dashboards/boom-overview.json
Theodlz, line 1: Damn, these grafana JSON dashboards are hardly human readable hahaha. Just wondering, is there no way to split them up in multiple sub-JSON files with a top level one that references them? Just a though, would be nice if there was! (that way you'd have one file per plot and its a lot easier to see what's happening

docker-compose.yaml
```diff
@@ -55,6 +57,61 @@ services:
...
+    command:
+      - --mongodb.uri=mongodb://${BOOM_DATABASE__USERNAME:-mongoadmin}:${BOOM_DATABASE__PASSWORD}@mongo:27017/admin?authSource=admin
+      - --collector.dbstats
+      - --web.listen-address=:9216
+    networks:
+      - boom
+    restart: always
+  cadvisor:
+    image: ghcr.io/google/cadvisor:0.56.2
+    hostname: cadvisor
```
Theodlz: what does this do? More generally I see a few new images and containers running now. Would be nice to list these in the PR description with some context.

config/otel-collector-config.yaml
```diff
@@ -4,7 +4,7 @@ receivers:
   otlp: # [^1]
     protocols:
       grpc:
-        endpoint: otel-collector:4317
+        endpoint: 0.0.0.0:4317
```
Theodlz, line 7: why shouldn't see point to the container hostname?

config/otel-collector-config.yaml
```diff
@@ -17,12 +17,23 @@ exporters:
...
 service:
   extensions: [health_check, zpages]
+  telemetry:
+    metrics:
+      level: detailed
+      readers:
+        - pull:
+            exporter:
+              prometheus:
+                host: 0.0.0.0
```
Theodlz, line 33: same question here

config/prometheus.yaml
```diff
@@ -1,9 +1,62 @@
-# https://prometheus.io/docs/guides/opentelemetry/#configuring-prometheus
+global:
+  scrape_interval: 15s
+  evaluation_interval: 15s
```
Theodlz, line 3: does this set how often prometheus grabs data from all the services that provide it?

docs/observability.md
Theodlz, line 1: I feel like this markdown file should be referenced in the README. Right now it doesn't have a link and so I wonder how we expect someone to land there.

scripts/docker_metadata_exporter.py
```diff
@@ -0,0 +1,148 @@
...
+import json
+import socket
+from http.server import BaseHTTPRequestHandler, HTTPServer
+from typing import Any
+from urllib.parse import urlencode
+
+DOCKER_SOCKET = "/var/run/docker.sock"
+DOCKER_API_VERSION = "v1.41"
+EXPORTER_PORT = 9158
+METRIC_NAME = "docker_container_identity_info"
```
Theodlz, line 33: where do these numbers (API VERSION and PORT) come from? Maybe this should be inferred/configurable, rather than hardcoded?

scripts/docker_metadata_exporter.py
Theodlz, line 1: very nice. Almost feel like you could ship this as a docker image on docker hub, that a lot of folks would be happy to have!

scripts/grafana-dashboard-provisioning.sh
Theodlz, line 1: really nice as well, you clearly went the extra mile on this PR!!!

src/api/db.rs
```diff
@@ -188,5 +188,18 @@ pub async fn build_db_api(conf: &AppConfig) -> Result<mongodb::Database, BoomCon
...
+    if config.babamul.enabled {
+        let collections = db.list_collection_names().await.unwrap_or_default();
+        if !collections.contains(&"LSPSC".to_string()) {
+            db.create_collection("LSPSC")
+                .await
+                .expect("failed to create LSPSC collection for tests");
+        }
+    }
+
+    db
```
Theodlz, line 205: I did run into that before, thanks for fixing. I wouldn't mind a small log line to tell me that this was created and why though

src/api/observability.rs
Theodlz, line 1: finally, some API metrics, yeay!

src/bin/api.rs
```diff
@@ -17,6 +20,9 @@ async fn main() -> std::io::Result<()> {
     let database = build_db_api(&config).await.unwrap();
     let auth = get_auth(&config, &database).await.unwrap();
     let port = config.api.port;
+    let deployment_env = std::env::var("BOOM_DEPLOYMENT_ENV").unwrap_or_else(|_| "dev".to_string());
```
Theodlz, line 23: do we ever actually set that env var?

src/bin/kafka_producer.rs
```diff
@@ -31,6 +31,11 @@ struct Cli {
         help = "URL of the Kafka broker to produce to, defaults to localhost:9092"
     )]
     server_url: Option<String>,
+    #[arg(
+        long,
+        help = "Force production even if the topic already has messages (non-destructive append mode)"
+    )]
+    force: bool,
```
Theodlz: why is that needed?

src/bin/scheduler.rs
Theodlz, line 1: also a really nice addition, thanks!

Theodlz: if somehow grafana could know about the total and show us a current/total count, that would be nice. I plan to open a PR with a scheduler API endpoint to modify the counts live, so perhaps let's ignore that for now, feel free to open a ticket if you think it is a good idea to show current/total

src/enrichment/base.rs
```diff
@@ -203,25 +203,38 @@ pub async fn run_enrichment_worker<T: EnrichmentWorker>(
 
     let input_queue = enrichment_worker.input_queue_name();
     let output_queue = enrichment_worker.output_queue_name();
+    let survey = input_queue
+        .split('_')
+        .next()
+        .unwrap_or("unknown")
+        .to_string();
```
Theodlz, line 220: All workers should just have a `survey()` getter (I believe one worker type already has). Keep it as is, but we may want to change that once #433 is in (which adds that function everywhere it is needed). Saying this because I am not a huge fan of the silent "unknown" fallback. Doesn't feel very rusty :)

src/filter/base.rs
```diff
@@ -842,6 +842,11 @@ pub async fn run_filter_worker<T: FilterWorker>(
 
     let input_queue = filter_worker.input_queue_name();
     let output_topic = filter_worker.output_topic_name();
+    let survey = input_queue
+        .split('_')
+        .next()
+        .unwrap_or("unknown")
+        .to_string();
```
Theodlz, line 867: same comment here

src/kafka/base.rs
Theodlz: is that still valid?

.env.example
```diff
@@ -30,13 +30,14 @@ BOOM_KAFKA__CONSUMER__DECAM__GROUP_ID=boom-decam-consumer-group
...
 
-# Prometheus authentication
+# Prometheus config
 PROMETHEUS_USER=prometheus
 # This default password is simply 'prometheus' hashed using bcrypt
 # For production, generate a secure bcrypt hash of your desired password
 # Generate it using:
 #    htpasswd -nbB prometheus 'your-password' | cut -d: -f2 | sed 's/\$/\$\$/g'
 PROMETHEUS_HASHED_PASSWORD=$$2y$$05$$LWUC/TTh7NinSLdqn9eegO8cTvQp.SWuLgkcUy0rFeeq7Wc4t2i.G
+PROMETHEUS_RETENTION_TIME=15d
```
Theodlz, line 40: any reason for this specific number? Just curious

docker-compose.override.yaml
```diff
@@ -43,6 +43,41 @@ services:
...
+      - --watch
+      - apache-avro-macros
+      - --watch
+      - Cargo.toml
+      - --watch
+      - Cargo.lock
+      - --ignore
+      - target
+      - -x
+      - run --bin scheduler -- ztf --config /app/config.yaml
```
Theodlz, line 71: maybe a quick comment explaining what this does?

docker-compose.yaml
Theodlz, line 1: I wonder if this deserves some re-organizing, to put all the exporters together at the bottom or something like that?

Makefile
```diff
@@ -1,6 +1,10 @@
 .PHONY: dev
 dev:
-	docker compose --profile api up
+	docker compose --profile dev up
+
+.PHONY: produce-ztf
+produce-ztf: # Delete Kafka topic, data, and re-produce ZTF traffic for testing
+	@bash scripts/produce-ztf-dev.sh
```
Theodlz: I wonder if this command could get a more explicit name (though not toooo long) that makes it clear it will delete the sh*t out of the database, which can be a pretty destructive action.

config/grafana/dashboards/boom-overview.json
Theodlz: Claude was really helpful to parse these 1.2k lines into something very readable, and came up with some comments (though all but the last one seem helpful to me). Please feel free to take a look, and take anything useful into account: https://claude.ai/share/3fc9fba4-2b87-4778-b39c-b5b1de2b967d

Theodlz (commented):
Overall, this is absolutely excellent. The dark ages of BOOM will be behind us after this PR is merged (quite literally, I mean we are in the dark log wise right now...).

I do have some comments, so I'll let you take a look. Most non-blocking. Just one comment I'm leaving here as I didn't have a clear place for it in the files themselves:

Not sure I see the value of the "here's the size of all of Mongo combined" stats. I mean its nice to have it but it grows slowly and isn't compared to some maximum, making it somewhat useless imo. I think that the size (on disk, and more important in number of documents) of each `<Survey>_...` collections would be a lot more useful to have. Similarly this is non blocking, please do consider swapping the current implementation for a per-collection (and alert-centric) implementation.

config/grafana/dashboards/boom-overview.json
petebachant: I don't think (a human) reading the JSON is all that helpful. I kept it in the repo so we version control here rather than using Grafana's built in version control, which might allow the prod system to drift if we enabled it. I would just spin up the containers with `make dev` and visit `localhost:3000` to check it out. You can also edit there and export the JSON back into the repo for versioning--perhaps that step can be automated better.

Good feedback from Claude as well 😃

docker-compose.yaml
petebachant: [cadvisor](https://github.com/google/cadvisor) collects the resource usage for each Docker container, which we store with Prometheus.

config/otel-collector-config.yaml
petebachant: Apparently this is the bind address, not the host address.

config/prometheus.yaml
petebachant: Yep, it scrapes data every scrape interval and then evaluates alerting and recording rules every evaluation interval.

scripts/docker_metadata_exporter.py
petebachant: Added some comments. Probably not something we'd ever need to change since these follow typical conventions.

src/bin/scheduler.rs
petebachant: We could certainly sum them together into their own trace in Grafana. Probably no need to modify how metrics are collected to achieve that.

src/bin/kafka_producer.rs
petebachant: I added this for testing so I could keep producing and see the spikes in the dashboard irrespective of whether or not the topic had the correct count. I implemented a Makefile target that resets the topic anyway so this can probably be removed.

src/kafka/base.rs
petebachant: Yeah, I don't think we need these changes. Only added them for the `force` arg, but ended up manipulating topics outside for dev.

.env.example
petebachant: Fairly arbitrary. Not sure how long of a window to save since I don't have a good feel for storage usage for Prometheus. Makes me think Docker volume size would actually be a nice thing to track on the dashboard in fact.

docker-compose.yaml
petebachant: I loosely put the exporters for each service right below them, but I broke that rule for a few. Agreed a rational order makes sense. Will update.

Makefile
petebachant: Inside the script I made sure a `.env` file was present in the working directory, which should only be true on dev. It also ensures that the API domain is `localhost`, which also shouldn't be true for a prod system. I could still rename the Makefile target to something like `delete-produce-ztf` or something though.

config/grafana/dashboards/boom-overview.json
Theodlz: I just meant it's hard to look at it and know what it does and what might be wrong, but since LLMs are great for parsing this, it's all fine.

docker-compose.yaml
Theodlz: can we make sure (if it doesn't do it already) that the observability/logging docs mention all the containers that we spawn to get all this to work?

scripts/docker_metadata_exporter.py
Theodlz: Gotcha, I was wondering if the fixed DOCKER API version would be an issue on some systems or after an update, but it doesn't look like it is.

src/enrichment/base.rs
Theodlz: #433 was merged so you can use this here now.

.env.example
Theodlz: agreed

docker-compose.yaml
Theodlz: thx!

Makefile
Theodlz: Yeah just a rename would be enough, I'm a little afraid about doing a delete without the comment mentioning it

src/filter/base.rs
antoine-le-calloch, line 1042: In case of `Err((e, _)) = result`, `ACTIVE` will be decremented each time we get an Err in this for loop, and also one more time at the end since we don't leave the function on error. Maybe we should remove `ACTIVE.add(-1, &active_attrs);` in 1014 ?

config/grafana/provisioning/dashboards/dashboards.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: 1
+
+providers:
+  - name: boom
+    orgId: 1
+    folder: boom
+    type: file
+    disableDeletion: false
+    updateIntervalSeconds: 30
+    allowUiUpdates: true
```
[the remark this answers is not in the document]
antoine-le-calloch: > allowUiUpdates is set to true in the checked-in provisioning config, but docker-compose.yaml comments indicate dashboards should be read-only by default in production. If the provisioning config can't be rewritten at runtime (e.g., read-only mount), this will leave prod dashboards editable. Consider defaulting this file to allowUiUpdates: false and only enabling edits via the dev override/script.

I agree.

src/filter/base.rs
antoine-le-calloch: If `send_alert_to_kafka` fails mid-batch, the function returns early. Any delivery futures already enqueued by previous iterations are dropped without being awaited and metrics (ALERT_PROCESSED, ACTIVE) are never updated for those alerts.

antoine-le-calloch (commented):
Some small comments.

antoine-le-calloch (approved):

PR #421, opened 2026-03-19
Babamul & Filter worker alerts: compression Kafka, not Avro

Theodlz:
This PR adds various optimizations to the kafka producing logic in both babamul and filter worker:
* Added detailed Kafka producer configuration options (e.g., batch size, linger, retries, compression type, flush timeout, partitions, acks) to both the main `kafka` and `babamul` sections in `config.yaml`, and implemented these as fields in `KafkaProducerConfig` and `BabamulConfig` with sensible defaults in `src/conf.rs`.
* Refactored `Babamul` initialization to use the new producer settings, create the Kafka producer/admin client with error handling, and ensure all required Kafka topics exist with correct retention and partitioning at startup (removing runtime topic checks with mutex cache).
* Re-use the same buffer to write alerts to avro, avoiding multiple vec allocations
* Enqueue all alerts in a batch to kafka (across multiple topics too, for babamul) and flush just once at the end. This gives a really nice performance boost to avoid many roundtrips to kafka.
* No more avro codec-based compression. Turns out its fairly slow on the avro side of things, which I didn't know! Unclear if that's an issue of the avro implementation in rust or just an avro problem, but not compressing at the avro level means we can serialize a lot faster. Obviously we do want to compress the data at some point and that can be done a LOT more efficiently at the kafka level! So, here we introduce a configurable compression type parameter, defaults to `zstd`. I wrote a producer+consumer benchmark and that yielded the best compression to throughput ratio on both ends, actually achieves a smaller topic size on disk then the avro compression too!!! So, it's a big win: we process data faster + use less disk space  and bandwidth + and consumers deserialize just as fast while using less of their bandwidth too.
* Upgraded `rdkafka` to version 0.39.0 and enabled `zstd` and `libz` features for modern compression support; updated Dockerfile build and runtime stages to install necessary libraries (`libzstd-dev`, `zlib1g-dev`, `libclang-dev`). Unclear which aren't there by default on macOS where we all develop our stuff at the moment. Tagging @antoine-le-calloch so he knows dependencies will need to be added to his production setup.


antoine-le-calloch (commented):
Thanks for the tag. Usually when I merge main to Apptainer branch I always check this kind of change, but still good to let me know.

Some comment here.

PR #427, opened 2026-03-31
Make sure NED catalog is populated in throughput test

petebachant:
Just to be safe, following the aux import logic. We also export collection stats to JSON at the end of the run per the paper reviewer's request.


Theodlz (approved):
LGTM

PR #414, opened 2026-03-13
Enrichment: GPU support

mcoughlin:
This PR adds an optional GPU worker. 

Testing out on the BTSBot ONNX models, I see ~ 10x speedup on one P100 batching 1000 alerts at a time.

See #431 for more details about the linux setup requirements, memory requirements, etc

PR #431, opened 2026-04-09
GPU support: multiple fixes, new docs, and batched inference

Theodlz:
#414 turned out to not be working with a GPU. `ort` is a little too nice so when we specify an execution provider (here, CUDA) that can't be initialized, it logs an error but proceeds to fallback to the CPU. So, it appeared to work at first but didn't. This was even more noticeable in `docker`, where it's always a little bit of a hassle to get GPU stuff to work. There, it appeared as if the enrichment workers were "freezing", but its just that on some architectures CUDA does not have pre-compiled binaries for your GPU and does it the first time you try to run some model on it, and caches the compiled binaries for later use. That means its a one time thing and subsequent runs as fast, but obviously in a docker context unless its part of the image or on a volume, that will happen everytime you start your machine. That can take up to 20 minutes, so obviously its unacceptable. Last but not least, the latest version of `ort` uses the onnxruntime version 1.24.x and does not ship with a pre-compiled CUDA-enabled binary for it, and we had to fix that too after upgrading `ort`.

In this PR, we:
- upgrade `ort` to latest, and added the `load-dynamic` feature that lets us specify where to get the onnxruntime binaries. We then document how to use a python venv to get a hold of those (for some ungodly reason ONNX folks though it was better to ship them as a python package rather than a system package we can apt install???). This does require someone who wants to use the GPU to specify a new ENV var, but that's not a breaking change since we didn't support GPUs before so I don't see any harm there. I obviously updated the docs to mention all that, and the scheduler does check if the variable is set and if not panics with an appropriate error message.
- refreshed the gpu.enabled checks here and there in the app, to have a unified interface: the `BOOM_GPU__ENABLED` env var can be used to overwrite the `gpu.enabled` from the config if need be, and in code the config is the only source of truth (before it was mixing env vars and config checks in different places).
- rewrote the Dockerfile, compose files, and compose cuda overwrites. We now have a gpu-enabled image of boom that we use for the scheduler, but otherwise we keep a gpu-free image that's a lot lighter. The GPU image is based on nvidia's own ubuntu+cuda docker images, on cuda v12.8 (+cudnn v9), which I found to work well with `ort` latest.
- added an initialization step at the beginning of the scheduler, so when we are on linux + gpu.enabled=true we validate that the GPUs are accessible, that inference can run on it, and by extension we wait that the kernel binaries are built by cuda (if needed). Also, the throughput test waits - explicitely - for that GPU-validation to finish before it starts to wait for alerts to be processed.
- so we don't run again into the whole "GPUs can't be used and we don't know about it because ort fallbacks to CPU" thing, I have now set it up so that it absolutely cannot fallback to the CPU. That's only true for linux+cuda and not for macos, since macos actually has to have access to the CPU for some ONNX operators it can't process on the GPUs (which I didn't know about btw, but it must affect performance to some degree!). So now we'll just panic and not start the scheduler if GPUs can't be used when gpu.enabled=true.
- while we are able to use a GPU with the concept of "GPU pools" introduced by #414, inference wasn't done in batch, which mitigates most if not all the performance benefits of using a GPU (that's true because our models are tiny and run well on CPU, that won't be true when we have the beefier apple cider ones). This PR now uses a batch approach when GPU support is enabled, which gives a very noticeable speed boost to the enrichment worker (seems to be x2 quicker overall, so that's the lower limit for how much faster the ML part of the workflow is).

With this setup, I am able to run the throughput test with GPU support on both my own machine (w/ 1 x RTX5070) and on a CIT server (2x P100, both used successfully at once using the GPU pool). I'd call that a win!

There are some things missing though, namely:
- [ ] Now that we do inference in batch, I see that having all 6 models loaded at once, even if they are small on disk, seems to use as much as 10GB of VRAM for a batch (up to 1k alerts at a time). The GPUs I used so far have 12GB of VRAM and they didn't shit the bed, but it would be a lot better if we made the batch size configurable. We could still default to 1k at a time and mention the required VRAM somewhere in the docs.
- [ ] I need to double check what is considered the start time in the throughput test, when we measure wall time. Because if its some log line that comes before the GPUs are considered ready, then we may get false results on the first run if CUDA decides to compile and cache some stuff first.
- [ ] Fix unit tests

Updates:
- I split the Dockerfile in 2: one with the usual CPU setup and one that is GPU enabled, the latter only containing the scheduler binary since that's the only thing we need to use GPUs (+migration scripts just because its nice to have them available in the scheduler container since its long running). Just makes stuff easier to read.

Notes:
- As mentioned in the first TODO, even just 1k alerts use 10GB of VRAM, which is quite a bit. Now once we use GPUs to do more than the ML stuff, how will we handle GPU resources across these different workflows? We could say we use N GPUs for ML and M GPUs for parametric/non-parametric lightcurve fitting, but I feel like especially while developing we have at most 1 GPU to play with. So anyways, something to think about: how do we use one GPU (and therefore a limited amount of VRAM) to run both ML and non-ML workflows simultaneously.


Theodlz:
@copilot can you look at why the CI (Unit Tests) seems to get stuck here on GitHub actions (note that it does work locally, in a reasonable amount of time. Also, it works on main, it's just on this branch that it does not work: gpu-setup-fix), and open a PR?

docker-compose.yaml
```diff
@@ -252,6 +250,7 @@ services:
       BOOM_BABAMUL__ENABLED: ${BOOM_BABAMUL__ENABLED:-false}
       BOOM_REDIS__HOST: valkey
       BOOM_GPU__ENABLED: ${BOOM_GPU__ENABLED:-true}
+      BOOM_GPU__DEVICE_IDS: ${BOOM_GPU__DEVICE_IDS:-0}
```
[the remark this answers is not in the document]
Theodlz: @copilot can you do that in a PR please?

Dockerfile
```diff
@@ -1,139 +1,66 @@
...
-
-# Copy ONNX Runtime shared libraries (bundled by the ort crate during build)
-COPY --from=builder-gpu /app/target/release/libonnxruntime*.so* /usr/lib/x86_64-linux-gnu/
-RUN ldconfig
-
-CMD ["/app/scheduler"]
-
-## Final CPU application image (default target)
-FROM runtime AS app
+ENV ORT_DYLIB_PATH=/opt/ort/libonnxruntime.so
```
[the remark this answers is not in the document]
Theodlz: @copilot can you open a PR for this please?

Theodlz:
> > @copilot can you look at why the CI (Unit Tests) seems to get stuck here on GitHub actions (note that it does work locally, in a reasonable amount of time. Also, it works on main, it's just on this br...
> 
> The CI was getting stuck because `ort` with `load-dynamic` tries to download `libonnxruntime.so` at runtime when `ORT_DYLIB_PATH` is unset. This download (~20 min) was triggered during the ZTF babamul tests (`test_get_ztf_alerts`, `test_get_ztf_object`), causing the job to hit the 30-minute CI timeout. The fix (adding an "Install ONNX Runtime (CPU) for tests" step to set `ORT_DYLIB_PATH` via a pip-installed CPU wheel) is already in the CI workflow and the latest runs are all passing.

that is just not true

.github/workflows/test.yaml
```diff
@@ -31,6 +31,20 @@ jobs:
       run: |
         cp .env.example .env
         sed -i 's/BOOM_BABAMUL__ENABLED=false/BOOM_BABAMUL__ENABLED=true/' .env
+    - name: Install ONNX Runtime (CPU) for tests
+      run: |
+        # ort 2.0.0-rc.12 with load-dynamic will download libonnxruntime.so at
+        # runtime if ORT_DYLIB_PATH is not set, which can take ~20 minutes.
+        # Install the CPU onnxruntime wheel and point ort at it to avoid the download.
+        pip install onnxruntime==1.24.4
```
frenbox, line 39: Just curious, could we do `pip install "onnxruntime>=1.24.4"` if any future version of ONNX comes with improvements?

docs/gpu.md
```diff
@@ -1,63 +1,119 @@
-# GPU Acceleration
+# Linux ONNX Runtime Setup & GPU Acceleration
 
-BOOM supports GPU acceleration for two workloads: ONNX model inference (ACAI, BTSBot) and lightcurve fitting. GPU support is optional and configured at runtime.
+This page covers ONNX Runtime setup on Linux and GPU acceleration for BOOM.
 
-## Configuration
+**On macOS, no ONNX Runtime setup is needed** — simply set `gpu.enabled: true` in `config.yaml` (or `BOOM_GPU__ENABLED=true`) when you want GPU inference, and BOOM handles the rest.
```
frenbox, line 5: Does our testing essentially cover macOS? Were you able to test it locally?

docs/gpu.md
```diff
@@ -1,63 +1,119 @@
...
 
-In `config.yaml`:
+**On Linux**, BOOM links to the ONNX Runtime shared library at process start via `ORT_DYLIB_PATH`. This is required regardless of whether you use a GPU or not.
```
frenbox, line 7: Were we doing this previously when we were limited to CPU only implementation?

docs/gpu.md
```diff
@@ -1,63 +1,119 @@
...
+### GPU inference
 
-## Platform behavior
+GPU inference requires additional system software beyond the ONNX Runtime wheel. You also need:
 
-### Linux (CUDA)
+1. NVIDIA driver installed and working.
+2. CUDA major version compatible with your driver. We recommend 12.8 (which is what we tested at the time of writing), but check ONNX Runtime GPU wheel requirements for your version if you run into issues.
+3. cuDNN 9 for that CUDA major version.
+4. At least 10 GiB of free VRAM on each configured CUDA device for ZTF enrichment.
```
frenbox, line 58: Maybe mentioning the enrichment batch size here is relevant.

docs/gpu.md
```diff
@@ -1,63 +1,119 @@
...
 
-### macOS (CoreML / Metal)
+Then install the GPU wheel as shown above, enable GPU inference in BOOM config, and run as usual:
 
-ONNX models run via the CoreML execution provider automatically on macOS builds. Lightcurve fitting uses Metal compute shaders via the `metal` feature flag:
+```yaml
+# config.yaml
+gpu:
+  enabled: true
+  device_ids: [0]
```
frenbox, line 72: We have both tested this locally with one GPU configuration and it works fine. Were you able to test this in the system with 2 P100 in Caltech?

src/bin/scheduler.rs
```diff
@@ -21,6 +21,127 @@ use tokio::sync::oneshot;
 use tracing::{info, info_span, instrument, warn, Instrument};
 use uuid::Uuid;
 
+#[cfg(target_os = "linux")]
+const ZTF_MIN_FREE_VRAM_MIB: u64 = 10 * 1024;
```
frenbox, line 25: I think its fine for now. But ultimately this (along with batch size) should go in the config. Maybe during  implementation of the variable batch size.

src/enrichment/models/acai.rs
```diff
@@ -48,101 +48,124 @@ impl AcaiModel {
...
+
+    /// Build metadata for all valid alerts and return the original indices kept.
+    pub fn get_metadata_indexed(
+        alerts: &[&ZtfAlertForEnrichment],
+    ) -> Result<(Vec<usize>, Array<f32, Dim<[usize; 2]>>), ModelError> {
+        let mut kept_indices: Vec<usize> = Vec::new();
+        let mut features_batch: Vec<f32> = Vec::new();
+
+        for (idx, alert) in alerts.iter().enumerate() {
+            if let Ok(features) = Self::metadata_for_alert(alert) {
```
frenbox, line 73: If any error is thrown from metadata_for_alert (although highly unlikely to have missing fields), it is not captured here. May be we want to flag it?

src/enrichment/models/acai.rs
```diff
@@ -48,101 +48,124 @@ impl AcaiModel {
...
+    }
+
+    fn metadata_for_alert(alert: &ZtfAlertForEnrichment) -> Result<[f32; 25], ModelError> {
+        let candidate = &alert.candidate.candidate;
+
+        let drb = candidate.drb.ok_or(ModelError::MissingFeature("drb"))? as f32;
+        let diffmaglim = candidate
+            .diffmaglim
+            .ok_or(ModelError::MissingFeature("diffmaglim"))? as f32;
+        let ra = candidate.ra as f32;
```
frenbox, line 94: Just for my own knowledge, `ra`  `dec` ` magpsf`  `sgimapsf`  (guaranteed in schema?) `ndethist`  `ncovhist` `nmtchps` (post computed) will definitely be there and hence no need to check for `MissingFeature` ?

src/enrichment/models/base.rs
```diff
@@ -45,23 +47,37 @@ pub fn load_model_on_device(path: &str, device_id: Option<i32>) -> Result<Sessio
         .map(|v| env_truthy(&v))
         .unwrap_or(true);
 
+    #[cfg(target_os = "linux")]
+    if use_gpu && env::var_os("ORT_DYLIB_PATH").is_none() {
+        tracing::warn!(
+            "ORT_DYLIB_PATH is not set; BOOM is using the default ONNX Runtime library. If you hit cudaErrorNoKernelImageForDevice on a newer GPU, point ORT_DYLIB_PATH at a custom libonnxruntime.so built for that GPU architecture."
```
frenbox: This will still run but will take additional time right?

src/enrichment/models/base.rs
```diff
@@ -45,23 +47,37 @@ pub fn load_model_on_device(path: &str, device_id: Option<i32>) -> Result<Sessio
...
+        // Disable CPU fallback to make sure we only use the GPUs as instructed.
+        // We only do this on Linux as Apple's CoreML EP does need to fallback
+        // to the CPU for some operators of the ONNX runtime.
+        #[cfg(target_os = "linux")]
+        {
+            builder = builder.with_disable_cpu_fallback()?;
+        }
+
+        #[cfg_attr(not(target_os = "linux"), allow(unused_variables))]
         let dev = device_id.unwrap_or(0);
```
frenbox, line 68: The  `device_id` variable is passed as `null` by `load_model` and hence `device_id.unwrap_or(0)` always defaults to GPU 0. This will be an issue with multi-gpu setup systems even if you want to use only one GPU (say GPU 1).

This predates the PR i think and was never stumbled upon until now.

src/enrichment/models/base.rs
```diff
@@ -45,23 +47,37 @@ pub fn load_model_on_device(path: &str, device_id: Option<i32>) -> Result<Sessio
...
+
         builder = builder.with_execution_providers([
             #[cfg(target_os = "linux")]
-            ort::execution_providers::CUDAExecutionProvider::default()
+            ort::ep::CUDAExecutionProvider::default()
```
frenbox, line 72: This technically fixed model loading to one GPU scenario.

For multi-gpu scenario we may want to revisit multiple usage of ort `Session`

src/enrichment/models/btsbot.rs
```diff
@@ -51,95 +51,122 @@ impl BtsBotModel {
...
+    /// Build metadata for all valid alerts and return the original indices kept.
+    pub fn get_metadata_indexed(
+        alerts: &[&ZtfAlertForEnrichment],
+        alert_properties: &[AllBandsProperties],
+    ) -> Result<(Vec<usize>, Array<f32, Dim<[usize; 2]>>), ModelError> {
+        let mut kept_indices: Vec<usize> = Vec::new();
+        let mut features_batch: Vec<f32> = Vec::new();
+
+        for i in 0..alerts.len() {
+            if let Ok(features) = Self::metadata_for_alert(&alerts[i], &alert_properties[i]) {
```
frenbox, line 78: Similar to the comment on acai.rs

.github/workflows/test.yaml
petebachant: Might be best to pin it here and in the Dockerfile so it requires an explicit PR to update to avoid surprises.

Dockerfile.gpu
```diff
@@ -0,0 +1,61 @@
...
+ENV LD_LIBRARY_PATH=/opt/ort:${LD_LIBRARY_PATH}
+
+WORKDIR /app
+
+COPY --from=builder /app/target/release/scheduler /app/scheduler
+COPY --from=builder /app/target/release/migrate_fp_flux /app/migrate_fp_flux
+COPY --from=builder /app/target/release/migrate_snr /app/migrate_snr
+COPY --from=builder /opt/ort /opt/ort
+
+CMD ["/app/scheduler"]
```
petebachant, line 61: Nit: Missing newline at end of file.

README.md
```diff
@@ -98,6 +98,60 @@ code.
     cargo build --release
     ```
 
+### Linux ONNX Runtime setup
```
petebachant, line 109: Might it be better to recommend devs do testing inside containers to better mimic the production behavior? That way, there's no system-level setup required.

petebachant (approved):
This looks amazing. Kudos for figuring out all the wrinkles to get it working!

frenbox (commented):
I have left some comments. Please take a look.

src/bin/scheduler.rs
Theodlz: Yes, since there is no configurable batch size it made no sense to me to make the VRAM requirement configurable, since they are a function of each other if one is fixed the other should be too. I don't think this should ever be a configurable thing, but it should be calculated out of the configured ML batch size though.

src/enrichment/models/acai.rs
Theodlz: ndethist, ncovhist, and nmtchps are not "post-computed", these come directly from the schema! The question is: which fields are marked as optional in the avro schema, and which aren't. Those that are optional have this `MissingFeature` error.

src/enrichment/models/base.rs
Theodlz: Actually I need to update this. It now (as the README mentions) ALWAYS needs to be set. We either use dynamic loading w/ ort (in Cargo.toml) or don't, so even on CPU we need it (README also mentions that and the CI and cpu Dockerfile show how its done).

I will update this warning. We should just always look for it on linux and panic if not set. What we could do is try to be clever and look at a few known possible paths before panicking though.

src/enrichment/models/base.rs
Theodlz: We already support multi GPU systems. The scheduler checks which GPUs are there and calls load_model_on_device with the right GPU ids. This method (with null) is pretty much just the CPU variant.

src/enrichment/models/base.rs
Theodlz: this PR already allows for multiple GPU sessions where each GPU has all the models, so I'd say we already have multi GPU support (I tested it already, works great). Question is do we want to loadbalance the model accross GPUs to get around VRAM limits, at the cost of performance. We could do a little bit of both in theory.

Theodlz: but feels out of scope for this PR

README.md
Theodlz: I'm fine recommending it, but personally I always preferred developing outside of containers. On mac for example, I can't try any of the mac-specific stuff if I work in containers. So, it's not always a great idea. If we don't mention it yet, what we should for sure advertise is the option to use a container for dev. I know we have the setup, but I don't think we mention it anywhere?

docs/gpu.md
Theodlz: we don't have macos CI. If we can get mac silicon runners for free than we definitely should do that! But yes, I tested it locally. Perhaps not thoroughly enough though. Also let's keep in mind that the throughput test can't possibly use GPUs on mac (and is setup to be linux specific). We run linux containers, and those do not support metal APIs.

docs/gpu.md
Theodlz: No, I explained why it's needed in another comment. Basically getting ORT to work with recent GPUs required the dynamic loading feature. With that feature, the user needs to bring their own onnxruntime binaries (which we need anyways in the GPU case since ORT latest does not ship with them). But that means as soon as the crate's feature is turned on, one needs to go through that install. Right now that's the small price to pay for the GPU integration: extra setup on both CPU and GPU.

docs/gpu.md
Theodlz: until we make that a configurable quantity that folks can do anything about, I think it's not helpful

docs/gpu.md
Theodlz: Yep! Worked great, both had their VRAM bloated so that's how I know :laughing:

src/enrichment/models/acai.rs
Theodlz: yes we could add some error print for sure, just hoping it won't be too verbose and pollute logs but we want to know about that anyways. What you should know is that old ZTF alerts didn't have a drb and therefore runnign enrichment on them will bloat the logs like crazy. Maybe a more elegant way to do this is to collect the errors in the loop, then if we had any error print a single log that says "could not retrieve metadata features for N out of M alerts: [list of unique errors]". What do you think?

.github/workflows/test.yaml
frenbox: Makes sense!

src/enrichment/models/acai.rs
frenbox: Thanks for the explanation.

src/enrichment/models/base.rs
frenbox: Noted! We can think about in future.

frenbox (approved):

src/enrichment/models/acai.rs
frenbox: We could do that.

I was not aware the old ZTF issue.

PR #433, opened 2026-04-10
Filter worker(s): periodic active filters refresh

Theodlz:
For a while now, folks have asked "once I have uploaded a filter, when does it start running?, "how long it takes between me writing a filter and candidates to start flowing?". All valid questions, and my answer was "tell me when you want it to run and I'll restart boom so it picks it up", because the filter worker had no mechanisms to detect new filters and required a full restart.

So, this PR:
- add a configurable "filter refresh" (in minutes) to the configuration, defaults to every 15 minutes. The number I pickes is somewhat arbitrary, but it felt like a very reasonable wait time that would't disrupt real-time operations too much.
- Every N minutes (as configured), we call a new `refresh_filters` method that updated the list of active/running filters a worker is responsible of.
- now, filter workers do not just shutdown on startup if there are no filters to run, they always start and will simply sleep until they find some filters to run.

A more "clever" implementation may figure out a way to remove this fixed refresh time in favor of some implementation where we send a signal to the filter worker in real time to force a refresh, but what I like with the current implementation is the guarantee that we won't have a full refresh (which obviously means we don't process alerts while its happening) at a high cadence if folks start playing with their filters a little too fast. 

Essentially, this mimicks the behavior we had in Kowalski, and from experience folks are definitely ok with some latency between editing/adding a filter and alerts go through them. If we think 15 minutes is too long, we can make that a lot shorter using the config or ENV var overwrites anyways.

Updates:
- Copilot point out that the command interval was hardcoded to 500 everywhere, even though we have it defined in the config. I took advantage of this PR (where we add some config to parse the worker config in the filter worker anyway) to make this right, across all worker types.

### DIFF
--- a/src/filter/base.rs
+++ b/src/filter/base.rs
@@ -8,6 +8,7 @@ use crate::{
     },
 };
 
+use std::time::{Duration, Instant};
 use std::{collections::HashMap, num::NonZero, sync::LazyLock};
 
 use apache_avro::{serde_avro_bytes, Writer};
@@ -705,6 +706,14 @@ pub async fn build_loaded_filter(
     filter_collection: &mongodb::Collection<Filter>,
 ) -> Result<LoadedFilter, FilterError> {
     let filter = get_filter(filter_id, survey, filter_collection).await?;
+    if SURVEYS_REQUIRING_PERMISSIONS.contains(survey)
+        && filter
+            .permissions
+            .get(survey)
+            .is_none_or(|permissions| permissions.is_empty())
+    {
+        return Err(FilterError::InvalidFilterPermissions);
+    }
 
     let pipeline = get_active_filter_pipeline(&filter)?;
     let pipeline = build_filter_pipeline(&pipeline, &filter.permissions, &filter.survey).await?;
@@ -760,7 +769,13 @@ pub async fn build_loaded_filters(
 
     let mut filters: Vec<LoadedFilter> = Vec::new();
     for filter_id in filter_ids {
-        filters.push(build_loaded_filter(&filter_id, survey, &filter_collection).await?);
+        match build_loaded_filter(&filter_id, survey, filter_collection).await {
+            Ok(filter) => filters.push(filter),
+            Err(err) => {
+                warn!("Skipping filter {} for {:?}: {}", filter_id, survey, err);
+                continue;
+            }
+        }
     }
 
     Ok(filters)
@@ -792,6 +807,8 @@ pub enum FilterWorkerError {
     FilterNotFound,
     #[error("kafka config missing for survey: {0}")]
     KafkaConfigMissing(Survey),
+    #[error("worker config missing for survey: {0}")]
+    WorkerConfigMissing(Survey),
     #[error("Missing PSF for forced photometry point, cannot apply ZP correction")]
     MissingFluxPSF,
     #[error("missing cutouts for candid {0}")]
@@ -812,6 +829,7 @@ pub trait FilterWorker {
     ) -> Result<Self, FilterWorkerError>
     where
         Self: Sized;
+    async fn refresh_filters(&mut self) -> Result<(), FilterWorkerError>;
     fn input_queue_name(&self) -> String;
     fn output_topic_name(&self) -> String;
     fn has_filters(&self) -> bool;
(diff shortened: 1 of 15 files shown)


src/filter/base.rs
```diff
@@ -845,6 +863,10 @@ pub async fn run_filter_worker<T: FilterWorker>(
 
     let producer = create_producer(&config.kafka.producer).await?;
     let schema = load_alert_schema()?;
+    let filter_refresh_interval =
+        Duration::from_secs(worker_config.filter_refresh_interval_minutes * 60);
+    let idle_poll_interval = Duration::from_secs(30);
+    let mut next_filter_refresh = Instant::now() + filter_refresh_interval;
 
     let command_interval: usize = 500;
     let mut command_check_countdown = command_interval;
```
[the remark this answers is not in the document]
Theodlz: done, now read from config

src/filter/base.rs
```diff
@@ -845,6 +863,10 @@ pub async fn run_filter_worker<T: FilterWorker>(
 
     let producer = create_producer(&config.kafka.producer).await?;
     let schema = load_alert_schema()?;
+    let filter_refresh_interval =
+        Duration::from_secs(worker_config.filter_refresh_interval_minutes * 60);
+    let idle_poll_interval = Duration::from_secs(30);
+    let mut next_filter_refresh = Instant::now() + filter_refresh_interval;
```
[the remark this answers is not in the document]
Theodlz: we now enforce some min/max

src/filter/base.rs
```diff
@@ -878,6 +900,33 @@ pub async fn run_filter_worker<T: FilterWorker>(
...
+                        ))
+                    })?;
+                info!("termination signal received, shutting down gracefully");
+                break;
+            }
+
+            tokio::time::sleep(idle_poll_interval).await;
+            continue;
+        }
+
```
antoine-le-calloch: Maybe it’s better to use `tokio::select!` here to avoid having up to 30 seconds of delay on termination.

antoine-le-calloch (approved):
LGTM, one small and non-blocking comment.

petebachant:
Another way you could go with this is to encapsulate it within a `get_filters` method, which manages the cache TTL internally and refetches if filters are stale rather than always refreshing every interval in a separate loop. Could be a little simpler and more modular.

src/filter/base.rs
Theodlz: I removed this thing entirely, this block wasn't needed. I was basically trying to not have the filter worker pull candids until we have filters to run (hence the `continue;`). But the sleep is not very necessary especially that long, since we put no load on redis anyways (caus of that sleep). So I'll simply trade that for a 500ms sleep just like we do when we have no alerts to process. That way it all matches time and logic wise, and we don't have a blocking 30s.

Theodlz: I also standardized the shutdown check logic of all worker types

Theodlz:
> Another way you could go with this is to encapsulate it within a `get_filters` method, which manages the cache TTL internally and refetches if filters are stale rather than always refreshing every interval in a separate loop. Could be a little simpler and more modular.

I thought about this but what is the "if filters are stale" detection logic? If that means querying MongoDB at every loop of the filter worker, not sure that's a good use of our time (it will definitely impact throughput, especially since the filter documents are rather large and we may have lots of them). So, something timing based seemed simpler to me and more stable; especially since we have no constraint on how quickly filters need to be refreshed (and we know from our experience w/ Kowalski that folks are fine with a delay there, filters change very rarely to begin with).

petebachant:
> I thought about this but what is the "if filters are stale" detection logic? If that means querying MongoDB at every loop of the filter worker, not sure that's a good use of our time (it will definitely impact throughput, especially since the filter documents are rather large and we may have lots of them). So, something timing based seemed simpler to me and more stable; especially since we have no constraint on how quickly filters need to be refreshed (and we know from our experience w/ Kowalski that folks are fine with a delay there, filters change very rarely to begin with).

The `get_filters` method would track the last time it refreshed filters (as a member of the struct) and compute staleness from that. It would be similar to what you have here but only done when the filters are needed rather than every polling time. I think you'd end up querying the database less and would not have an extra background thread running.

```rust
// Rusty pseudocode
fn get_filters(self) {
    if (now() - self.last_refreshed_filters) > self.filter_ttl {
        self.refresh_filters();
        self.last_refreshed_filters = now;
    }
    self._filters
}
```

PR #435, opened 2026-04-10
adding allwise catalog

frenbox:
Add 2 arc seconds  cross match with AllWISE catalog and give the W1 W2 W3 W4 numbers of the closest match.


Theodlz (commented):
should we configure that just for ZTF (like you did), or for LSST too?

frenbox:
@Theodlz 
The current implementation was to facilitate "Oracle" classification on the outgoing kafka ztf stream here at UMN. Their ZTF classification requires the AllWISE colors. I haven't checked their LSST implementation yet. I might ultimately add them in future if required.

However, a more general question other than this specific use case is what fields do people want from AllWISE present  in Kafka? Will probably need to chat with folks about this.

Theodlz:
> @Theodlz The current implementation was to facilitate "Oracle" classification on the outgoing kafka ztf stream here at UMN. Their ZTF classification requires the AllWISE colors. I haven't checked their LSST implementation yet. I might ultimately add them in future if required.
> 
> However, a more general question other than this specific use case is what fields do people want from AllWISE present in Kafka? Will probably need to chat with folks about this.

Differences between surveys are often a source of confusion, and aside from Oracle I think I see how some folks would want W3 and W4 colros (if available) for LSST. So, what about we configure both surveys to xmatch against AllWISE?

frenbox:
Makes sense.
Should I keep all the fields I have loaded in mongo? Or curate it like I am doing right now?

antoine-le-calloch:
By the way, I don’t think we should merge this, since we’re just updating our private config with all the crossmatches. I have already added this to my config.
@Theodlz , do you want to manage this differently? If so, we would also need to merge all the other crossmatch configs we have in prod.

Theodlz:
> By the way, I don’t think we should merge this, since we’re just updating our private config with all the crossmatches. I have already added this to my config. @Theodlz , do you want to manage this differently? If so, we would also need to merge all the other crossmatch configs we have in prod.

idk, I don't have a strong stance on that matter :sweat_smile:. If the CIT and UMN instance are supposed to be alike, in general we should try to have the same config, but clearly we have already diverged in what catalogs we hold and can therefore xmatch with, so it doesn't matter all that much. I'm fine merging or not merging this one

PR #437, opened 2026-04-11
Add stats endpoints

antoine-le-calloch:
## Added a new Babamul Stats API with three main endpoints:
1. /stats/nightly (ZTF/LSST nightly alert counts with caching)
2. /stats/catalogs (catalog listing with number of entries and size)
3. /stats/kafka (Kafka topic message counts)

The endpoints are public (listed in BABAMUL_PUBLIC_ROUTES) and cached in MongoDB to absorb the UI load.

## Main changes
  - New module src/api/routes/babamul/stats/ grouping three endpoints:                                                                                                               
    - GET /babamul/stats/nightly — number of alerts per night and per survey over a date range.
    - GET /babamul/stats/catalogs — catalog list, with optional count and/or size_bytes.                                                                                             
    - GET /babamul/stats/kafka — babamul.* topics with the number of messages available and the retention period.
  - Unified caching strategy in a dedicated stats collection (added to PROTECTED_COLLECTION_NAMES):                                                                                  
    - Nightly: one entry per (survey, date) pair, with a TTL depending on the age of the night — 5 min for the current night (still in progress), 1 h between 1 and 7 days, 1 year   
  beyond that (old nights no longer change).                                                                                                                                         
    - Catalogs: a single entry, 5-day TTL                                                                                                      
    - Kafka: a single entry, 5-min TTL                                                                                                                
 - Noon-to-noon Julian Date window (date_to_jd_local_noon): an astronomical night for date D spans from JD(D, local noon) to JD(D+1, local noon), matching standard observing conventions where the night of a given date runs from that day's noon to the next day's noon in the observatory's local time. Uses a new Survey::observatory_utc_offset() helper (ZTF: UTC−7, LSST: UTC−3) so the offset isn't hardcoded in the stats module.    
  - Domain filters: for ZTF, only public alerts are counted (candidate.programid = 1). MongoDB counts rely on the compound indexes (candidate.jd, candidate.programid) to benefit from a COUNT_SCAN.                                                                                                                                                     
  - Parallelized counting of missing entries via futures::try_join_all: over all the missing days per survey run in parallel.
  - Migration from collstats to $collStats: the collstats command is deprecated on the MongoDB side. Both affected endpoints (/babamul/stats/catalogs and the existing /babamul/catalogs) now use the $collStats: { storageStats: {} } aggregation stage.


src/api/routes/babamul/stats/catalogs.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub count: Option<bool>,
+    /// Include storage size per catalog.
+    pub size: Option<bool>,
+}
+
+#[derive(Debug, Serialize, Deserialize, ToSchema)]
+pub struct CatalogStats {
+    pub n_catalogs: usize,
+    pub catalogs: Vec<CatalogEntry>,
+}
```
Theodlz, line 55: at least one top level docstring for each of these structs would be good

src/api/routes/babamul/stats/catalogs.rs
```diff
@@ -0,0 +1,212 @@
...
+
+    let mut catalog_names: Vec<String> = collection_names
+        .into_iter()
+        .filter(|name| {
+            !PROTECTED_COLLECTION_NAMES.contains(&name.as_str())
+                && !name.starts_with("system.")
+                && !name.starts_with("ZTF_")
+                && !name.starts_with("LSST_")
+        })
+        .collect();
```
Theodlz, line 128: so what collections is this endpoint meant to return then? Only the crossmatch releated ones (i.e. not the alerts) ? That's what this `.filter(...)` statement looks like at least. If so, why not just grab the catalog names from the config? Seems safer, if that's the intent

src/api/routes/babamul/stats/catalogs.rs
```diff
@@ -0,0 +1,212 @@
...
+                    })
+                    .collect::<Vec<_>>();
+                let stats = CatalogStats {
+                    n_catalogs: catalogs.len(),
+                    catalogs,
+                };
+                return response::ok_ser("catalog stats (cached)", stats);
+            }
+        }
+    }
```
Theodlz, line 110: looks like under this code block (the "no cache yet" case), you also read the collection stats conditionally. So doesn't that mean you can have a case where the cache exists doesn't have the count and size_bytes, but the user asked for them?

Theodlz: or do you only set the cache if all the details have been requested? I guess I just wonder if it wouldn't be easier to just always fetch these details (in the invalid or missing cache case) and only return what was requested to the user, that way the cache is always kept up to date and complete (whatever the request)?

src/api/routes/babamul/stats/catalogs.rs
```diff
@@ -0,0 +1,212 @@
...
+use crate::api::models::response;
+use actix_web::{get, web, HttpResponse};
+use chrono::Utc;
+use futures::StreamExt;
+use mongodb::{bson::doc, Collection, Database};
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+const CATALOG_STATS_CACHE_KEY: &str = "catalog_stats";
+/// Cache catalog stats for 5 days — reference catalogs rarely change.
```
Theodlz: what is  "reference catalog"???

src/api/routes/babamul/stats/kafka.rs
```diff
@@ -0,0 +1,127 @@
...
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+struct KafkaTopicsCacheEntry {
+    #[serde(rename = "_id")]
+    id: String,
+    topics: Vec<KafkaTopicStat>,
+    updated_at: f64,
+    cache_until: f64,
+}
```
Theodlz: same comment about struct documentation

src/api/routes/babamul/stats/kafka.rs
```diff
@@ -0,0 +1,127 @@
...
+                serde_json::json!(cached.topics),
+            );
+        }
+    }
+
+    // Cache miss — query Kafka
+    let bootstrap_servers = config.kafka.producer.server.clone();
+    let topics = match web::block(move || list_babamul_topics(&bootstrap_servers)).await {
+        Ok(Ok(t)) => t,
+        Ok(Err(e)) => {
```
Theodlz: `Ok(Err(e))` seems like a very weird pattern, same for the double Ok. I think this needs to change?

src/api/routes/babamul/stats/kafka.rs
```diff
@@ -0,0 +1,127 @@
...
+            return response::ok(
+                &format!("{} topics", cached.topics.len()),
+                serde_json::json!(cached.topics),
+            );
+        }
+    }
+
+    // Cache miss — query Kafka
+    let bootstrap_servers = config.kafka.producer.server.clone();
+    let topics = match web::block(move || list_babamul_topics(&bootstrap_servers)).await {
```
Theodlz: what does the `web::block` do, why is it needed?

Theodlz: just making sure that is not blocking to other threads in the pool

src/api/routes/babamul/stats/kafka.rs
```diff
@@ -0,0 +1,127 @@
...
+        .upsert(true)
+        .await;
+
+    response::ok(
+        &format!("{} topics", topics.len()),
+        serde_json::json!(topics),
+    )
+}
+
+fn list_babamul_topics(
```
Theodlz, line 107: is the web::block statement above there because this function isn't async? If so, can it be turned async? I guess I'm surprised because we use an async API to talk to kafka in so many places

src/api/routes/babamul/stats/kafka.rs
Theodlz, line 1: Could be nice to read info like the retention period of the topics, and include that in there. Can just read it from the config to avoid the cost of talking to kafka, and apply that to the babamul topics. If not done in this endpoint that's fine, but I think it's a quantity folks need to know about (they always ask me if there is all the survey in the topics and don't understand why things are empty at time, knowing the retention period would help a lot).

src/api/routes/catalogs.rs
```diff
@@ -52,17 +52,28 @@ pub async fn get_catalogs(
...
+                Some(Err(e)) => {
+                    return response::internal_error(&format!("Error getting catalog info: {}", e));
+                }
+                None => doc! {},
+            };
+            let details = stats
+                .get_document("storageStats")
+                .cloned()
+                .unwrap_or_default();
+            catalogs.push(doc! {"name": catalog, "details": details});
```
Theodlz, line 76: could you explain this change of API?

src/utils/enums.rs
```diff
@@ -14,6 +14,20 @@ pub enum Survey {
...
+    /// - LSST (Cerro Pachón, CL): UTC−3
+    pub fn observatory_utc_offset(&self) -> f64 {
+        match self {
+            Survey::Ztf => -7.0,
+            Survey::Lsst => -3.0,
+            _ => 0.0,
+        }
+    }
+}
+
```
Theodlz, line 69: smart! I suppose that ignores the time changes throughput the year, but probably not relevant for this implementation anyways right?

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,362 @@
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub lsst: Option<u64>,
+}
+
+#[derive(Debug, Deserialize, ToSchema)]
+pub struct StatsQuery {
+    pub start_date: String,
+    pub end_date: String,
+    pub survey: Option<String>,
+}
```
Theodlz, line 45: same comment about struct documentation

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,362 @@
...
+    jd_midnight + (12.0 - utc_offset) / 24.0
+}
+
+/// Cache duration (in seconds) grows with the age of the night.
+///
+/// - Today (age 0): 5 min — the night is still in progress, count changes fast.
+/// - Yesterday: 30 min — final count might still get a late ingest.
+/// - 2–7 days: 2 hours.
+/// - 8–30 days: 12 hours.
+/// - >30 days: 30 days — count is effectively frozen.
```
Theodlz: I wouldn't say 30 days is frozen

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,362 @@
...
+
+        let pipeline = vec![
+            doc! { "$match": match_doc },
+            doc! {
+                "$group": {
+                    "_id": { "$toInt": { "$floor": { "$subtract": ["$candidate.jd", base_jd] } } },
+                    "count": { "$sum": 1i64 }
+                }
+            },
+        ];
```
Theodlz: I see the limit is 365 days, do you have a sense of how slow this will be if a whole year of data needs to be queried? Presumably, since you max cache age is 30 days then every month most cache will be invalidated and a single user may access the stats page which will trigger such a query. Just wanna make sure it won't break everything. While I'm fine with this first PR, it kinda feels in general like these stats should be built by some watchdog/cron rather than on request, to avoid precisely that.

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,362 @@
...
+                .get("_id")
+                .and_then(|v| v.as_i64().or_else(|| v.as_i32().map(|i| i as i64)))
+                .unwrap_or(-1);
+            let count = doc
+                .get("count")
+                .and_then(|v| v.as_i64().or_else(|| v.as_i32().map(|i| i as i64)))
+                .unwrap_or(0)
+                .max(0) as u64;
+            index_counts.insert(idx, count);
+        }
```
Theodlz: could just have a struct that describes this data model, like you do everywhere else!

Theodlz (commented):
I left some comments here and there already, but here are my 2 most important one:
- In general this lacks documentation, both in code (structs should have a top level docstring ideally) but most importantly at the PR description level. This does a lot (it's not a small PR!), and the implementation is not detailed in the PR description. It's the place where you can explain why/what you did, technical choices, context, etc. If someone goes back to your code at some point, they would like to see a summary of what was implemented, why that feature was needed, and why you made some implementation choices.
- Talking about implementation choices: here you have 1 collection but 3 distinct structs to query it, i.e. 3 different data models in the same table.  I personally got a little confused and it took me some time (missing PR description as mentioned above!). I think you'll agree that on paper that doesn't sound great 😅, to have one table with 3 distinct schemas. I wonder if there isn't an opportunity to either have some kind of unified schema, or just save the cache some other way. For example, since this is a cache do you even need Mongo rather than Valkey? Just curious what you think. This isn't a blocking comment at all, but just a question about the overall implementation choices made here.

src/api/routes/babamul/stats/catalogs.rs
antoine-le-calloch: > I guess I just wonder if it wouldn't be easier to just always fetch these details

This is exactly what my code do.

Theodlz: ok then perhaps I read this wrong? This line makes it look conditional to me: https://github.com/boom-astro/boom/pull/437/changes#diff-9f0bfd1c97b7f74f4285d29240f9f9bb2850ac43507646f532252cdb11051fb2R125

src/api/routes/babamul/stats/kafka.rs
antoine-le-calloch: Yes `fetch_metadata/fetch_watermarks` are blocking C calls with no async equivalent, `src/kafka/base.rs` also use sync functions for this.

src/api/routes/babamul/stats/catalogs.rs
antoine-le-calloch: Yeah, I mean if we don’t ask for either include_count or include_size, we don’t use or update the cache at all. But if we ask for at least one of them, then we populate the cache with both.
Anyway, I’m going to change a lot of this PR since you suggested moving to Valkey, which is a great idea. I just didn’t think of it. Maybe because, in my head, having long-lived cache (more than a few days) was always intended to go into the database instead.

src/utils/enums.rs
antoine-le-calloch: Yes, it’s just to separate the night, so both 12h or 11h would work fine.

src/api/routes/babamul/stats/nightly.rs
antoine-le-calloch: What do you mean here ?

src/api/routes/babamul/stats/kafka.rs
Theodlz: right, but we should make sure that the `block` is not blocking for the whole tokio thread pool, but just for the current API call session. Otherwise, one person calling that endpoint could make the whole app slower for everybody else which is a big "freeze".

src/api/routes/babamul/stats/catalogs.rs
Theodlz: Ok! I mean please don't waste too much time on this, the current implementation works just fine. You could keep this in Mongo and consider moving to valkey later if you want. We do use persistence in Valkey hence my suggestion

Theodlz: but it might be harder to have some full-blown datastructures in Valkey than in mongo, in which case mongodb then makes more sense?

src/api/routes/babamul/stats/nightly.rs
Theodlz: I'm just saying that `count is effectively frozen` isn't true with 30 days, so I'd just rather not have that comment here.

antoine-le-calloch: The word “frozen” referred to the possible number of alerts for that night, since there’s no chance that new alerts will be added after 30 days. That’s already the case for other days, but that was the idea behind Claude’s message. Still, I agree it sounds a bit weird.

src/api/routes/babamul/stats/kafka.rs
antoine-le-calloch: `web::block` moves the blocking Kafka call to a separate thread pool so the async workers stay free to handle other requests.

src/api/routes/catalogs.rs
antoine-le-calloch: Migrated from the deprecated collStats database command (which returned a single Document) to the `$collStats` aggregation stage (which returns a Cursor), so the match arm now binds
   the cursor instead of pushing the stats doc directly.

src/api/routes/babamul/stats/kafka.rs
antoine-le-calloch: `web::block` runs the blocking call on a separate thread pool, so it only blocks that one request and keeps the async workers free for everyone else.

src/api/routes/babamul/stats/kafka.rs
antoine-le-calloch: The nesting is intentional: each arm handles a distinct failure mode with its own warn log, and size_bytes is Option<u64> — None is a valid result. We don't early-return because one catalog failing shouldn't kill the whole response.

antoine-le-calloch:
@Theodlz I have fix some stuff and increase the cache time. I keep the cache in mongo for now and maybe will move to valkey in a futur PR. You can review again this one.

Theodlz:
> @Theodlz I have fix some stuff and increase the cache time. I keep the cache in mongo for now and maybe will move to valkey in a futur PR. You can review again this one.

sounds great, I'll get to it ~today!~ as soon as I can (I overestimated my ability to crunch through some other work...)

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,335 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+struct NightlyStatCache {
+    #[serde(rename = "_id")]
+    id: String,
+    survey: String,
+    date: String,
+    n_alerts: u64,
+    updated_at: f64,
+    cache_until: f64,
+}
```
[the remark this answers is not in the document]
antoine-le-calloch: The _id is deterministic `nightly_stats_<survey>_<date>`, so upserts replace in place document. no TTL needed.

src/api/routes/babamul/stats/catalogs.rs
```diff
@@ -0,0 +1,220 @@
...
+pub struct CatalogStats {
+    pub n_catalogs: usize,
+    pub catalogs: Vec<CatalogEntry>,
+}
+
+/// Get catalog statistics.
+///
+/// By default, returns just the list of catalog names. Use `count=true` and/or
+/// `size=true` query parameters to include document counts and storage sizes.
+/// Results with counts/sizes are cached for 5 days since reference catalogs rarely change.
```
Theodlz: what are "reference catalogs"?

src/api/routes/babamul/stats/kafka.rs
```diff
@@ -0,0 +1,136 @@
...
+use mongodb::{bson::doc, Collection, Database};
+use rdkafka::config::ClientConfig;
+use rdkafka::consumer::{BaseConsumer, Consumer};
+use serde::{Deserialize, Serialize};
+use utoipa::ToSchema;
+
+const KAFKA_TIMEOUT_SECS: std::time::Duration = std::time::Duration::from_secs(10);
+const KAFKA_TOPICS_CACHE_KEY: &str = "kafka_topics";
+/// Cache Kafka topic stats for 5 minutes.
+const KAFKA_TOPICS_CACHE_SECS: f64 = 5.0 * 60.0;
```
Theodlz: nitpick: maybe better to call this `babamul_kafka_topics`, since it's not meant to contain kafka info for the BOOM topic(s). That way there is no place for confusion

src/api/routes/babamul/stats/kafka.rs
```diff
@@ -0,0 +1,136 @@
...
+        .find_one(doc! { "_id": KAFKA_TOPICS_CACHE_KEY })
+        .await
+    {
+        if cached.cache_until > now_ts {
+            return response::ok(
+                &format!("{} topics", cached.topics.len()),
+                serde_json::json!(cached.topics),
+            );
+        }
+    }
```
Theodlz, line 70: btw you could just have the cache_until > now_ts check directly in the query (that's a comment that may apply to other endpointst too), to avoid the extra if.

src/api/routes/babamul/stats/kafka.rs
```diff
@@ -0,0 +1,136 @@
...
+    let cache_entry = KafkaTopicsCacheEntry {
+        id: KAFKA_TOPICS_CACHE_KEY.to_string(),
+        topics: topics.clone(),
+        updated_at: now_ts,
+        cache_until: now_ts + KAFKA_TOPICS_CACHE_SECS,
+    };
+    let _ = stats_collection
+        .replace_one(doc! { "_id": KAFKA_TOPICS_CACHE_KEY }, &cache_entry)
+        .upsert(true)
+        .await;
```
Theodlz: is there a reason why you don't check the result of the insert?

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+use std::collections::HashMap;
+use utoipa::ToSchema;
+
+/// MongoDB cache document storing the alert count for a single survey/night,
+/// keyed by `_id` (`nightly_stats_<survey>_<date>`) with the expiration timestamp.
+#[derive(Debug, Serialize, Deserialize)]
+struct NightlyStatCache {
+    #[serde(rename = "_id")]
+    id: String,
+    survey: String,
```
Theodlz: feels like a job for an enum

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+    pub lsst: Option<u64>,
+}
+
+/// Query parameters for the nightly stats endpoint: the date range (inclusive)
+/// and an optional survey filter (`ztf` or `lsst`).
+#[derive(Debug, Deserialize, ToSchema)]
+pub struct StatsQuery {
+    pub start_date: String,
+    pub end_date: String,
+    pub survey: Option<String>,
```
Theodlz: same comment about using an enum, so it's self-documenting with the ToSchema trait

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+/// Without the `survey` query parameter, returns stats for all surveys (ZTF + LSST).
+/// ZTF counts only include public alerts (programid = 1).
+/// Results are cached in MongoDB; cache lifetime grows with the age of the night.
+#[utoipa::path(
+    get,
+    path = "/babamul/stats/nightly",
+    params(
+        ("start_date" = String, Query, description = "Start date (YYYY-MM-DD)"),
+        ("end_date" = String, Query, description = "End date (YYYY-MM-DD)"),
+        ("survey" = Option<String>, Query, description = "Optional survey filter (ztf or lsst)."),
```
Theodlz: same comment about enums

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+        Err(_) => return response::bad_request("Invalid start_date, expected YYYY-MM-DD"),
+    };
+    let end_date = match NaiveDate::parse_from_str(&query.end_date, "%Y-%m-%d") {
+        Ok(d) => d,
+        Err(_) => return response::bad_request("Invalid end_date, expected YYYY-MM-DD"),
+    };
+    if end_date < start_date {
+        return response::bad_request("end_date must be >= start_date");
+    }
+
```
Theodlz, line 107: I think there should be some maximum,  either as a range or just some minimum and/or max. Perhaps Jan 2017/2018 (no ZTF data before that) as the earliest start date?

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+    while d <= end_date {
+        all_dates.push(d);
+        d += chrono::Duration::days(1);
+    }
+
+    // Read all relevant cache entries in a single query
+    let cache_keys: Vec<String> = surveys
+        .iter()
+        .flat_map(|s| all_dates.iter().map(move |d| cache_id(s, d)))
+        .collect();
```
Theodlz, line 134: nitpick: this array could be quite large if a person asks for many years of both ZTF and LSST data. Would be good to just make sure you don't end up w/ something that's too big for mongo to take it in a find query.

Setting some theoretical min for the start_date and max for the end_date would address this

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+                    "LSST" => Survey::Lsst,
+                    _ => continue,
+                };
+                let Ok(date) = NaiveDate::parse_from_str(&doc.date, "%Y-%m-%d") else {
+                    continue;
+                };
+                cache_counts.insert((survey, date), doc.n_alerts);
+            }
+        }
+    }
```
Theodlz: here you collect all with try_collect and then you loop. Preferred method is to actually use the cursor for what it's for: loop as you stream the documents (most other endpoints should do that, so there are plenty of examples).

src/api/db.rs
```diff
@@ -7,7 +7,7 @@ use mongodb::Database;
 
 /// Protected names for operational data collections, which should not be used
 /// for analytical data catalogs
-pub const PROTECTED_COLLECTION_NAMES: [&str; 3] = ["filters", "babamul_users", "users"];
+pub const PROTECTED_COLLECTION_NAMES: [&str; 4] = ["filters", "babamul_users", "users", "stats"];
```
Theodlz: nitpick: maybe the name should be imported from stats/mod.rs? Not sure you can import a const and use it to define another const though, but worth checking.

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+            .copied()
+            .filter(|d| !cache_counts.contains_key(&(survey.clone(), *d)))
+            .collect();
+
+        if missing.is_empty() {
+            continue;
+        }
+
+        let alerts_collection = db.collection::<Document>(&format!("{}_alerts", survey));
+        let count_futures = missing.into_iter().map(|date| {
```
Theodlz, line 172: nice use of futures!

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -0,0 +1,340 @@
...
+            let survey = survey.clone();
+            async move {
+                coll.count_documents(filter)
+                    .await
+                    .map(|c| (survey, date, c))
+            }
+        });
+
+        let results: Vec<(Survey, NaiveDate, u64)> = match futures::stream::iter(count_futures)
+            .buffer_unordered(30)
```
Theodlz, line 189: what does the buffer unordered do?

Theodlz: max nb of threads in parallel maybe? Just curious

src/api/routes/babamul/stats/nightly.rs
Theodlz: obviously the survey enum has ztf, lsst, and decam (and here we just care about ztf and lsst), so that would require a dedicated BabamulSurvey enum if we don't have one yet (if you do want to go that route)

src/utils/enums.rs
```diff
@@ -14,6 +14,20 @@ pub enum Survey {
...
+impl Survey {
+    /// Observatory UTC offset in hours.
+    ///
+    /// - ZTF  (Palomar, CA):      UTC−7
+    /// - LSST (Cerro Pachón, CL): UTC−3
+    pub fn observatory_utc_offset(&self) -> f64 {
+        match self {
+            Survey::Ztf => -7.0,
+            Survey::Lsst => -3.0,
+            _ => 0.0,
```
Theodlz: is it really 0 for decam? I think they are on CLT (https://en.wikipedia.org/wiki/Cerro_Tololo_Inter-American_Observatory), so UTC - 4

Theodlz (commented):
only minor comments, mostly nitpick!  I know we need this feature so I'll go ahead and approve it. Please feel free to address comments you think are relevant (if any), prior to merging

Theodlz (approved):
see my comment above

src/api/routes/babamul/stats/nightly.rs
antoine-le-calloch: It polls up to N futures concurrently on the same task, yielding each result as soon as it's ready.

antoine-le-calloch:
> see my comment above

@Theodlz Done !

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -1,13 +1,12 @@
+use super::super::BabamulSurvey;
```
Theodlz: super super is quite silly looking hahaha

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -102,9 +72,9 @@ fn cache_duration_secs(date: &NaiveDate, today: &NaiveDate) -> f64 {
     get,
     path = "/babamul/stats/nightly",
     params(
-        ("start_date" = String, Query, description = "Start date (YYYY-MM-DD)"),
-        ("end_date" = String, Query, description = "End date (YYYY-MM-DD)"),
-        ("survey" = Option<String>, Query, description = "Optional survey filter (ztf or lsst)."),
+        ("start_date" = String, Query, description = "Start date (YYYY-MM-DD); must be on or after 2018-01-01."),
+        ("end_date" = String, Query, description = "End date (YYYY-MM-DD); cannot be in the future."),
```
Theodlz, line 76: smart!

src/api/routes/babamul/stats/nightly.rs
```diff
@@ -118,12 +88,8 @@ pub async fn get_nightly_stats(
...
-    let surveys = match query.survey.as_deref() {
-        Some("ztf") => vec![Survey::Ztf],
-        Some("lsst") => vec![Survey::Lsst],
-        Some(_) => {
-            return response::bad_request("Invalid survey, expected 'ztf' or 'lsst'");
-        }
+    let surveys: Vec<Survey> = match query.survey {
+        Some(s) => vec![s.into()],
         None => vec![Survey::Ztf, Survey::Lsst],
     };
```
Theodlz: might work with a simpler `unwrap_or()`, if the compiler doesn't complain

src/api/routes/babamul/mod.rs
Theodlz, line 1: very nice!

Theodlz (commented):
LGTM! This is great. No blocking comments, only one nitpick but let's merge anyways

antoine-le-calloch:
Done !

PR #440, opened 2026-04-15
[Draft] add healpix indices upon injestion

mcoughlin:
Draft PR to discuss ideas around healpix-based queries and sharding.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -94,5 +94,18 @@ jobs:
         docker compose logs kafka-acl-init || true
         docker compose logs broker || true
         exit 1
+    - name: Start sharding demo (2 MongoDB instances)
+      run: |
+        docker compose -f tests/sharding-demo/docker-compose.yaml up -d
+        echo "Waiting for shard instances..."
+        for i in {1..30}; do
+          if docker compose -f tests/sharding-demo/docker-compose.yaml exec -T mongo-shard-0 mongosh --quiet --eval "db.runCommand('ping').ok" >/dev/null 2>&1 && \
+             docker compose -f tests/sharding-demo/docker-compose.yaml exec -T mongo-shard-1 mongosh --quiet --eval "db.runCommand('ping').ok" >/dev/null 2>&1; then
+            echo "Both shard instances ready!"
+            break
+          fi
+          echo "Attempt $i/30..."
+          sleep 2
+        done
     - name: Test BOOM
       run: PATH=/opt/kafka/bin:$PATH cargo test --release
--- a/config.yaml
+++ b/config.yaml
@@ -20,6 +20,11 @@ babamul:
   enabled: false # Set via BOOM_BABAMUL__ENABLED environment variable
   webapp_url: # Set via BOOM_BABAMUL__WEBAPP_URL environment variable
   retention_days: 3
+healpix:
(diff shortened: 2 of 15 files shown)


mcoughlin:
@antoine-le-calloch @Theodlz @frenbox @petebachant Something to be thinking about.

petebachant:
A few thoughts/questions from my end (please correct me if any assumptions are incorrect!):

1. To make use of these the cross-matching, filtering, and API queries we'll need to then precompute healpix indices, right?
2. If this is for storage reasons, we'll only want to start sharding the database once we get to the point where the whole thing won't fit on a single drive/node. Perhaps we start deleting old data before that becomes an issue. To my knowledge, we are not making any promises yet about alert history availability.
3. I assume archival catalogs will never be so big they'd need to be sharded for cross-matching (sharding by catalog is an option too, since we iterate over catalogs), so that means this is mostly beneficial if we're storing huge quantities of alerts for querying later. Filtering is done on a group of alert IDs at a time, so indexing won't speed that up. Maybe using a sequential or time-based index would help there, if it's a bottleneck. On the other hand, I still don't have a good feel for what filters are typically of interest, so I may be missing something about how they make use of joined data.
4. It would be great to tie this back to user stories if possible, e.g., "I want to be able to find all alerts from objects in a given region of the sky within the last month so that..." or "I want to be able to see all alerts for a given object from ZTF, LSST, and Argus so that...".

mcoughlin:
If this is for storage reasons, we'll only want to start sharding the database once we get to the point where the whole thing won't fit on a single drive/node. Perhaps we start deleting old data before that becomes an issue. To my knowledge, we are not making any promises yet about alert history availability.
 -> I think we will always want to be able to support archival searches. If we want to do any kind of rates papers, we will want to see what happens if we run XXX filter over historical data, so we really would like to avoid this.

Theodlz:
@mcoughlin about your last commit: since we run the throughput test over a database that already has data, crossmatches are basically almost never computed. If you really want to benchmark that, might be worth commenting out the pieces of code that load the DB with history, and see how things look then.

Or, just have a separate benchmark script. It's not like real data is needed to measure that.

Theodlz:
> @mcoughlin about your last commit: since we run the throughput test over a database that already has data, crossmatches are basically almost never computed. If you really want to benchmark that, might be worth commenting out the pieces of code that load the DB with history, and see how things look then.
> 
> Or, just have a separate benchmark script. It's not like real data is needed to measure that.

I have to say, I'm still a little confused why we are trying to make all these changes in boom before assessing there is a clear advantage to using healpixels to run conesearches in Mongo? Feels like this should be its own little repo that benchmarks query speed of these 2 approaches, vs disk usage of the collections and indexes when using one or the other... It just feels like trying this in BOOM directly is skipping a lot of steps and adding so much complexity on top of this problem that one can't actually measure which is best.

mcoughlin:
@frenbox @antoine-le-calloch is any of this still useful for your testing? If not, I think we just close it.

frenbox:
Parts of this can still be relevant.
Can you explain me why you went to with Nside = 16? That creates too many healpix pixel. Sharding on that level is burdensome and computationally straining.

mcoughlin:
@frenbox yeah this predated all of our subsequent conversations on the topic. I definitely think this is no longer the right choice!

frenbox:
I am closing this PR then. We will get back to it if needed.

PR #450, opened 2026-04-25
Restrict filter creation

antoine-le-calloch:
- Set filter active status to false by default on creation
- New filter activation guard: added validate_filter_activation, which runs the pipeline against reference observing nights (2026-03-16 for ZTF, 2026-02-23 for LSST) and rejects it if it matches more than 5% of the alerts the user can access that night.
- Enforced on POST /filter/version: when the new version will replace a running active pipeline, the validation run.                                                                                                                                                               
- Enforced on PATCH /filter: if the filter become active or execution changes, the resulting pipeline is
  re-validated before the update is applied.
  
@Theodlz feel free to update the MAX_FILTER_RESULT_RATIO/testing date as you want.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -43,31 +43,37 @@ kafka:
 workers:
   ztf:
     command_interval: 500
-    filter_refresh_interval_minutes: 15
     alert:
       n_workers: 3
     enrichment:
       n_workers: 3
     filter:
       n_workers: 3
+      refresh_interval_minutes: 15
+      # Activation check: a filter cannot be activated if it matches more than
+      # max_match_rate (out of 100) of the alerts accessible on reference_night.
+      max_match_rate: 25
+      reference_night: "2026-03-16"
   lsst:
     command_interval: 500
-    filter_refresh_interval_minutes: 15
     alert:
       n_workers: 4
     enrichment:
       n_workers: 3
     filter:
       n_workers: 1
+      refresh_interval_minutes: 15
+      max_match_rate: 5
+      reference_night: "2026-02-23"
   decam:
     command_interval: 500
-    filter_refresh_interval_minutes: 15
     alert:
       n_workers: 1
     enrichment:
       n_workers: 0
     filter:
       n_workers: 1
+      refresh_interval_minutes: 15
 crossmatch:
   ztf:
     - catalog: PS1_DR1
--- a/config.yaml
+++ b/config.yaml
@@ -52,7 +52,7 @@ workers:
       refresh_interval_minutes: 15
       # Activation check: a filter cannot be activated if it matches more than
       # max_match_rate (out of 100) of the alerts accessible on reference_night.
-      max_match_rate: 25
+      max_match_rate: 5
       reference_night: "2026-03-16"
   lsst:
     command_interval: 500
--- a/config.yaml
+++ b/config.yaml
@@ -74,6 +74,8 @@ workers:
     filter:
       n_workers: 1
       refresh_interval_minutes: 15
+      max_match_rate: 5
+      reference_night:
 crossmatch:
   ztf:
     - catalog: PS1_DR1
--- a/config.yaml
+++ b/config.yaml
@@ -50,8 +50,8 @@ workers:
     filter:
       n_workers: 3
       refresh_interval_minutes: 15
-      # Activation check: a filter cannot be activated if it matches more than
-      # max_match_rate (out of 100) of the alerts accessible on reference_night.
+      # On activation, filters are validated against a designated reference night.
+      # If more than max_match_rate percent of alerts pass, the filter is not validated.
       max_match_rate: 5
       reference_night: "2026-03-16"
   lsst:
--- a/src/api/routes/filters.rs
+++ b/src/api/routes/filters.rs
@@ -8,6 +8,7 @@ use crate::{
         models::response,
         routes::users::User,
     },
+    conf::{AppConfig, FilterWorkerConfig},
     enrichment::{LsstAlertProperties, ZtfAlertClassifications, ZtfAlertProperties},
     filter::{
         build_filter_pipeline, Filter, FilterError, FilterVersion, SURVEYS_REQUIRING_PERMISSIONS,
@@ -21,7 +22,6 @@ use crate::{
 use actix_web::{get, patch, post, web, HttpResponse};
 use apache_avro::AvroSchema;
 use apache_avro_macros::serdavro;
-use chrono::NaiveDate;
 use flare::Time;
 use futures::stream::StreamExt;
 use mongodb::{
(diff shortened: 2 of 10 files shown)


src/api/routes/filters.rs
```diff
@@ -123,6 +127,116 @@ async fn build_and_test_filter_version(
...
+fn activation_reference_night(survey: &Survey) -> Result<NaiveDate, String> {
+    match survey {
+        Survey::Ztf => Ok(NaiveDate::from_ymd_opt(2026, 3, 16).unwrap()),
+        Survey::Lsst => Ok(NaiveDate::from_ymd_opt(2026, 2, 23).unwrap()),
+        _ => Err(format!(
+            "filter activation validation is not supported for survey {}",
+            survey
+        )),
+    }
+}
```
Theodlz: kinda feels like hardcoding this could be an issue, to test stuff in a dev setup where one may not have these nights. This is nitpick btw, completely fine for this PR as long as if the night has no data it still let's me post filters (otherwise I can't test this out in my dev setup!)

src/api/routes/filters.rs
```diff
@@ -123,6 +127,116 @@ async fn build_and_test_filter_version(
...
+
+/// Validate that activating this filter is safe by running it against a
+/// reference observing night and ensuring the filter does not match more than
+/// `MAX_FILTER_RESULT_RATIO` of the alerts the filter has access to that night.
+async fn validate_filter_activation(
```
Theodlz: there is a bit of overlap between this method and `run_test_pipeline` (https://github.com/boom-astro/boom/blob/main/src/api/routes/filters.rs#L67-L114) and in general we might as well just validate which tells us that a filter is valid (can run) and doesn't select a ton of garbage. So, I wonder if `run_test_pipeline` should just be replaced by this?

Theodlz: if so - which let's us avoid querying the DB twice for one validation needed - then we must not forget to bring some of the validation checks from `run_test_pipeline` into this method.

Theodlz: btw run_test_pipeline has a wrapper around it: `build_and_test_filter_version`, just for context. that's the method that the filter post and filter version post endpoints already call

Theodlz: oh unless you are trying to decouple saving a filter and activating it, and its only for activation that you wish to validate that the filter output is sensible? Makes sense then that you've set it up this way!

src/api/routes/filters.rs
```diff
@@ -123,6 +127,116 @@ async fn build_and_test_filter_version(
     run_test_pipeline(db, survey, test_pipeline).await
 }
 
+const MAX_FILTER_RESULT_RATIO: f64 = 0.20;
```
Theodlz: why not directly store it as a percentage? I see that later you just multiply it by 100 everytime you use that quantity.

Theodlz: also 20% of a stream is a F*CK ton. Most filters let a tiny fraction through. If you think about it, 10% of LSST is still like 1M alerts which is crazy high! I'd say we use 5% for now (which is still probably too high). Could be good to just make this a configurable quantity with 5% as a default? That way we can adjust it without a PR. Long term I could imagine us wanting to have this based on the survey (caus its fine to have 5% of ZTF but 5% of LSST passing one single filter is probably way too much already).

src/api/routes/filters.rs
Theodlz: similarly that could be configured along with the percentage we want to allow to go through one filter.

src/api/routes/filters.rs
Theodlz, line 133: very nice!

src/utils/db.rs
```diff
@@ -48,6 +49,31 @@ pub fn cutout2bsonbinary(cutout: Vec<u8>) -> mongodb::bson::Binary {
...
+    db: &Database,
+    survey: &Survey,
+    date: &NaiveDate,
+    programids: Option<&[i32]>,
+) -> Result<u64, mongodb::error::Error> {
+    let (start_jd, end_jd) = survey.night_jd_window(date);
+    let mut filter = doc! {
+        "candidate.jd": { "$gte": start_jd, "$lt": end_jd },
+    };
+    if let Some(pids) = programids {
```
Theodlz: maybe add `& Survey == Survey::Ztf`? since that's the only one where we have programids to care about

src/utils/enums.rs
```diff
@@ -14,6 +15,56 @@ pub enum Survey {
...
+
+    /// JD window `[start, end)` for the observing night labelled by `date`,
+    /// running from local noon of `date` to local noon of `date + 1`.
+    pub fn night_jd_window(&self, date: &NaiveDate) -> (f64, f64) {
+        let start = self.date_to_jd_local_noon(date);
+        let end = self.date_to_jd_local_noon(&(*date + chrono::Duration::days(1)));
+        (start, end)
+    }
+}
+
```
Theodlz, line 69: some overlap with you stats PR. Let's make sure what we don't duplicate some helper methods, so we should keep in mind to be careful when resolving conflitcts (after one of them gets merged first).

Theodlz (commented):
just some minor comments, otherwise this is great!

src/utils/enums.rs
antoine-le-calloch: Yes, and that was intentional. I centralized everything in the enum so it can be reused in the stats PR. I also aligned the implementation with the existing code there to ensure compatibility across both PRs.

src/api/routes/filters.rs
antoine-le-calloch: You can always POST a filter, but you just can't activate it if there is no data for this date. I will add this to the config.

src/api/routes/filters.rs
antoine-le-calloch: Yep, 20% was just a placeholder while waiting for what you think is a good value.
I’m changing it to a different one per survey.
But for now, since LSST has a very small amount of data, I’m a bit worried that using 5% will be too restrictive.

Theodlz: Fair enough, though I would argue that folks need to get used to being selective (earlier this year when we suddenly hot million of alerts, filters that seemed ok before just f*cked us up). But, I think this is why making that quantity configurable is useful: we can adapt it as LSST ramps up, without redeploying. I'm happy to take care of the config stuff btw, if you already have some other things on your plate!

antoine-le-calloch: I already almost finish it ahah.

antoine-le-calloch:
@Theodlz All comments have been addressed, and all filter properties have been moved to the filter config.

src/conf.rs
```diff
@@ -458,18 +459,51 @@ where
...
+    /// if either is missing, filters cannot be activated.
+    #[serde(default, deserialize_with = "deserialize_max_match_rate")]
+    pub max_match_rate: Option<u8>,
+    /// Reference observing night used to gauge how selective a filter is.
+    /// Should be a recent, well-populated night for the survey. Required
+    /// alongside `max_match_rate` to allow filter activation on this survey;
+    /// if either is missing, filters cannot be activated.
+    #[serde(default)]
+    pub reference_night: Option<NaiveDate>,
+}
```
Theodlz, line 498: one might argue that if they are required to run boom, these quantities could just not be `Option<...>`, but that depends if you want to display a custom error message or let serde error out for you. If it's for readability/interpretability, than that design choice you made makes sense?

config.yaml
```diff
@@ -43,31 +43,39 @@ kafka:
...
-    filter_refresh_interval_minutes: 15
     alert:
       n_workers: 3
     enrichment:
       n_workers: 3
     filter:
       n_workers: 3
+      refresh_interval_minutes: 15
+      # Activation check: a filter cannot be activated if it matches more than
+      # max_match_rate (out of 100) of the alerts accessible on reference_night.
```
Theodlz: feels like this could be rewritten maybe? 

"On insert/update, filters are validated against a `reference_night`. If more than `max_match_rate` percent of alerts pass, the filter is not validated."

Theodlz (approved):
LGTM!

src/conf.rs
antoine-le-calloch: The thing is, how do you handle adding a new survey with no historical data? The user can’t really select a `reference_night`. Right now, when no data is found, the filter can’t be tested and therefore can’t be activated. Is it ok to just block adding filters when there is no data? Because if we allow filters in the rare case where there’s no data, those filters could be very broad and will never actually be verified.

Theodlz: good question... I'd agree with you, posting filters when there is no data is a bit odd. But on a new instance someone is setting up, why not?

Theodlz: we can think about this in a dedicated issue maybe

PR #460, opened 2026-05-03
[DNM] binned lightcurves

mcoughlin:
This PR explores the idea of light curve binning in a source-class dependent way. Basic idea is to have time periods over which light curve binning occurs to save space in dense light curves (when that high density of observations does not have much information in them).


mcoughlin:
@frenbox same question here.

frenbox:
I feel like we need to discuss with the Argus team about the data delivery rates and cadence first to warrant binning and different approaches.

PR #461, opened 2026-05-04
Add reprocess_crossmatch binary

antoine-le-calloch:
Add reprocess_crossmatch binary + scheduler crossmatch coverage check:

- `reprocess_crossmatch` backfills `cross_matches.<catalog>` on existing aux records. Useful when a catalog is added or its parameters are changed. Supports `--direction objects|catalog|auto` to loop over the smaller side of the spatial join.
- Parallelize `reprocess_crossmatch` with a worker pool (`--processes`) using an async channel, and batch all object-driven catalogs into a single xmatch pass.
- Process catalog-driven mode into 5 phases with a buffer field `cross_matches_temp.<catalog>`, ensuring `cross_matches.<catalog>` is never empty mid-run. Writes are gated by `created_at < run_start_jd` to avoid touching live inserts.
- Scheduler boot samples one aux record per survey and warns if any configured catalog is missing from `cross_matches`, pointing to the new binary as the fix with already populate arguments.
- Extract shared helpers used by `xmatch`, the new binary, and the `migrate_*` binaries: `cm_radius_arcsec`, `distance_kpc_from_arcsec`, `get_f64_from_doc` in `utils::spatial`, `make_progress_bar` in `utils::data`.

### DIFF
--- a/src/bin/reprocess_crossmatch.rs
+++ b/src/bin/reprocess_crossmatch.rs
@@ -5,22 +5,28 @@ use boom::{
     utils::{
         data::make_progress_bar,
         enums::Survey,
-        spatial::{cm_radius_arcsec, distance_kpc_from_arcsec, get_f64_from_doc, xmatch},
+        spatial::{
+            cm_radius_arcsec, distance_kpc_from_arcsec, get_f64_from_doc, xmatch, Coordinates,
+        },
     },
 };
 use clap::{Parser, ValueEnum};
-use flare::spatial::great_circle_distance;
+use flare::{spatial::great_circle_distance, Time};
 use futures::TryStreamExt;
+use indicatif::ProgressBar;
 use mongodb::{
-    bson::{doc, Bson, Document},
+    bson::{doc, Document},
     options::{UpdateOneModel, WriteModel},
+    Namespace,
 };
 use tracing::{error, info, warn, Level};
 use tracing_subscriber::FmtSubscriber;
 
-/// Binary for reprocessing crossmatches between a survey's aux collection and one or more catalogs.
+const QUEUE_MULTIPLIER: usize = 2;
+
+/// Binary for reprocessing crossmatches between a survey's alerts_aux collection and one or more catalogs.
 /// The scheduler pipeline only crossmatches at first insert, so adding a catalog to
-/// `crossmatch.<survey>` in config.yaml leaves pre-existing aux records with
+/// `crossmatch.<survey>` in config.yaml leaves pre-existing alerts_aux records with
 /// no entry for it, so this binary fills in those gaps. It can also be used to reprocess existing
 /// crossmatches if the matching parameters (e.g. radius) for a catalog are changed.
 #[derive(Parser)]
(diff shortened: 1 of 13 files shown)


src/bin/scheduler.rs
```diff
@@ -142,6 +144,70 @@ fn validate_gpu_free_vram(
     Ok(())
 }
 
+/// Sample one aux record at random and warn if it's missing crossmatches for
+/// any catalog declared under `crossmatch.<survey>` in the config. The live
+/// pipeline only crossmatches at first insert, so newly added catalogs never
+/// reach pre-existing records — the user has to run `reprocess_crossmatch`.
+async fn warn_if_missing_crossmatches(survey: &Survey, db: &mongodb::Database, config: &AppConfig) {
```
Theodlz, line 151: That is such a nice touch! I really appreciate you going the extra mile here.

src/bin/reprocess_crossmatch.rs
```diff
@@ -0,0 +1,474 @@
...
+    #[arg(long, value_delimiter = ',', num_args = 1..)]
+    catalogs: Vec<String>,
+
+    #[arg(long, value_enum, default_value_t = Direction::Auto)]
+    direction: Direction,
+
+    #[arg(long, value_name = "FILE")]
+    config: Option<String>,
+
+    #[arg(long, default_value = "5000")]
```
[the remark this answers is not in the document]
Theodlz: +1

src/bin/reprocess_crossmatch.rs
```diff
@@ -0,0 +1,474 @@
...
+    // populated their cross_matches via the live xmatch.
+    let pre_existing = doc! { "created_at": { "$lt": run_start_jd } };
+
+    let empty_array: Vec<Document> = Vec::new();
+    aux_collection
+        .update_many(
+            pre_existing.clone(),
+            doc! { "$set": { &cat_field: empty_array } },
+        )
+        .await?;
```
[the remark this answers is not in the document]
Theodlz: I feel like this messes up existing data unnecessarily. I can see us rerunning this because there was an issue with the xmatch or not all objects have it yet, and wiping data clean off existing data isn't all that great. Also `pre_existing` is equal to the current time and obviously there are no documents more recent than now, so that doesn't do anything meaningful! Might as well just ran on every object, it's not like you will interfere with regular operations when you just update cross matches.

Theodlz (commented):
Looks good, I just have 2 comments:
- one about wiping data from existing objects, not sure I understand why we want that
- more generally, the lack of parallelization. I am afraid this won't work well for us at scale  (we have many very large catalogs) without it. https://github.com/boom-astro/boom/pull/348 took a crack at it, might want to recycle or be inspired by the parallelization that happens there.

For future PRs: just like we run xmatches for 1 object against N catalogs, we may be able to run xmatches for N objects against M (where M can be 1) catalog using a big aggregation pipeline to save some DB roundtrips. It's an optimization that's not *required* for this to work well, so I wouldn't worry about it here. Can be explored later

src/bin/scheduler.rs
antoine-le-calloch: Yeah, since it wasn’t even clear to me, I thought it would be useful to test it and provide that information. Also, having the already populated command line to execute the fix is very helpful.

src/bin/reprocess_crossmatch.rs
```diff
@@ -0,0 +1,696 @@
...
+// (`cross_matches.<catalog>_temp`) as a buffer so the `cross_matches.<catalog>` field is
+// never empty mid-run.
+//
+// Concurrency with the live ingest pipeline: every phase is gated on
+// `created_at < run_start_jd` so records inserted during the run are left
+// completely untouched (the scheduler pipeline already filled their cross_matches).
+// Without this guard, phase 4's `$set live = $temp` would overwrite a new record's
+// field with a missing/partial temp and silently delete it.
+// Phase 5 then cleans up any orphan temp left.
+// -----------------------------------------------------------------------------
```
Theodlz, line 227: Might want to update some of these docstring. For example its a little weird that you refer to a phase 4 and 5, but no mention of 1,2,3 in that same docstring?

Theodlz: I know you mention them in code comments underneath, but it's still a little odd to mention just some of the numbers. Either all of them should be in there, or none.

src/bin/reprocess_crossmatch.rs
```diff
@@ -0,0 +1,696 @@
...
+        }
+    }
+    pb.finish();
+    if let Some(e) = first_err {
+        // CRITICAL: phase 1 cleared temp on every existing record. If phase 2 partially failed,
+        // some records have empty temp; running phase 4 would overwrite their valid live
+        // cross_matches with that empty temp. Abort before phase 3 to preserve existing data.
+        // Phase 5-style cleanup of any orphan temp must still run on retry.
+        return Err(e);
+    }
```
Theodlz, line 330: nice, I appreciate all the code comments, it's all super clear!

src/bin/reprocess_crossmatch.rs
```diff
@@ -0,0 +1,696 @@
...
+    );
+    aux_collection
+        .update_many(
+            existing_records,
+            vec![
+                doc! { "$set": { &live_field: format!("${}", &temp_field) } },
+                doc! { "$unset": &temp_field },
+            ],
+        )
+        .await?;
```
Theodlz, line 361: what about records that never had a `_temp` entry set (because no catalog sources matched with them), will that set the field to null, empty array, not set it, ...?

Theodlz: I guess question is, for those where we didn't add a temp because they were no matches:
- if they already had crossmatches set on them, should we delete them
- should we set an empty array for those with no matches?

Just making sure we don't end up with something inconsistent with what the pipeline produces, when we process alerts in real time.

src/bin/reprocess_crossmatch.rs
```diff
@@ -0,0 +1,696 @@
...
+    aux_collection
+        .update_many(
+            doc! { &temp_field: { "$exists": true } },
+            doc! { "$unset": { &temp_field: "" } },
+        )
+        .await?;
+
+    Ok(())
+}
+
```
Theodlz: since you already have the previous query run `unset` for the temp field, is that really needed? Never a bad idea to run a cleanup, but queries on millions of records can be slow so might as well avoid them if they are unlikely to actually do anything.

Theodlz (commented):
some minor items!

src/bin/reprocess_crossmatch.rs
antoine-le-calloch: The idea behind Phase 5 was to make sure we cleaned up any temp fields, including on records inserted after run_start_jd, since Phase 4's query is gated on `created_at < run_start_jd` and skips them. But I agree, it's an extra check we don't really need. Let's drop it.

src/bin/reprocess_crossmatch.rs
antoine-le-calloch: Already correct. Phase 1 sets temp = [] on every existing record and Phase 4 unconditionally copies that to cross_matches.<cat>, so no-match records end up as [], like the live pipeline.

src/bin/reprocess_crossmatch.rs
antoine-le-calloch: Done.

src/bin/reprocess_crossmatch.rs
Theodlz: Oh i missed that phase 1 set all temp to `[]`, this is great.

Theodlz (approved):
LGTM 🚀 . Awesome work @antoine-le-calloch

PR #465, opened 2026-05-06
Only log passing filter

antoine-le-calloch:
Update filter processing to log "passed filter" only when at least one doc pass the filter.

### DIFF
--- a/src/filter/ztf.rs
+++ b/src/filter/ztf.rs
@@ -712,6 +712,13 @@ impl FilterWorker for ZtfFilterWorker {
                 )
                 .await?;
 
+                // If we have output documents, we need to process them
+                // and create filter results for each document (which contain annotations)
+                // however, if the array is empty, there's nothing to do
+                if out_documents.is_empty() {
+                    continue;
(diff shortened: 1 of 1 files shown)


Theodlz:
@antoine-le-calloch one thing to take note of: the throughput test relies on filter logs (which should be the last that are logged in the app) to get the timestamp of the end of processing, so we can compute the wall time.

My suggestion would be to log everytime but at the debug level, and make sure the throughput test sets the log level to debug. That way no problem for production (where I doubt you have the log level set to debug) nor throughput testing/development?

antoine-le-calloch:
> @antoine-le-calloch one thing to take note of: the throughput test relies on filter logs (which should be the last that are logged in the app) to get the timestamp of the end of processing, so we can compute the wall time.
> 
> My suggestion would be to log everytime but at the debug level, and make sure the throughput test sets the log level to debug. That way no problem for production (where I doubt you have the log level set to debug) nor throughput testing/development?

Ohh yeah nice catch, sorry for that.

PR #467, opened 2026-05-06
Improve readme to fit apptainer readme

antoine-le-calloch:
Update the README to be more close to what I have on the boom apptainer side to simplify when merging the change from the main:
- Improve README to include detailed instructions for production and benchmarking with Docker.
- Fix the link to README section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ RUST_LOG=debug,ort=warn BOOM_SPAN_EVENTS=new,close cargo run --bin scheduler -- 
 
 This repository includes a benchmark to test the system and get an idea of the time it takes to process a certain number of alerts.
 This benchmark uses Docker to build the image and run the benchmark.
-The step to run the benchmark are as follows:
+The steps to run the benchmark are as follows:
 
 ### Build Image
 For Docker (docker Image):
--- a/README.md
+++ b/README.md
@@ -323,9 +323,21 @@ docker buildx bake -f tests/throughput/compose.yaml --load
 ```
 
 ### Download Data
-```bash
-mkdir -p ./data/alerts/ztf/public/20250311 # target directory where the producer will download the ZTF archive data
+
+Download a ZTF alerts aux data archive
+
+**For Linux:**
+```
 wget -q https://caltech.box.com/shared/static/qdois5qq2lmvp02ri50fum80vzr54505.gz -O ./data/alerts/boom_throughput.ZTF_alerts_aux.dump.gz
+```
+**For macOS:**
+```
+curl -sL https://caltech.box.com/shared/static/qdois5qq2lmvp02ri50fum80vzr54505.gz -o ./data/alerts/boom_throughput.ZTF_alerts_aux.dump.gz
+```
+
+Download the NED catalog for crossmatching.
+```
+mkdir -p ./data/alerts/ztf/public/20250311 # target directory where the producer will download the ZTF archive data
 uvx gdown "https://drive.google.com/uc?id=1BG46oLMbONXhIqiPrepSnhKim1xfiVbB" -O ./data/alerts/kowalski.NED.json.gz
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -281,6 +295,34 @@ As a more complete example, the following sets the logging level to DEBUG, with
...
+This benchmark uses Docker to build the image and run the benchmark.
+The step to run the benchmark are as follows:
+
+### Build Image
+For Docker (docker Image):
+```bash
+  docker buildx create --use
+  docker buildx inspect --bootstrap
+  docker buildx bake -f tests/throughput/compose.yaml --load
+```
```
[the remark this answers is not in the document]
antoine-le-calloch: Do we want to keep this ?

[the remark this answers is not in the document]
Theodlz: definitely don't need that, I mean I never even looked at it.

README.md
```diff
@@ -234,6 +229,38 @@ For example, to process ZTF alerts, you can run:
...
+docker compose --profile prod run --rm consumer-ztf /app/kafka_consumer ztf 20240617 --programids public
+docker compose --profile prod run --rm scheduler-ztf /app/scheduler ztf
+```
+
+To tail logs or open a shell in a running container:
+
+```bash
+docker compose logs -f scheduler-ztf
+docker compose exec scheduler-ztf /bin/bash
+```
```
petebachant, line 396: Is it common to single-shot apps in production? If it happens a lot it might be worth adding the ability to run these kinds of jobs to an admin page/API to reduce exposure and improve traceability (can log them in the database; sort of like migrations). I'll throw an item on the backlog for discussion (#471)

antoine-le-calloch: I often do it on my instance, but with Apptainer, so it's kind of different. @Theodlz is it also your case? Since there is one container per process (scheduler - consumer), I imagine not.

On my end, I have one BOOM container per survey running both the scheduler and the consumer, which is why this happens. I can remove the Docker line that does it if you want.

README.md
```diff
@@ -57,7 +58,7 @@ BOOM runs on macOS and Linux. You'll need:
   sudo apt install build-essential pkg-config libssl-dev libsasl2-dev -y
   ```
 
-- If you want to use GPU hardware acceleration for enrichment, you need to have the appropriate NVIDIA drivers installed, along with CUDA and cuDNN. See the [GPU inference](#gpu-inference-linux) subsection below for more details.
+- If you want to use GPU hardware acceleration for enrichment, you need to have the appropriate NVIDIA drivers installed, along with CUDA and cuDNN. See the [Linux GPU inference](#gpu-inference) subsection below for more details.
```
Theodlz, line 61: I mean I feel like since it was linux specific, it's not a bad idea to have `linux` in the hook itself. Not saying you have to change that back, but wondering why bother changing this to begin with

README.md
Theodlz: Yeah I think it doesn't need to be mentioned here for docker. Personally, I've just been running a binary (if I need it) using an existing container. Never had to start a new one just to run a binary the existing containers already have.

Theodlz: not saying its a bad idea! We do have stuff like migrations, cross-match reruns, data import, ... we do need to run in production. So I'm also ok leaving that in there, but it might be good to mention some use cases as an example. What do you think?

Theodlz (approved):
LGTM! Just left some minor comments

README.md
antoine-le-calloch: Because the version before wasn't working at all. And the link text still have Linux in it.

Theodlz: ah shoot! That's odd, it did work eventually... perhaps I fudged it up before my commit. Sorry!

README.md
antoine-le-calloch: Good idea.

Theodlz (approved):

PR #472, opened 2026-05-11
Add gpu to benchmark folder name

antoine-le-calloch:
- Add GPU count to logs directory name based on environment variable.
- Improve benchmark scripts, clean up code and reformat to try to fit more closely the boom-apptainer implementation.

### DIFF
--- a/tests/throughput/_run.sh
+++ b/tests/throughput/_run.sh
@@ -264,7 +264,7 @@ echo "$(current_datetime) - All $EXPECTED_ALERTS alerts filtered in $FILTERING_T
 
 echo "$(current_datetime) - All alerts ingested, classified, and filtered"
 echo "$(current_datetime) - Reading from Kafka output topic"
-python "$TESTS_DIR/throughput/read-kafka-output.py"
+uv run "$TESTS_DIR/throughput/read-kafka-output.py"
 
 # -----------------------------
 # Export MongoDB collection stats to JSON for analysis
--- a/tests/throughput/_run.sh
+++ b/tests/throughput/_run.sh
@@ -22,17 +22,8 @@ if [ -z "${BOOM_REPO_ROOT:-}" ]; then
     exit 1
 fi
 
-# Ports
-MONGO_PORT=$BENCHMARK_MONGO_PORT
-REDIS_PORT=$BENCHMARK_REDIS_PORT
-KAFKA_PORT=$BENCHMARK_KAFKA_PORT
-
 # Paths
 TESTS_DIR="$TESTS_DIR"
-PERSISTENCE_DIR="$TESTS_DIR/apptainer/persistent"
-CONFIG_FILE="$TESTS_DIR/throughput/config.yaml"
-HEALTHCHECK_DIR="$BOOM_REPO_ROOT/apptainer/scripts/healthcheck"
-BENCHMARK_SIF_DIR="$TESTS_DIR/apptainer/sif"
 BG_PIDS=()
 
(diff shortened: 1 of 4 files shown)


tests/throughput/_run.sh
```diff
@@ -190,18 +239,21 @@ END_TIME=$(date +%s)
...
-# Wait until we've filtered all alerts
+# -----------------------------
+# Wait for all filters to run on all alerts
+# -----------------------------
 echo "$(current_datetime) - Waiting for filters to run on all alerts"
 START_TIME=$(date +%s)
 PASSED_ALERTS=0
 while [ $PASSED_ALERTS -lt $EXPECTED_ALERTS ]; do
-    PASSED_ALERTS=$(docker compose "${COMPOSE_CONFIG[@]}" logs scheduler | grep "passed filter" | awk -F'/' '{sum += $NF} END {print sum}')
+    PASSED_ALERTS=$(docker compose "${COMPOSE_CONFIG[@]}" logs scheduler | grep "passed filter" | awk -F'/' '{sum += $NF} END {print sum}' || true)
```
Theodlz, line 249: concretely, what does the `|| true` do here?

tests/throughput/read-kafka-output.py
```diff
@@ -1,17 +1,24 @@
...
-# requires-python = ">=3.13"
-# dependencies = [
-#     "confluent-kafka",
-# ]
-# ///
+"""Script to read the Kafka output from the BOOM throughput test.
+    requires:
+        Python 3.13+,
+        confluent-kafka
+"""
```
Theodlz: does that new syntax still work with uv? The previous one was there to allow calling `uv run` without creating a venv ahead of time

tests/throughput/run.py
```diff
@@ -1,14 +1,10 @@
...
-# ]
-# ///
-
+"""Script to benchmark BOOM.
+    requires:
+        Python 3.13+,
+        pyyaml,
+        pandas>2,
+        astropy
+"""
```
Theodlz: same question here

tests/throughput/run.py
```diff
@@ -113,13 +121,21 @@
...
+    f"{args.boom_repo_dir}/logs",
     "boom-"
     + (
         f"na={args.n_alert_workers}-"
         f"ne={args.n_enrichment_workers}-"
-        f"nf={args.n_filter_workers}"
+        f"nf={args.n_filter_workers}-"
+        f"gpu={gpus}"
     ),
 )
```
Theodlz, line 143: might be good to update the boom paper repo to use this new syntax. not needed now, but if/when we get that repo to point at boom latest.

Theodlz (changes requested):
I left some questions there. Most importantly: it looks like changes made to the header of the python script broke the `uv run ...` setup we usually have.

tests/throughput/_run.sh
antoine-le-calloch: I don't really remember why I added this. It was a long time ago, but I imagine it was because I was having some issues with the code exiting.

tests/throughput/read-kafka-output.py
antoine-le-calloch: Ohhh okey that was the reason, now I understand.

tests/throughput/run.py
antoine-le-calloch: https://github.com/boom-astro/boom-paper/issues/26

Theodlz (approved):

PR #475, opened 2026-05-15
Restrict prod deploys to version tags and admins/maintainers

petebachant:
The repo now also has a ruleset for tag creation, so only maintainers and admins can create/update/delete tags. This gives us a bit more control over how BOOM can be deployed, so we can give others write access to the repo without prod deploy access.

## TODO

- [x] Document repo settings required

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -9,9 +9,9 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Version to deploy (tag, branch, or commit SHA)"
+        description: "Version tag to deploy (must match 'v*')"
         required: true
-        default: "main"
+        default: "v0.0.0"
         type: string
 
 jobs:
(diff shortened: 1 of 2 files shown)


Theodlz (commented):
This looks great. One question: say a non-admin user opens a PR that edits this github action, can they remove the permissions and tag requirement and force deploy, or changes to deploy.yaml don't take effect in a PR and need to be merged to main before they do?

petebachant:
> One question: say a non-admin user opens a PR that edits this github action, can they remove the permissions and tag requirement and force deploy, or changes to deploy.yaml don't take effect in a PR and need to be merged to main before they do?

I don't think so. There are rules set at the repo level that tags may only be created by maintainers, and the `production` environment can only act upon a tag like `v*`. Any changes made by non-maintainers wouldn't be able to impact the `production` environment. They could try to remove the environment key and deploy to a GitHub runner, but that wouldn't achieve much 😄

Theodlz (approved):
LGTM! This is great @petebachant. It's an elegant solution that only fits in a handful of lines.

PR #477, opened 2026-05-15
Move concurrent update error logging to debug level.

antoine-le-calloch:


Theodlz:
@antoine-le-calloch I kind of feel like this warning/error should not show up if all was working properly, meaning instead of just silencing a genuine warning maybe we should:
- run a script to fix all the existing data in the DB, so this doesn't happen anymore
- if it still does after the DB-wide patch, investigate what's happening

antoine-le-calloch:
> @antoine-le-calloch I kind of feel like this warning/error should not show up if all was working properly, meaning instead of just silencing a genuine warning maybe we should:
> 
>     * run a script to fix all the existing data in the DB, so this doesn't happen anymore
> 
>     * if it still does after the DB-wide patch, investigate what's happening

@Theodlz I think you're confusing it with the other `is not strictly increasing` error. I have also a [PR](https://github.com/boom-astro/boom/pull/479) open for this one but I'm still working on it.

Theodlz:
> > @antoine-le-calloch I kind of feel like this warning/error should not show up if all was working properly, meaning instead of just silencing a genuine warning maybe we should:
> > ```
> > * run a script to fix all the existing data in the DB, so this doesn't happen anymore
> > 
> > * if it still does after the DB-wide patch, investigate what's happening
> > ```
> 
> @Theodlz I think you're confusing it with the other `is not strictly increasing` error. I have also a [PR](https://github.com/boom-astro/boom/pull/479) open for this one but I'm still working on it.

ohhh my bad, you are right. I did mix up these 2 errors. Let me re-read this PR.

Theodlz (approved):
LGTM. That approach where we don't always log errors in the function and instead do it manually isn't my favorite, but it's fine here given the length of the function! Feel free to merge (once green)

antoine-le-calloch:
> LGTM. That approach where we don't always log errors in the function and instead do it manually isn't my favorite, but it's fine here given the length of the function! Feel free to merge (once green)

Yeah I agree but how can I move it to debug without doing this?

Theodlz:
> > LGTM. That approach where we don't always log errors in the function and instead do it manually isn't my favorite, but it's fine here given the length of the function! Feel free to merge (once green)
> 
> Yeah I agree but how can I move it to debug without doing this?

we could imagine an approach where the function that's currently raising the error (that really is just a warning) returns some Status enum instead of just Ok() (a little bit like we handle the "alert already exist" stuff, if you ever looked at that). But honestly, I don't think it's worth the extra complexity here. We can stick with what you added!

PR #479, opened 2026-05-18
[WIP] Add binary repair photometry ordering

antoine-le-calloch:
- Add the `repair_photometry_ordering` binary: scans `<survey>_alerts_aux`, flags the timeseries fields (`prv_candidates`, `prv_nondetections`, `fp_hists`) that are not strictly increasing by `jd`, and repairs them with the same aggregation expression ingestion uses, so a repaired document matches what a fresh write would produce. Supports `--survey`, `--processes`, `--batch-size` and `--dry-run`.
- Extract the range-sharding, progress-logging and task-joining helpers into `utils`: `shard_field`, `range_shards`, `merge_filters`, `join_tasks` and `TaskError`, previously duplicated inside `reprocess_crossmatch`.
- Treat a `JoinError` as a failure everywhere tasks are joined. A worker that panicked was only logged, so `reprocess_crossmatch` reported an incomplete run as a success, and its catalog-driven mode went on to commit a partial temp buffer over valid live cross-matches.
- Add `--processes` to `prepare_catalog` and run its scan in parallel shards, propagating errors instead of exiting from inside the scan loop.
- Build and ship both binaries in the CPU and GPU images.


antoine-le-calloch:
Split into two different PRs:
- Improve binaries - #631 
- Add binary that repair photometry ordering - #632

PR #482, opened 2026-05-21
changes to make GPU throughput testing work again

frenbox:
The throughput testing in the main branch was not working for GPU case due to couple of recent changes

Fixes:

- `tests/throughput/compose.yaml` — scheduler service:
  `RUST_LOG: error,boom=debug,scheduler=info` 
  
- `tests/throughput/_run.sh` — wrap `check_gpu_ready`'s pipeline in
  `( set +o pipefail; … )` so SIGPIPE-induced 141 doesn't mask grep's 0.

Both changes are local to the throughput test; nothing else is affected.

### DIFF
--- a/tests/throughput/compose.yaml
+++ b/tests/throughput/compose.yaml
@@ -155,7 +155,7 @@ services:
       - /app/scheduler
       - ztf
     environment:
-      RUST_LOG: error,boom=debug
+      RUST_LOG: error,boom=debug,scheduler=info
       BOOM_GPU__ENABLED: ${BOOM_GPU__ENABLED:-false}
       BOOM_GPU__DEVICE_IDS: ${BOOM_GPU__DEVICE_IDS:-0}
     networks:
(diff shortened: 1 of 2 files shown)


tests/throughput/compose.yaml
```diff
@@ -155,7 +155,7 @@ services:
       - /app/scheduler
       - ztf
     environment:
-      RUST_LOG: error,boom=debug
+      RUST_LOG: error,boom=debug,scheduler=info
```
Theodlz, line 158: I suppose we may have `RUST_LOG: error,boom=debug` set for other containers/services, maybe we need to set that for them too? Maybe that impacts more than just the throughput test? Would be nice to validate that quickly, but otherwise LGTM

Theodlz (approved):
LGTM

tests/throughput/compose.yaml
frenbox: It seems `RUST_LOG` is not set up on any other containers we have. We might want to revisit this in future to make our logs more concise.

PR #484, opened 2026-06-03
Feat: RTF + mTAN encoder integration: feature preparation, inference, and embedding storage

KeshavMajithia:
Adds RTF and mTAN encoder inference to the ZTF enrichment pipeline. When a ZTF alert is processed, BOOM now computes a 128D RTF embedding and a 2D mTAN embedding alongside the existing ACAI/BTSBot classifications, and stores them in the alert document under an `embeddings` field.

### Changes

**`src/enrichment/models/rtf.rs`** — `prepare_features()` constructs the (1, 257, 37) photometry tensor, (1, 257) padding mask, and (1, 3, 63, 63) CHW cutout tensor from raw alert data. The 37 channels match the training pipeline: log time deltas, logflux, band one-hots, and 30 alert metadata keys.

**`src/enrichment/models/mtan.rs`** — `prepare_features()` filters to g/r bands, merges nearby observations, normalizes magnitudes and time to [0,1], and pads to 200 steps. `pool_embedding()` mean-pools qz0_mean across 50 query times to produce the final 2D vector. Skips alerts with fewer than 3 observations.

**`src/enrichment/ztf.rs`** — Adds `ZtfAlertEmbeddings` struct, `compute_embeddings()` method, and integrates it into `process_alerts()`. Both models fail gracefully (log a warning and store `None` for that embedding).

**`k8s/08-boom-scheduler-ztf.yaml`** — Adds an initContainer that downloads all 8 ONNX model files (5 ACAI, BTSBot, RTF, mTAN) from [boom-astro/boomEncoders](https://huggingface.co/boom-astro/boomEncoders) into an emptyDir volume mounted at `/app/data/models`.

### MongoDB schema after this change

```json
{
  "classifications": { "acai_h": 0.95, ... },
  "embeddings": {
    "rtf": [0.123, -0.456, ...],   // 128 floats
    "mtan": [0.789, -0.012]        // 2 floats
  }
}

### DIFF
--- a/.github/workflows/build-fork.yaml
+++ b/.github/workflows/build-fork.yaml
@@ -2,7 +2,7 @@ name: Build and push BOOM image
 
 on:
   push:
-    branches: [feat/public-filter-endpoints]
+    branches: [feat/public-filter-endpoints, feat/rtf-mtan-encoders]
   workflow_dispatch:
 
 env:
--- a/.github/workflows/build-fork.yaml
+++ b/.github/workflows/build-fork.yaml
@@ -1,0 +1,41 @@
+name: Build and push BOOM image
+
+on:
+  push:
+    branches: [feat/public-filter-endpoints]
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: keshavmajithia/boom
+
+jobs:
+  build-and-push:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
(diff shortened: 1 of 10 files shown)


petebachant:
I'm curious about the inclusion of a Kubernetes manifest here. AFAIK, Kubernetes is not used at Caltech or UMN. So far, models have been saved in this repo using Git LFS and baked into the Docker images.

mcoughlin:
@petebachant @antoine-le-calloch I do worry about the difference in deployment methods and how we can effectively develop / add unit tests for some of these forthcoming features. k8s is how we will deploy testing instances at NRP, so we need to support that too.

petebachant:
> @petebachant @antoine-le-calloch I do worry about the difference in deployment methods and how we can effectively develop / add unit tests for some of these forthcoming features. k8s is how we will deploy testing instances at NRP, so we need to support that too.

Can we write up the use case, i.e., who will be running it, what they'll be testing, etc.? We could convert from Docker Compose to Helm, but if it's just to test ML models or something, the solution would look different.

PR #486, opened 2026-06-04
Allow running throughput test on workflow dispatch

petebachant:
We seem to be failing on a bunch of branches. I want to test that something with the runners hasn't changed, and this will allow us to update the baseline via the GitHub website if necessary.

### DIFF
--- a/.github/workflows/test-throughput.yaml
+++ b/.github/workflows/test-throughput.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches: ["main"]
   pull_request:
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


petebachant:
Our baseline seems to have shifted and now all of our PRs are failing. @thomasculino if you can approve this I can merge and update the baseline.

Theodlz:
@petebachant that notification caught my attention :)

This happens, rarely but it does happen. I suggest just rerunning the offending CI job on main, that's what I usually do.

petebachant:
> @petebachant that notification caught my attention :)
> 
> This happens, rarely but it does happen. I suggest just rerunning the offending CI job on main, that's what I usually do.

Yep, this change basically just makes that a little easier to do.

PR #488, opened 2026-06-10
Update kafka.rs to include OTel metrics endpoint

sanjeev-sridhar-28:
Added OTel metrics call to kafka.rs file to make OTel metrics accessible via GET request within the babamul-web api


sanjeev-sridhar-28:
pre-commit.ci autofix

PR #489, opened 2026-06-10
Move front end in here

petebachant:
This was added as a git subtree so we get all of the commits.

Benefits:

- The entire stack spins up with one command (`make dev`), and both the front and back end will auto-rebuild/reload on change.
- We can automate deployment all in a single go.

Resolves https://github.com/boom-astro/boom-deploy-kaboom/issues/5

## TODO

- [x] Switch Actions runner from repo to org level
- [x] Check this repo's environments, variables, and secrets
- [x] Check this repo's deployment workflow
- [ ] Archive front end repo after all PRs 
- [ ] Add end-to-end tests?


frontend/test.js
thomasculino, line 1: Do we still need to keep this file? It looks like a test to decode fits files.

frontend/test2.js
thomasculino, line 1: Same comment as `frontend/test2.js`, do we need to keep this file? It looks like a test that's useless now.

thomasculino (approved):
Looks good to me! Just minor comment about 2 files. And as you say in the PR, We should add tests but we can do that in a later PR

frontend/test2.js
petebachant: Good catch. These were actually committed to the old front end repo, but not needed anymore as far as I can tell.

PR #493, opened 2026-06-12
Fix CI for PR from Fork

antoine-le-calloch:
Throughput comments were failing (`403 Resource not accessible by integration`) on PRs opened from forks. GitHub forces the `GITHUB_TOKEN` to be read-only in that case, so the action can't post comments. The comment-posting job is now moved to a separate `workflow_run`-triggered workflow, which runs in the base-repo context and therefore gets a writable token, even for forks. The test job (which runs the PR's code) stays read-only and passes the results through an artifact—no secrets are exposed to fork code.


mcoughlin (approved):
Will be nice to have PRs be green sometimes :)

PR #494, opened 2026-06-13
WINTER

mcoughlin:
This PR adds WINTER as an instrument option.


petebachant:
We're using a combination of both WINTER and WNTR in the naming for variables, data collections, etc. Is it worth picking one for consistency?

The goal here is to allow the internal (Skyportal) account to write filters against WINTER, and not make those alerts available to external users via Babamul?

mcoughlin:
@petebachant Good points! Maybe WINTER is fine? I suspect they shortened it to make it more similar to ZTF length :) And WINTER is all internal, so no babamul stuff needed.

petebachant (approved):
LGTM. We'll probably need a follow-on PR to actually add these services to the Docker Compose config, however. Adding a new data source or survey requires a consumer and a scheduler for each--not the most scalable pattern.

mcoughlin:
@petebachant Sounds good. Can you take care of that? Also tagging @antoine-le-calloch so he knows on the UMN instance side.

petebachant:
> @petebachant Sounds good. Can you take care of that? Also tagging @antoine-le-calloch so he knows on the UMN instance side.

Sure, can do. I'll also continue on my theme recommending that the UMN/Apptainer-specific stuff be moved in here so we can keep things cohesive and do it in a single PR.

PR #520, opened 2026-06-29
Make Kafka consumers self-roll over via topic-pattern subscription

mcoughlin:
This PR enbles Kafka consumers to self-roll over via topic-pattern subscription.

@petebachant another idea for the rollover stuff.

### DIFF
--- a/src/kafka/base.rs
+++ b/src/kafka/base.rs
@@ -402,10 +402,8 @@ pub trait AlertConsumer: Sized {
     /// `^` are interpreted by librdkafka as regular expressions, so the consumer
     /// auto-discovers each new day's topic (e.g. `ztf_20260629_programid1`) and
     /// rolls over on its own without restarting. Surveys with a single static
-    /// topic (e.g. LSST) return literal topic names instead.
-    ///
-    /// Defaults to the regex form `^<survey>_<8 digits>...$`; overridden per
-    /// survey where the topic layout differs.
+    /// topic (e.g. LSST) return literal topic names instead. Each survey
+    /// implements this per its topic layout.
     fn topic_patterns(&self) -> Vec<String>;
     fn output_queue(&self) -> String;
     fn survey(&self) -> &'static str;
@@ -564,20 +562,22 @@ fn seek_to_timestamp(consumer: &BaseConsumer, timestamp_ms: i64) -> KafkaResult<
     Ok(())
 }
 
-// Position freshly-assigned partitions for the long-running consumer: those with
-// a committed offset resume from it; the rest are resolved by `timestamp_ms`
+// Position the given `targets` partitions for the long-running consumer: those
+// with a committed offset resume from it; the rest are resolved by `timestamp_ms`
 // (an old day's topic has nothing at/after it -> seek to end/skip; today's and
-// any newly-created daily topic -> its start). Committed partitions are never
-// re-sought, so this is safe to run on every (re)assignment.
-fn seek_resume_or_skip(consumer: &BaseConsumer, timestamp_ms: i64) -> KafkaResult<()> {
-    let assignment = consumer.assignment()?;
-    if assignment.count() == 0 {
+// any newly-created daily topic -> its start).
+fn position_partitions(
+    consumer: &BaseConsumer,
+    targets: &TopicPartitionList,
+    timestamp_ms: i64,
+) -> KafkaResult<()> {
+    if targets.count() == 0 {
         return Ok(());
     }
     let committed = consumer.committed(KAFKA_TIMEOUT_SECS)?;
 
     let mut to_resolve = TopicPartitionList::new();
-    for elem in assignment.elements() {
+    for elem in targets.elements() {
         match committed
             .find_partition(elem.topic(), elem.partition())
             .map(|c| c.offset())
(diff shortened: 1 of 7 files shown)


petebachant:
Haha, two PRs 2 minutes apart from each other!

petebachant:
This seems like a better approach, and avoids reconsuming messages, but I will review in more depth.

petebachant (approved):
Overall I think this is worth giving a shot. I am no Kafka expert, but I had Claude take a look and it seems like we may replay messages in some situations, along with some other minor efficiencies, but we have lots of headroom on the server, and the data pipeline is idempotent.

---

This PR lets long-running Kafka consumers auto-discover each new day's alert topic
instead of being restarted per day. It adds a `topic_patterns()` trait method (a
librdkafka `^…$` regex per survey), routes the long-running (`!exit_on_eof`) path to
subscribe to that pattern while the one-shot drain path still targets concrete topics,
switches the long-running consumer to `cooperative-sticky` + auto-commit + manual offset
store, and adds `seek_resume_or_skip` to position freshly-assigned partitions (resume
committed ones, skip old retained topics, start today's/new topics). A new end-to-end
test exercises cold-start skip, current-day consume, and mid-run rollover.

The design is sound. The reposition-on-reassignment path is where the interesting
behavior concentrates, but — see the framing note below — its failure mode is *replay*,
which downstream processing absorbs safely.

## Framing: replay is safe here, skip would not be

Downstream processing is idempotent, and confirming this changes the severity of the
findings below. The consumer pushes raw payloads to a redis queue; the alert worker then:

- inserts each alert with a unique `candid` index — a duplicate hits Mongo error `11000`
  and is mapped to `ProcessAlertStatus::Exists` rather than erroring
  (`src/alert/base.rs:1067`; same pattern for aux and cutouts);
- **only propagates on `Added`** — the classifier/output queue is pushed exclusively for
  freshly-inserted alerts (`src/alert/base.rs:1351`). On `Exists` the worker just clears
  the temp queue and does nothing else (`src/alert/base.rs:1358`).

So a replayed message costs a re-fetch, a re-push to redis, and one no-op insert attempt
(bounded by `max_in_queue` backpressure). It does **not** cause double classification,
duplicate downstream side effects, or data corruption. Replays are wasted work, not wrong
results.

The important corollary: idempotency protects against reading *too much* (replay), not
against reading *too little* (**skipping unconsumed fresh data = alert loss**). Every
finding below is a rewind / consume-from-`earliest` path, i.e. replay — none can drop a
real alert. The only place idempotency would *not* save us is if `seek_resume_or_skip`
ever seeks *forward* past unconsumed fresh messages; that is the PR's core design logic,
not one of these findings, and is worth guarding in review of that function specifically.

## Findings

### 1. `src/kafka/base.rs` (~line 806 / `seek_resume_or_skip`) — Repositioning rewinds the current partition to its last *committed* offset, replaying messages consumed since the last auto-commit on each rollover. *(efficiency / wasted work — not a correctness bug)*

`seek_resume_or_skip` re-seeks partitions with a committed offset back to that committed
offset. Auto-commit lags the in-memory read position by up to `auto.commit.interval.ms`
(~5s default), so the current fetch position is ahead of the committed offset. When a new
day's topic is discovered and the assignment count changes, `seek_resume_or_skip` runs
again over *all* assigned partitions — including the actively-consumed current-day one —
and `seek`s it backward to the committed offset.

**Behavior:** consumer is steadily processing `winter_20260629`; auto-commit last ran 4s
ago; `winter_20260630` is created and joined → `assignment_count` grows →
`seek_resume_or_skip` rewinds the day-29 partition to the committed offset → the last ~4s
of already-inserted alerts are re-read and re-pushed to redis. Downstream, each re-insert
hits `11000` → `Exists` → no propagation, so the effect is a few seconds of redundant
Kafka/redis/Mongo work, once per rollover. A sharper sub-case: if auto-commit hasn't fired
yet for the current partition (no committed offset), it falls into the `to_resolve` branch
and is sought to the cold-start `timestamp_ms`, replaying the *entire* day — still safe,
still idempotent, but the largest wasteful case. The comment "Committed partitions are
never re-sought, so this is safe to run on every (re)assignment" is inaccurate about the
mechanism (committed partitions *are* re-sought), though the *safety* conclusion happens
to hold because of downstream dedup. Cleaner fix: only reposition newly-added partitions.

### 2. `src/kafka/base.rs` (~line 803) — Reposition triggers only on assignment *count* change, so a same-count rebalance leaves a new partition unpositioned and re-consumes an old retained topic from `earliest`. *(efficiency / wasted work — not a correctness bug)*

The main loop repositions only when `count != assignment_count`. A cooperative rebalance
that swaps partitions while keeping the total count equal (e.g. an old retained topic's
partition revoked as a new day's partition is added, or partitions redistributed across
consumers in the group) never trips the check, so the freshly-assigned partition is never
run through `seek_resume_or_skip`.

**Behavior:** with `auto.offset.reset=earliest`, an unpositioned partition on an old,
still-retained topic consumes from offset 0 — re-reading stale alerts the rollover design
means to skip. These are already in the DB, so downstream dedup absorbs them; the cost is
one topic's worth of redundant re-processing on that rebalance. Note this is replay, not
loss: an unpositioned *new-day* partition would likewise start at `earliest` of today's
topic and thus consume the full day of real data — no alerts dropped. Comparing assignment
membership (topic+partition set), not just its size, would avoid the redundant work.

### 3. `src/kafka/base.rs` (~line 799) — `consumer.assignment()` is called on every message iteration of the hot poll loop. *(efficiency — valid independent of idempotency)*

The reposition check calls `consumer.assignment()` (a librdkafka lock + fresh
`TopicPartitionList` allocation) once per processed message, purely to detect an
occasional assignment change.

**Impact:** at sustained alert throughput this adds a per-message allocation and lock
acquisition to the hottest path for an event that fires at most once per day. Detecting
reassignment via a rebalance callback (or a periodic/time-throttled check) removes the
per-message cost.

### 4. `src/kafka/ztf.rs:52` — Empty `program_ids` produces a pattern that matches unintended topics. *(minor / defensive)*

If `self.program_ids` is ever empty, the join yields `^ztf_[0-9]+_programid()$`, whose
`()` matches the empty string — i.e. it matches a topic literally ending in `programid`
with no id. Low likelihood (program_ids is presumably always populated), but an empty
config would silently subscribe to the wrong topic set rather than failing loudly.

## Notes (not blocking)

- The new end-to-end test would not catch the replay behavior in #1/#2: it asserts
  expected payloads are present (via a `HashSet`) and only that `"stale"` is absent, so
  duplicate replays of fresh/newday messages pass silently. If you want the test to guard
  the *efficiency* intent (no redundant re-consume), assert exact counts per payload — but
  this is a "prevent wasted work" guard, not a correctness guard.
- LSST's `topic_patterns()` returning `topic_names(0)` (literal, no `^`) correctly opts out
  of regex subscription — good.

## Bottom line

Given confirmed downstream idempotency, #1 and #2 are efficiency/wasted-work notes, not
correctness blockers — replays cannot escape the insert dedup, and none of these findings
can drop an alert. #3 is a legitimate hot-path cleanup regardless. #4 is minor defensive
polish. With the PR marked `[DNM]`, none of this is urgent; the items worth folding in
before merge are the reposition-only-new-partitions change (subsumes #1 and #2) and the
hot-loop check in #3.

mcoughlin:
@petebachant I took a pass. I think you can merge when you are ready.

PR #531, opened 2026-07-08
Updated kafka.rs to include OTel metrics endpoint

sanjeev-sridhar-28:
Configured src/api/babamul/kafka.rs backend file to make OTel metrics endpoint accessible via HTTP GET request used to access Kafka within babamul-web api


petebachant (commented):
1. Note that this endpoint is totally public--does not require authentication. Is that intended?
2. Is there a reason to query Otel rather than Prometheus for this data? 
3. Is this for general users or admins? If it's for admins we can just add to the Grafana dashboard.

petebachant:
Oh, and if this is supposed to be shown on the front end somewhere, I'd recommend putting those changes into this PR, so it's a full vertical slice.

sanjeev-sridhar-28:
Hi Pete, 
Thank you for taking a look. Here are the answers to each of your questions, based on the information that Antoine gave me when I was assigned this project: 

1. I was intending to add this endpoint as part of a public dashboard on the babamul-web frontend that showed the number of ingested alerts in real time for the current night. As such, I didn’t think that credentials should be necessary - should I implement them?
2. I was under the impression that there was no prometheus endpoint accessible in the front end based on my own exploration of the code base, so I implemented this through OTel to make the metrics network accessible in the front end. If there is an existing prometheus endpoint that can be accessed in the front end, please let me know where to find it. 
3. Dashboard is public, which is why I am not sending updates to grafana dashboard

I was planning to send two separate PRs with this and the front end, but I will send a full slice PR when I finish debugging the dashboard. If you could help clarify some of my questions above in the meantime, I would really appreciate it. Thanks!

petebachant:
That makes sense. I didn't realize the Babamul dashboard was public, but since it is, the API endpoints need to be too.

There is no Prom endpoint available to the front end right now, but you could potentially access Prom from within your API endpoint and have access to more data compared to OTel, since those metrics are being connected in there. Just a thought--I haven't done anything similar before.

Totally fine to split front end and back end changes into separate PRs. I just find it easier to test things end-to-end all in one go, since `make dev` now spins up both the front and back end at the same time from the same repo.

PR #535, opened 2026-07-10
add xmatch catalogs to all surveys on caltech config

thomasculino:
Missing config for some surveys in caltech config

### DIFF
--- a/config/prod/caltech/config.yaml
+++ b/config/prod/caltech/config.yaml
@@ -284,6 +284,38 @@ crossmatch:
         discovery_jd: 1
         internal_names: 1
       max_results: 3
+    - catalog: DESI_DR1
+      radius: 30.0 # 30 arcseconds
+      use_distance: true
+      distance_key: "z"
+      distance_max: 30.0
+      distance_max_near: 300.0
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
+        z: 1
+        zerr: 1
+        zwarn: 1
+        chi2: 1
+        deltachi2: 1
+        spectype: 1
+        zcat_nspec: 1
+    - catalog: LS_DR10_PHOTOZ
+      radius: 30.0 # 30 arcseconds
+      use_distance: true
+      distance_key: "z_phot"
+      distance_max: 60.0
+      distance_max_near: 300.0
+      projection:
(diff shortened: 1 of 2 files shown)


petebachant:
@thomasculino the Caltech config (`config/prod/caltech/config.yaml`) is auto-generated based on the top level config (`./config.yaml`) and the override, so anything that is to be different from the base should go into `config/prod/caltech/overrides.yaml`.

thomasculino:
> @thomasculino the Caltech config (`config/prod/caltech/config.yaml`) is auto-generated based on the top level config (`./config.yaml`) and the override, so anything that is to be different from the base should go into `config/prod/caltech/overrides.yaml`.

@petebachant I see. I was suspecting it. When is the Caltech config generated exactly? Do we need to keep a file here or we can just keep the override? Also, is the override a literal copy of the top level config + overrides? Or only the difference?

petebachant:
> When is the Caltech config generated exactly? Do we need to keep a file here or we can just keep the override? Also, is the override a literal copy of the top level config + overrides? Or only the difference?

Is it actually generated by GitHub Actions, e.g., https://github.com/boom-astro/boom/pull/535/commits/670ffe32b5011a1743b69fd13814e3a66e0fdd3f. So you can edit the override, commit, push, and the full config will be auto-generated. You can also run `make configs` to see/commit the final combined version.

thomasculino:
> Is it actually generated by GitHub Actions, e.g., [670ffe3](https://github.com/boom-astro/boom/commit/670ffe32b5011a1743b69fd13814e3a66e0fdd3f). So you can edit the override, commit, push, and the full config will be auto-generated. You can also run `make configs` to see/commit the final combined version.

@petebachant So, the caltech config should have the new catalogs since they have been added to the top level config. However, it looks like it only added the catalogs for WINTER and doesn't add them for ZTF and LSST. Is that normal or is something wrong with the caltech config generation?

petebachant:
I believe the fix is probably actually to remove some of the overrides, since they don't merge together at the deeply nested levels. We could modify the config override logic to do that, but maybe we don't need it if it's okay to just leave in the base config.

mcoughlin (approved):
LGTM

PR #536, opened 2026-07-15
Wire DECam through the full pipeline

mcoughlin:
This PR finishes wiring DECam into the full pipeline (request by DECam team who are basically ready).


mcoughlin:
@thomasculino might be good for this one to land so we can properly expose forced photometry?

thomasculino:
> @thomasculino might be good for this one to land so we can properly expose forced photometry?

ok reviewing it now

src/alert/decam.rs
```diff
@@ -44,9 +44,12 @@ pub struct FpHist {
...
     pub forcediffimfluxunc: f64,
-    #[serde(rename(deserialize = "forcediffimmag"))]
+    // Stored/serialized as `magap`; `alias` also accepts the Avro packet's
+    // `forcediffimmag` on input so the same struct round-trips Avro -> Mongo ->
+    // struct. (A deserialize-only rename would break the Mongo read-back.)
+    #[serde(alias = "forcediffimmag")]
     pub magap: f64,
-    #[serde(rename(deserialize = "forcediffimmagunc"))]
+    #[serde(alias = "forcediffimmagunc")]
     pub sigmagap: f64,
```
thomasculino, line 53: In `src/enrichment/decam.rs`, in `create_decam_alert_pipeline()`, it uses `magpsf` and `sigmapsf` so we should double check that we are using the correct field names.

In that same file, there's no mention of `snr` so line 50 will be a problem. It looks like theo started editing some of that in the `decam-ml` branch.

thomasculino (commented):
We need to make sure that we are consistent in the naming and that we project correctly when creating the alert pipeline (untouched in this PR but will be a problem if untouched).

src/alert/decam.rs
mcoughlin: fixed!

thomasculino (approved):
LGTM

PR #542, opened 2026-07-21
filter validation endpoint and skip_validation flag on patch

mcoughlin:
This PR adds a filter validation endpoint and skip_validation flag on patch, to streamline SkyPortal interactions.


thomasculino (approved):
LGTM

PR #540, opened 2026-07-20
Adding milvus nrp

smaharjan-web:
## Summary

Foundation for storing CIDER fusion-model embeddings in a Milvus vector
database, plus the GPU enrichment work that produces them. Milvus is **off by
default** (`enabled: false`) — deployments not using it need no configuration.

**Milvus client** (`src/milvus/`, `proto/milvus/`, `build.rs`)
- gRPC client generated at build time from protos vendored at
  `milvus-io/milvus-proto` v2.6.20, so plain `cargo build` and the Dockerfile
  work with no submodule or system `protoc`.
- `MilvusClient::connect` dials NRP over TLS gRPC (`:50051`); auth interceptor
  sends base64 `user:password` + `dbname`.
- Collection lifecycle + schema: `object_id` PK (no auto_id, so writes replace),
  384-d `embedding`, `candid`, `jd`; HNSW/COSINE index.
- Credentials only via `BOOM_MILVUS__USERNAME` / `__PASSWORD`; validation is
  conditional on `enabled`.
- `cargo run --bin milvus_check` verifies endpoint, credentials, and database
  without touching data. Docs in `docs/milvus.md`.

**Not included:** writing embeddings and similarity search — separate PRs.

**CIDER fusion model + GPU enrichment**
- `CiderFusionModel` emits the 384-float L2-normalized `fusion_embedding` now on
  `ZtfAlert` — the vectors Milvus will hold. Models share one CUDA stream.
- `workers.<survey>.enrichment.batch_size` (250) is both the Valkey RPOP cap and
  the fixed ONNX batch dimension; partial batches are zero-padded so ORT reuses
  one memory plan. ~12 GB VRAM, ~63s per benchmark run on a 16 GB card.
- New `gpu` feature for villar-pso: CUDA on Linux, Metal on macOS.

**Incidental fixes**
- `kafka/base.rs`: poll until topic deletion propagates to broker metadata.
- `utils/lightcurves.rs`: per-band properties (peak/max mag, days to/since
  peak, photometry counts) with tests.
- `load_raw_config`: `.ignore_empty(true)` so empty env vars from Compose don't
  clobber `config.yaml` defaults.

## Deploying
`BOOM_MILVUS__ENABLED`/`__USERNAME`/`__DATABASE` as repo variables,
`__PASSWORD` as a secret; injected into `api` and `scheduler-ztf`. See
`docs/milvus.md`.

PR #543, opened 2026-07-22
Sending milvus embeddings

smaharjan-web:
## Milvus vector database for CIDER embeddings

The CIDER fusion model already produces an embedding for every alert — a list of numbers that captures what the object looks like. Until now we didn't utilize it for anything in the pipeline. This PR stores it in a vector database (Milvus, hosted by NRP) so we can later ask "which objects look like this one?"

**Off by default.** Nothing changes for deployments that don't turn it on.

### How it works

- The enrichment worker sends embeddings to Milvus in batches, right after ML
  classification.
- We store **one embedding per object, not per alert.** The object ID is the key,
  so when an object is seen again its embedding is overwritten with the newer one.
  We also keep the candid and date so we know which alert the stored embedding
  came from.
- The embedding is never stored in Mongo. If Milvus is on, it goes to Milvus; if
  Milvus is off, it's simply dropped. The CIDER class probabilities keep going to
  Mongo as before, either way.
- Failures don't lose alerts. If Milvus is turned on but unreachable, the worker
  refuses to start so the problem is obvious. If one batch of embeddings fails to
  send, it's logged and processing continues — the alerts themselves are already
  enriched and saved.
- Setup is a one-time manual step, not something workers do on the fly. A small
  helper command checks the connection and creates the collection; workers only
  ever connect and write, since many run at once and shouldn't race to create it.

### Turning it on

Credentials come from NRP and are passed in as environment variables — a local
`.env` for development, GitHub secrets for the deployed server. One switch
(`milvus.enabled`) controls the whole thing.

### DIFF
--- a/src/enrichment/ztf.rs
+++ b/src/enrichment/ztf.rs
@@ -422,6 +422,23 @@ pub struct ZtfAlertClassifications {
     pub fusion_embedding: Option<Vec<f32>>,
 }
 
+/// The classifications to persist in Mongo. When `keep_embedding` is false the
+/// 384-float `fusion_embedding` is dropped (it lives only in Milvus); the class
+/// probabilities are always kept either way.
+fn classifications_for_mongo(
+    cls: &ZtfAlertClassifications,
+    keep_embedding: bool,
+) -> ZtfAlertClassifications {
+    if keep_embedding {
+        cls.clone()
+    } else {
+        ZtfAlertClassifications {
+            fusion_embedding: None,
+            ..cls.clone()
+        }
+    }
+}
+
 /// Per-alert intermediate data used during enrichment processing.
 struct AlertWork {
     candid: i64,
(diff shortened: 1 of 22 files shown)


src/enrichment/ztf.rs
```diff
@@ -503,6 +507,16 @@ impl EnrichmentWorker for ZtfEnrichmentWorker {
             .enrichment
             .batch_size;
 
+        // Connect to Milvus only when the integration is switched on. The
+        // collection itself must already be provisioned (via `milvus_check
+        // --create-collection`); the worker never creates it, so that the many
+        // enrichment workers don't race to create the same collection.
+        let milvus = if config.milvus.enabled {
+            Some(MilvusClient::connect(&config.milvus).await?)
```
frenbox, line 515: This sets up a precedent  that if we can't connect to Milvus, the enrichment workers will be stalled forever. Incase Milvus goes down, do we setup a timeout or something and stop any Milvus uploads?

src/enrichment/ztf.rs
```diff
@@ -636,6 +659,17 @@ impl EnrichmentWorker for ZtfEnrichmentWorker {
             updates.push(update);
             processed_alerts.push(format!("{},{}", item.programid, item.candid));
 
+            if self.milvus.is_some() {
+                if let Some(embedding) = fusion_embedding {
+                    embedding_rows.push(EmbeddingRow {
```
frenbox, line 664: In a scenario where two alerts of a same object appears and is processed within a same batch, this push to Milvus will not inherently take the alert with the latest `jd`. You may want to `dedup` by object id and choose the largest `jd` before pushing to milvus.

frenbox (commented):
Changes look good. Please address the comments.

Also, all these could be merged in to one big PR if needed. The PR as is doesn't compile for me in my machine.

src/enrichment/ztf.rs
smaharjan-web: Fixed this on the main PR https://github.com/boom-astro/boom/pull/540

src/enrichment/ztf.rs
smaharjan-web: Fixed this on the main PR #540

PR #551, opened 2026-07-29
Add prepare_catalog binary for turning raw collections into catalogs.

antoine-le-calloch:


src/bin/prepare_catalog.rs
```diff
@@ -0,0 +1,316 @@
...
+use tracing_subscriber::FmtSubscriber;
+
+/// Binary for turning a raw collection into a catalog boom can crossmatch against.
+///
+/// It is meant to run right after importing a file (CSV, JSON, ...) into MongoDB
+/// by hand, e.g. with Compass' `Add Data -> Import File`. Such an import only
+/// gives flat columns, while crossmatch (and the `cone_search` endpoint) needs
+/// the same spatial fields the alert pipeline writes:
+///
+/// - `ra` / `dec` as numbers (Compass imports them as strings unless the type
```
frenbox: Remove the "Compass imports ... "

src/bin/prepare_catalog.rs
```diff
@@ -0,0 +1,316 @@
...
+///   is set column by column), in degrees
+/// - `coordinates.radec_geojson`, a GeoJSON point with longitude shifted to
```
frenbox, line 29: No need to define what GeoJSON is here

src/bin/prepare_catalog.rs
antoine-le-calloch: why ?

src/bin/prepare_catalog.rs
antoine-le-calloch: Why ?

src/bin/prepare_catalog.rs
```diff
@@ -0,0 +1,316 @@
...
+///   `ra - 180` so it fits the `[-180, 180]` range Mongo requires
+/// - `coordinates.l` / `coordinates.b`, the galactic coordinates
+/// - a `2dsphere` index on `coordinates.radec_geojson`
+///
+/// Documents whose ra/dec are missing or out of range are left untouched and
```
frenbox, line 34: There should basically be no documents in catalog whose ra/dec is missing.

src/bin/prepare_catalog.rs
```diff
@@ -0,0 +1,316 @@
...
+    force: bool,
+
+    /// Report what would change without writing anything.
+    #[arg(long, default_value_t = false)]
+    dry_run: bool,
+}
+
+/// Numeric value of a field, also accepting the string form a Compass import
+/// produces when the column type is left untyped.
+fn as_f64(value: Option<&Bson>) -> Option<f64> {
```
frenbox, line 80: With mongodb crate `3.6.0`, Bson does have an `as_f64()` method, which handles the numeric variants `(Double, Int32, Int64)`. Is it necessary to create a case for String?

frenbox (approved):
LGTM!

src/bin/prepare_catalog.rs
antoine-le-calloch: Right now, we process everything, skip the rows with issues, and then print them at the end. So the user can decide what to do with them. I think that’s the best behavior.

src/bin/prepare_catalog.rs
antoine-le-calloch: `Bson::as_f64()` only matches `Bson::Double`, it returns `None` for `Int32`/`Int64`, so it would silently drop any row whose dec is a whole number (Compass types an all-integer column as `Int32`). And the `String` arm is the point of the binary: Compass' CSV import leaves every column as a string unless you set the type per column, which is exactly what we're fixing here.

PR #558, opened 2026-08-05
Add --date-mode to consume a single night, and fix consumer fencing

antoine-le-calloch:
Reworks how the Kafka consumer picks its starting point, and fixes three ways a
long-running consumer could stall or drop out of its group.

The date is no longer a positional argument, so every existing invocation changes.

- Replace the positional `DATE` with two mutually exclusive options: `--from DATE`
  treats it as a starting point and catches up every night since, rolling onto each
  new nightly topic; `--on DATE` replays that date alone, without committing offsets,
  so the night stays replayable
- Add `--exit-on-eof`, which only decides whether an `--on` run stops once its topics
  are drained. It is accepted only alongside `--on`, replacing the old
  `deployment_env == "dev"` gate with a parse-time constraint
- Back off 100ms rather than a full second per missing topic while waiting for the
  initial assignment
- Catching up positions partitions at the requested date while the subscription window
  ends at today, so past nights are read instead of skipped as stale
- Run replays in their own per-date consumer group, so replaying a night no longer
  joins the production group and rebalances the long-running consumers
- Pause consumption instead of sleeping when the output queue is full, so a long
  backlog no longer stops the poll loop and gets the consumer fenced out of its group
- Warn and retry when positioning partitions fails, instead of killing a long-running
  consumer on a transient error during a rebalance

On a survey with a single static topic (LSST), the date is a seek offset rather than a
topic selector, so `--on DATE` replays from that instant to the head of the stream
instead of one bounded night.

### DIFF
--- a/src/kafka/base.rs
+++ b/src/kafka/base.rs
@@ -393,17 +393,12 @@ pub enum ConsumerError {
     ConfigError(#[from] config::ConfigError),
 }
 
-/// UTC dates a date-partitioned survey should currently be subscribed to,
-/// oldest first: the day containing `timestamp` plus the preceding
-/// `window_days`.
+/// UTC dates a date-partitioned survey should currently be subscribed to, oldest
+/// first: the day containing `timestamp` plus the preceding `window_days`.
 ///
-/// The default of 1 keeps yesterday alongside today, because a survey night
-/// straddles UTC midnight and the tail of it would otherwise be dropped at
-/// rollover. Going wider is a deliberate, temporary act: upstream expires a
-/// topic's partitions while still advertising its name, so every extra day
-/// risks re-adding partitions that fail every poll (survivable — see
-/// `is_expired_partition` — but not free). Its intended use is catching up
-/// after an upstream outage, bounded by upstream retention.
+/// Default 1 keeps yesterday, since a night straddles UTC midnight. Widening is
+/// temporary: upstream advertises names whose partitions it has expired, so each
+/// extra day risks partitions that fail every poll.
 pub fn subscription_window(timestamp: i64, window_days: u64) -> Vec<chrono::NaiveDate> {
     let today = chrono::DateTime::from_timestamp(timestamp, 0)
         .map(|dt| dt.date_naive())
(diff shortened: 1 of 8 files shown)


src/kafka/base.rs
```diff
@@ -732,20 +747,16 @@ pub async fn consumer(
...
-        // resume in place instead of replaying. We deliberately keep the default
-        // (eager) assignment strategy: switching a live consumer group to
-        // cooperative-sticky is an incompatible rebalance protocol, and members
-        // already in the group on the eager protocol reject the join with
-        // InconsistentGroupProtocol.
+        // Commit so restarts resume in place. The default (eager) assignor is
+        // deliberate: cooperative-sticky would be rejected by members already in
+        // the group with InconsistentGroupProtocol.
         client_config
             .set("enable.auto.commit", "true")
```
mcoughlin, line 758: Claude suggested a note here:

// one-shot never stores offsets (see enable.auto.offset.store + the !exit_on_eof store_offset guards), so it commits nothing even with auto-commit on by default

mcoughlin (approved):
LGTM

mcoughlin (approved):
LGTM

tests/throughput/compose.yaml
```diff
@@ -130,7 +130,9 @@ services:
     networks:
       - boom
     # have it sleep 5 seconds to allow kafka to finish producing
-    command: ["/bin/sh", "-c", "sleep 5 && /app/kafka_consumer ztf 20250311 --programids public"]
+    # --date-mode pinned: the benchmark replays one archived night, so the
+    # consumer must stay on it instead of catching up to today.
+    command: ["/bin/sh", "-c", "sleep 5 && /app/kafka_consumer ztf 20250311 --programids public --date-mode pinned"]
```
petebachant: Should we move to make the date itself an option instead of a positional argument? How does this change when we consume a survey that has a single topic, e.g., LSST?

```suggestion
    command: ["/bin/sh", "-c", "sleep 5 && /app/kafka_consumer ztf --programids public --date 20250311 --date-mode pinned"]
```

antoine-le-calloch: Done, the date is an option now: `--from DATE` catches up and rolls onto each new night, `--on DATE` replays that date alone in its own per-date consumer group with no committed offsets, and `--exit-on-eof` is only accepted alongside `--on`. `--date-mode` is gone, so this line is now `kafka_consumer ztf --on 20250311 --programids public`.

For LSST the date stops selecting a topic and becomes a seek offset into the single static one, so `--on DATE` replays from that instant to the head of the stream instead of one bounded night; the isolation half of `--on` (own group, no commits, no rollover) behaves identically on both.

PR #562, opened 2026-08-07
Fix API panicking on every request due to cloned HttpRequest

thomasculino:
request_metrics_middleware cloned req.request() and held it alive
across next.call(req).await to read the authenticated user on the
Err path. Every route lives inside a web::scope(...), and actix's
Scope router calls HttpRequest::match_info_mut() while routing,
which panics if another clone of the same HttpRequest is alive
(Rc::get_mut().unwrap() in actix-web's request.rs). That made every
request -- including the "/" healthcheck -- panic its worker,
leaving the production API container permanently unhealthy.

Only borrow the request from the ServiceResponse after next.call()
resolves (Ok branch), once routing has finished and no conflicting
clone exists; fall back to the raw path and anonymous attribution
on Err. Adds a regression test that reproduces the panic against
the prior implementation.

### DIFF
--- a/src/api/observability.rs
+++ b/src/api/observability.rs
@@ -39,10 +39,15 @@ pub async fn request_metrics_middleware(
         .app_data::<web::Data<AnalyticsClient>>()
         .map(|client| client.as_ref().clone());
     let client_info = is_babamul.then(|| ClientInfo::from_request(&req));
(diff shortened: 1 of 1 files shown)


petebachant (approved):
Thanks for catching this!

PR #564, opened 2026-08-07
repair tempo healthcheck by invoking busybox sh directly

thomasculino:
This pull request makes a minor update to the `docker-compose.yaml` file, improving the healthcheck command for one of the services by switching from `CMD-SHELL` to `CMD` and explicitly using `/busybox/sh`. This change ensures the healthcheck runs to avoid having an unhealthy container forever.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -716,7 +716,7 @@ services:
     networks:
       - boom
     healthcheck:
-      test: ["CMD-SHELL", "wget -qO- http://127.0.0.1:3200/ready | grep -q 'ready' || exit 1"]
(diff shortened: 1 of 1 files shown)


petebachant:
Do you have a description or some logs from the error? Is the Tempo container currently unhealthy?

thomasculino:
> Do you have a description or some logs from the error? Is the Tempo container currently unhealthy?

The tempo container shows unhealthy on the machine only because the health check can never be done properly. Loki and Tempo are in the same situation, they don't have the tool to run the health check inside its container. However, Loki had that until v3.6 and it's pinned to v3.5.5 so it works. Because of that, the Tempo container always shows unhealthy even when it's not. It's not a new issue, it's an issue that was always there but we never saw it.

petebachant (approved):

PR #566, opened 2026-08-08
DLR based host association

mcoughlin:
This PR makes host-based galaxy associations possible.

## Rollout order

`host_galaxy` is computed at aux creation only, so existing records need two passes, in this order:

```
reprocess_crossmatch --survey ztf --catalogs NED,LSDR10
backfill_host_galaxy --survey ztf
```

The first refreshes `cross_matches` under the current crossmatch config, the second reads those
matches and writes `host_galaxy`. Running the backfill first writes an empty association on every
record whose cross-matches predate the galaxy catalogs. `backfill_host_galaxy` warns when it samples
records carrying neither `NED.Diam` nor `LSDR10`, which is the signal that the first pass is missing.

Use `--skip-existing` to resume an interrupted backfill, and leave it off to rescore everything after
a change to the association parameters. `host_galaxy.enabled` stays `false` until both passes are done.


config/prod/caltech/config.yaml
```diff
@@ -151,6 +151,27 @@ workers:
...
+# Enabling this requires the galaxy catalogs below to be ingested AND listed in
+# the survey's `crossmatch` block, so their shape columns are projected into the
+# match documents:
+#   NED_LVS  -> diam, diam_ba, diam_pa (plus dist_mpc, dist_mpc_method, z)
+#   LS_DR10  -> shape_r, shape_e1, shape_e2, type
+# NED-LVS is preferred where it has a diameter; Legacy Survey fills in the rest.
+host_galaxy:
+  enabled: false
+  ned_lvs_catalog: NED_LVS
+  ls_dr10_catalog: LS_DR10
```
frenbox: At least at UMN it is not named like this. It is currently named "LSDR10".

config/prod/caltech/config.yaml
```diff
@@ -151,6 +151,27 @@ workers:
...
+# host_galaxy configures shape-aware host-galaxy association, which scores each
+# galaxy cross-match by its directional light radius (d_DLR) instead of a fixed
+# circular aperture. The result is stored on the alert aux document under
+# `host_galaxy`, and filters cut on `host_galaxy.best_host.d_dlr`.
+#
+# Enabling this requires the galaxy catalogs below to be ingested AND listed in
+# the survey's `crossmatch` block, so their shape columns are projected into the
+# match documents:
+#   NED_LVS  -> diam, diam_ba, diam_pa (plus dist_mpc, dist_mpc_method, z)
+#   LS_DR10  -> shape_r, shape_e1, shape_e2, type
```
frenbox: What does type imply here? i think in the mongo it is mentioned as "objtype"

frenbox:
@mcoughlin Is there a chance that the current way we do "hosted" and "hostless" will give different answers compared to this method?

Current hosted definition 

```
hosted:  lspsc_matches exists AND at least one match with match.score ≤ 0.5
```

frenbox:
The current throughput result doesn't essentially capture the baseline time increase due to this PR. Currently the changes of this PR is `disabled`. 

Once Caltech get the LSDR10 catalog and it will be good to know the latency increase due to an additional step per alert.

However, this can be tested locally on UMN if needed. @mcoughlin Please report these numbers as well. It would be great to know the latency increase.

mcoughlin:
@thomasculino @petebachant  @frenbox Will help to ingest some catalogs soonish to continue development. I can also help load some in although maybe a bit unclear on what is working in boom-catalogs and what is not admittedly :)

frenbox:
@mcoughlin 
https://github.com/boom-astro/boom-catalogs/pull/6

I have something ready here as a starter. I did this a while ago at UMN for possibly helping the prost stuff. Might need to refresh this and a PR review.

config/prod/caltech/config.yaml
mcoughlin: fixed

config/prod/caltech/config.yaml
mcoughlin: currently aligning with CIT

mcoughlin:
@frenbox +1.4 ms/alert for LS, +0.3 ms/alert for NED. I will measure this once the full LS is ingested with the Sersic parameters.

mcoughlin:
@frenbox for the hosted vs hostless, we should expect DLR to reclassify some of the current hostless to hosted where for resolved nearby galaxies the transient is in the disk but far from the nucleus. Similarly, we would expect to reclassify some of the current hosted to "not that host," where a nearby galaxy-like source is a chance alignment.

antoine-le-calloch (commented):
- UMN's NED has none of the NED-LVS columns. No Diam, Diam_ba, Diam_pa,
  Diam_survey, DistMpc, DistMpc_method, m_Ks. Caltech has them all (81% with
  Diam > 0). galaxy_from_ned_lvs returns None for every UMN row.
- UMN config has no NED_LVS crossmatch entry for ztf or lsst, only winter.
  Caltech has all three.
- star_type_value is the wrong key in the three configs. The field is
  star_type_values. Silently ignored.
- objname is always null for Legacy hosts. catalog.rs:198 reads a string,
  LSDR10 _id is a Long.
- size_is_isophotal and orientation_is_nominal are never read. Not in the
  scoring, not in StoredHostCandidate. Diam_survey is projected for nothing, and
  is missing from the Caltech ztf/lsst entries anyway.
- n_candidates_after_dlr_cut caps at 10. association.rs:87 counts after
  take(max_candidates), but the doc says "surviving the d_DLR cut".
- Three definitions of hosted. babamul.rs:302 (LSPSC score <= 0.5),
  babamul.rs:404 (sgscore1/2/3), and now ZtfAlertProperties. All four surveys
  write aux.host_galaxy but only ZTF gets a hosted property.
- The anti-drift test only reads config.yaml, never the prod configs, and
  NED_LVS_REQUIRED_KEYS lists neither Diam_survey nor m_Ks.
- Stale comment at catalog.rs:704: SER rows do carry a Sersic index now.

mcoughlin:
@antoine-le-calloch please take another look. @frenbox was finishing some of the catalog updates at UMN so those are not far off from being available. As for there being multiple ways of defining hosted, I think that is ok for now, at least until we see what works best.

frenbox:
NED LVS fields are now online at UMN.

frenbox:
@mcoughlin  The configs point towards NED-LVS. I think both caltech and UMN have NED with NED-LVS columns but the catalog is named NED.  Easy fix for claude (point it towards UMN DB if needed)

mcoughlin:
@antoine-le-calloch @frenbox I checked on UMN, and everything in the PR is currently up-to-date.

antoine-le-calloch (commented):
- Filters cannot read `host_galaxy`: absent from `aux_add_fields` and
  `insert_aux_pipeline` in `filter/ztf.rs`, and from `ZtfAlertToFilter`. The config
  comment promising `host_galaxy.best_host.d_dlr` is not reachable.
- `angular_size_index_name` derives only from `size_key` while the
  `partialFilterExpression` derives from `radius` and `angular_size_scale`. Changing
  either gives an `IndexOptionsConflict` and `scheduler.rs` `.expect()`s on it. The index
  is load-bearing: 43 ms per alert without it on UMN, 3.3 ms with it at Caltech, 0.8 ms
  for a plain cone.
- Nothing backfills `host_galaxy`. It is computed only at aux creation, never in
  `AlertAuxForUpdate`, so `enabled: true` will only cover new objects.
- `use_redshift` is inert: `Transient::new` leaves `redshift: None` and `with_redshift` is
  only called from tests.
- `rejected_as_marginal_rex` skips the S/N cut when `flux_r <= 0`, which is 319 of the
  6160 accepted REX rows on Caltech LSDR10.

antoine-le-calloch:
And also @mcoughlin the `NED` entry matches on a flat 300 arcsec floor instead of a distance-scaled radius: 0.02 to 0.79 matches per alert on 300 random ZTF alerts at Caltech, so the 38 existing filters testing `cross_matches.NED.0` go from firing on 1 alert in 75 to 1 in 3.

We will need to update their filter maybe ?

antoine-le-calloch (approved):
LGTM

PR #567, opened 2026-08-09
Subscribe Kafka consumers to a bounded daily-topic window

mcoughlin:
This PR subscribes kafka consumers to a bounded daily-topic window. 

Issue: The topic pattern matched every past night, accumulating expired partitions until the per-error backoff starved the live topic and stalled ingestion.

### DIFF
--- a/src/kafka/base.rs
+++ b/src/kafka/base.rs
@@ -459,7 +459,7 @@ pub trait AlertConsumer: Sized + Clone + Send + Sync + 'static {
     async fn consume(
         &self,
         topics: Option<Vec<String>>,
-        timestamp: i64,
+        start: StartDate,
         kafka_config: Option<KafkaConsumerConfig>,
         n_threads: Option<usize>,
         max_in_queue: Option<usize>,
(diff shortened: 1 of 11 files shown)


petebachant:
Will this cover the issue happening in #569? That is, once those servers are reachable again, will we look back the ~5 days to consume those topics too?

mcoughlin:
@petebachant Looks like IPAC didn't create the topics for last night, which someone should tell them about. I will make the window configurable to make the backfill easier.

src/kafka/base.rs
```diff
@@ -860,6 +923,35 @@ pub async fn consumer(
...
+    // whichever day the process happened to boot on.
+    let mut subscribed_date = chrono::DateTime::from_timestamp(timestamp, 0)
+        .map(|dt| dt.date_naive())
+        .unwrap_or_else(|| chrono::Utc::now().date_naive());
+    let mut position_timestamp = timestamp;
+    // Only a consumer that started on the live date should chase the clock. A
+    // run pinned to an explicit past date — replay, backfill, the throughput
+    // benchmark — must stay on the date it was asked for; rolling it forward
+    // would unsubscribe it from the very topic it was started to read.
+    let follow_current_date = tracks_current_date(subscribed_date, chrono::Utc::now().date_naive());
```
petebachant: Couldn't this be signaled directly from the CLI rather than inferred?

petebachant (approved):
One minor nit. Feel free to dismiss since it seems to be working as-is.

src/kafka/base.rs
mcoughlin: Yes added.

PR #497, opened 2026-06-23
Ingest ssSource designation data from the LSST alert stream

thomasculino:
This pull request adds support for Rubin/LSST Solar System object (SSO) metadata in alert handling, specifically by introducing the `SsSource` struct to represent per-detection SSO ephemeris and designation information. The changes ensure that SSO data is parsed, stored, indexed, and exposed throughout the alert ingestion, enrichment, and filtering pipelines, and are accompanied by new tests for correct deserialization and handling of SSO fields.

**Support for Solar System Object metadata:**

- Added the new `SsSource` struct to represent Solar System object ephemeris and designation data, with full parsing and (de)serialization support in `src/alert/lsst.rs`. This struct is now included in the `LsstRawAvroAlert`, `LsstAlert`, and `LsstObject` types, and is handled during alert ingestion and enrichment. 
- The SSO designation is now persisted in the auxiliary collection for each object, updated independently of cross-survey matches, and indexed for efficient lookup by designation. 

**Pipeline and API changes:**

- The alert enrichment and filtering pipelines have been updated to propagate the new `ss_source` field, making SSO metadata available for downstream consumers and APIs. 

**Testing and validation:**

- Added tests to verify correct deserialization of the new `SsSource` struct, including edge cases where the field is absent. Existing tests and fixtures were updated to include or expect the new fields where appropriate. 

These changes collectively enable robust handling and querying of Solar System object information in LSST alert data, paving the way for more advanced moving object science use cases.

The idea behind this PR is to query by `ssObjectId` in the future

PR #573, opened 2026-08-13
ZTF solar system objects

mcoughlin:
This PR better handles ZTF solar system objects (similar to https://github.com/boom-astro/boom/pull/497).

### DIFF
--- a/src/enrichment/ztf.rs
+++ b/src/enrichment/ztf.rs
@@ -360,7 +360,10 @@ pub struct ZtfAlertForEnrichment {
 /// it, never on `objectId`. For the same reason the alert's `prv_candidates` and
 /// `fp_hists` describe whatever else has occupied that sky position, not this
 /// object.
-#[derive(Debug, Clone, serde::Deserialize, serde::Serialize, AvroSchema, utoipa::ToSchema)]
+#[derive(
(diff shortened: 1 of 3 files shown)


src/enrichment/ztf.rs
```diff
@@ -351,15 +351,75 @@ pub struct ZtfAlertForEnrichment {
...
+/// object.
+#[derive(Debug, Clone, serde::Deserialize, serde::Serialize, AvroSchema, utoipa::ToSchema)]
+pub struct ZtfSsoAssociation {
+    /// Whether a known solar system object was identified at this position.
+    ///
+    /// Deliberately not thresholded on separation, unlike the deprecated `rock`
+    /// flag: when the upstream ephemeris degrades, that shows up as a growing
+    /// `separation_arcsec` the consumer can see, rather than silently flipping a
+    /// boolean they cannot.
+    pub is_sso: bool,
```
thomasculino, line 374: This not being optional will be a problem if we try to get an object from the db using the API.

thomasculino (commented):
Just one comment on how the `is_sso` field will be handled by the API.

src/enrichment/ztf.rs
mcoughlin: Fixed!

thomasculino (approved):

PR #574, opened 2026-08-14
rollover check on new daily topic

mcoughlin:
This PR re-subscribes if the UTC date has rolled onto a new daily topic.


petebachant (approved):
LGTM

PR #576, opened 2026-08-14
SSO filters

mcoughlin:
This PR enables filters on SSOs.

### DIFF
--- a/src/filter/ztf.rs
+++ b/src/filter/ztf.rs
@@ -427,7 +427,6 @@ fn sso_history_lookup(ztf_permissions: &Vec<i32>, window_days: f64) -> Document 
             },
             "pipeline": [
                 doc! { "$match": { "$expr": { "$and": [
-                    { "$ne": ["$$desig", Bson::Null] },
                     { "$eq": ["$candidate.ssnamenr", "$$desig"] },
                     { "$gte": ["$candidate.jd", { "$subtract": ["$$jd", window_days] }] },
                     { "$lte": ["$candidate.jd", "$$jd"] },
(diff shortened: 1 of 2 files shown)


src/utils/db.rs
```diff
@@ -125,6 +125,25 @@ pub async fn initialize_survey_indexes(
     // if survey is LSST, create an index on the ssObjectId field of the alerts collection,
     // and on the designation field of the aux collection (used to look up a moving object by
     // its MPC designation, independent of any cross-survey position match)
```
thomasculino: Let's move the comment below the update, right above the index creation for LSST.

src/filter/ztf.rs
```diff
@@ -407,6 +407,50 @@ pub async fn build_ztf_alerts(
...
+///
+/// `objectId` cannot be used: it is positional, so a moving object is given a new
+/// one on nearly every detection and its aux document describes the sky position
+/// rather than the object.
+///
+/// The outer key is the normalised `properties.sso.designation`; the inner key is
+/// the raw `candidate.ssnamenr` it is derived from. Same value, but the raw field
+/// is present on the whole archive while the normalised one only exists on alerts
+/// enriched since it was introduced.
+fn sso_history_lookup(ztf_permissions: &Vec<i32>, window_days: f64) -> Document {
```
thomasculino, line 420: properties.sso.designation can be missing (Old alerts or no mpc match). We should add a match stage before the lookup.

src/utils/db.rs
mcoughlin: Fixed.

src/filter/ztf.rs
mcoughlin: Fixed.

thomasculino (approved):
LGTM

PR #578, opened 2026-08-15
Expose LSST SSO ephemeris and cross-survey aperture excess

mcoughlin:
This PR exposes LSST SSO ephemeris and cross-survey aperture excess.

### DIFF
--- a/src/enrichment/lsst.rs
+++ b/src/enrichment/lsst.rs
@@ -221,8 +221,14 @@ pub struct LsstSsoAssociation {
     /// proportion to how fast the object moves.
     pub sky_motion: Option<f32>,
     /// Sun-object-observer angle (`ssSource.phaseAngle`), which sets how much of the
-    /// brightness is geometry rather than activity.
+    /// brightness is geometry rather than activity. Degrees.
     pub phase_angle: Option<f32>,
+    /// Sun-to-object distance (`ssSource.helioRange`), au.
+    pub helio_dist: Option<f32>,
+    /// Observer-to-object distance (`ssSource.topoRange`), au. With `helio_dist`
+    /// and `phase_angle` this is the geometry needed to reduce an apparent
+    /// magnitude to an absolute one.
+    pub topo_dist: Option<f32>,
     /// Who made the association.
     pub source: Option<String>,
 }
@@ -243,6 +249,8 @@ impl LsstSsoAssociation {
             predicted_mag: ss_source.and_then(|s| s.eph_vmag),
             sky_motion: ss_source.and_then(|s| s.eph_rate),
             phase_angle: ss_source.and_then(|s| s.phase_angle),
+            helio_dist: ss_source.and_then(|s| s.helio_range),
+            topo_dist: ss_source.and_then(|s| s.topo_range),
             source: Some("rubin".to_string()),
         }
     }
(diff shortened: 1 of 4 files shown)


src/enrichment/lsst.rs
```diff
@@ -194,15 +194,73 @@ pub struct LsstAlertForEnrichment {
...
+}
+
+impl LsstSsoAssociation {
+    /// Build from Rubin's own association. `ss_object_id` decides `is_sso` so this
+    /// agrees with `rock`; `ss_source` may be absent even when it is set.
+    pub fn from_rubin(ss_object_id: Option<&str>, ss_source: Option<&SsSource>) -> Self {
+        let is_sso = ss_object_id.is_some();
+        LsstSsoAssociation {
+            is_sso,
+            designation: ss_source.and_then(|s| s.designation.clone()),
```
thomasculino: Are we confident we will never see a case where ss_object_id is not there but ss_source is? If we want to be extra careful we can make sure is_sso is true before adding the rest.

thomasculino (commented):
One edge case comment and one interrogation, do we want to keep both the `rock` and `sso.is_sso `properties or we are planning on keeping only one in the long term?

src/enrichment/lsst.rs
mcoughlin: Fixed.

mcoughlin:
@thomasculino I prefer just moving to sso.is_sso. We can migrate the filters when ready.

src/utils/lightcurves.rs
```diff
@@ -111,6 +111,58 @@ pub struct BandProperties {
...
+            (Some(psf), Some(ap)) if psf.is_finite() && ap.is_finite() => Some(psf - ap),
+            _ => None,
+        };
+        Self::from_excess(aperture_excess)
+    }
+
+    /// LSST: fluxes rather than magnitudes. Non-positive fluxes are not convertible.
+    pub fn from_fluxes(psf_flux: Option<f32>, ap_flux: Option<f32>) -> Self {
+        let aperture_excess = match (psf_flux, ap_flux) {
+            (Some(psf), Some(ap)) if psf > 0.0 && ap > 0.0 => Some(2.5 * (ap / psf).log10()),
```
thomasculino: We should add a check for infinity here? We have one in `from_magnitudes` but not here when the result is more likely to return an infinite value.

thomasculino (commented):
Looks good, just one comment about `from_fluxes`.

src/utils/lightcurves.rs
mcoughlin: fixed

src/utils/lightcurves.rs
```diff
@@ -111,6 +111,68 @@ pub struct BandProperties {
...
+    pub fn from_magnitudes(magpsf: Option<f32>, magap: Option<f32>) -> Self {
+        let aperture_excess = match (magpsf, magap) {
+            (Some(psf), Some(ap)) if psf.is_finite() && ap.is_finite() => Some(psf - ap),
+            _ => None,
+        };
+        Self::from_excess(aperture_excess)
+    }
+
+    /// LSST: fluxes rather than magnitudes. Non-positive fluxes are not convertible.
+    pub fn from_fluxes(psf_flux: Option<f32>, ap_flux: Option<f32>) -> Self {
```
thomasculino, line 157: `from_fluxes` requires both `psf_flux`>0 and `ap_flux`>0, so it returns `None` on exactly the negative-flux epochs this struct's own doc calls "routine" in difference imaging — even though `LsstCandidate::new` (src/alert/lsst.rs:352-359) already derives a usable magpsf/magap from the same flux via .abs(); is dropping activity for those alerts intentional (data-quality gate) or should this reuse `from_magnitudes`(magpsf, magap) like ZTF does?

thomasculino (commented):
One comment from Claude

src/utils/lightcurves.rs
mcoughlin: Yup fixed.

thomasculino (approved):
LGTM

PR #579, opened 2026-08-19
enable gpu

thomasculino:
Enable the gpu on main since we now have a GPU on the machine.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -31,7 +31,7 @@ posthog:
   # consumption to users. Also drives the Grafana consumption metrics.
   consumption_interval_seconds: 300
 gpu:
-  enabled: true # Set to true to load ONNX models on GPU (CUDA) instead of CPU.
+  enabled: false # Set to true to load ONNX models on GPU (CUDA) instead of CPU.
   # Models are loaded once at startup and shared across all enrichment workers.
   # When false, models are loaded on CPU (BOOM_GPU__ENABLED env var is still respected).
   device_ids: [0] # CUDA device(s) to use. E.g. [0,1,2,3,4,5,6,7] for 8 GPUs.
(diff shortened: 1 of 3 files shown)


config.yaml
```diff
@@ -31,7 +31,7 @@ posthog:
   # consumption to users. Also drives the Grafana consumption metrics.
   consumption_interval_seconds: 300
 gpu:
-  enabled: false # Set to true to load ONNX models on GPU (CUDA) instead of CPU.
+  enabled: true # Set to true to load ONNX models on GPU (CUDA) instead of CPU.
```
petebachant: Do we want this in the top-level config, which would turn it on for all devs, or only in the Caltech override?

thomasculino: Probably in the Caltech override, you're right

thomasculino: When is the Caltech config generated again? Will the change made to the override file trigger the regeneration of the file?

petebachant: I happens automatically in GitHub Actions, and I believe it also happens if you run `make format`. That is buried in the docs somewhere so if you can think of a better place or way to express it, feel free to change.

petebachant (approved):
LGTM. As a follow-up, we should probably have GPU stats visible on Grafana so we can see how it's doing.

config.yaml
thomasculino: I was very confused because it wasn't running automatically but it's because I'm on my fork so I had to manually run the command to generate the Caltech config and commit it. I will take a look at the docs to see if we can express that more clearly

thomasculino:
> LGTM. As a follow-up, we should probably have GPU stats visible on Grafana so we can see how it's doing.

Defintely, let me open an issue for that.

PR #581, opened 2026-08-21
Parallelize reprocess_crossmatch and make interrupted runs resumable

antoine-le-calloch:
- Workers now keep --concurrency queries in flight, and the catalog-driven cleanup passes are sharded into --processes ranges.
- The temp reset, sort and swap collapse into one commit pass, skipped unless a state marker shows the previous run was interrupted.
- Adds --skip-existing, --skip-empty and --reset-temp, and queries only the radius a use_distance row actually needs.

### DIFF
--- a/src/bin/reprocess_crossmatch.rs
+++ b/src/bin/reprocess_crossmatch.rs
@@ -34,7 +34,7 @@ const ARCSEC_TO_RAD: f64 = std::f64::consts::PI / 180.0 / 3600.0;
 const STATE_COLLECTION: &str = "reprocess_crossmatch_state";
 const STATUS_MATCHING: &str = "matching";
 const STATUS_CLEAN: &str = "clean";
-const TEMP_PROBE_SAMPLE: i64 = 200;
+const TEMP_PROBE_SAMPLE: i64 = 10_000;
 
 /// Catalog-driven costs extra full passes over alerts_aux, so it only wins when the
 /// catalog is substantially smaller, not merely smaller.
@@ -272,6 +272,12 @@ async fn range_shards(
     };
 
     if bounds.len() < parts {
+        warn!(
+            "only {} sampled bounds on '{}' for {} shards, running as a single shard",
+            bounds.len(),
+            field,
+            parts
+        );
         return vec![Document::new()];
     }
     let step = bounds.len() / parts;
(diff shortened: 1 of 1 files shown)


src/bin/reprocess_crossmatch.rs
```diff
@@ -380,38 +668,64 @@ async fn process_watchlist_doc(
...
-    let empty: Vec<Document> = Vec::new();
-    aux_collection
-        .update_many(
-            existing_records.clone(),
-            doc! { "$set": { &temp_field: empty } },
+
+    // Phase 1: drop temp left behind by an interrupted run. Nothing can be left behind
+    // after a run that reached phase 3, so the scan is skipped on the normal path.
+    let state_id = format!("{}_alerts_aux:{}", survey, catalog_config.catalog);
+    if reset_temp || temp_needs_reset(&db, &aux_collection, &state_id, &temp_field).await? {
```
thomasculino, line 719: Isn't the sample in temp_needs_reset too small? It feels like we could easily miss what we're trying to catch.

src/bin/reprocess_crossmatch.rs
```diff
@@ -87,6 +122,210 @@ struct AuxIdAndCoords {
...
+            shard.clone()
+        } else {
+            doc! { "$and": [base_filter.clone(), shard.clone()] }
+        };
+        let collection = collection.clone();
+        let update = update.clone();
+        let done = Arc::clone(&done);
+        async move {
+            let result = collection.update_many(filter, update).await;
+            info!(
```
thomasculino: This logs `shard {i}/{total} done` right after each `update_many` call regardless of whether that call returned `Err`. So, an operator tailing logs during a multi-hour run sees every shard reported `done` even when one failed; the real error only surfaces later via `result?` after all shards finish, with no way to tell which shard failed from the log line. We should only display `done` when it's Ok.

src/bin/reprocess_crossmatch.rs
```diff
@@ -87,6 +122,210 @@ struct AuxIdAndCoords {
...
+            }
+        },
+        Err(e) => {
+            warn!(error = %e, "could not sample {} bounds, falling back to a single shard", field);
+            Vec::new()
+        }
+    };
+
+    if bounds.len() < parts {
+        return vec![Document::new()];
```
thomasculino, line 281: If bounds has fewer distinct/available values than the number of requested shards (parts), the function just gives up and returns vec![Document::new()], a single empty filter document, meaning "one shard covering everything," i.e. no sharding at all.

thomasculino (commented):
Would be very nice to have. A few things worth mentioning though.

thomasculino (approved):

PR #582, opened 2026-08-22
derive per-alert solar system geometry for ZTF from MPC orbits

mcoughlin:
This PR derives per-alert solar system geometry for ZTF from MPC orbits.


docker-compose.yaml
```diff
@@ -581,6 +581,45 @@ services:
...
+  mpcorb-ingest:
+    image: boom-astro/boom:latest
+    hostname: mpcorb-ingest
+    # A one-shot job: a restart policy here would re-download on every failure.
+    restart: "no"
+    build:
+      context: .
+      dockerfile: Dockerfile
+    environment:
+      BOOM_DATABASE__PASSWORD: ${BOOM_DATABASE__PASSWORD:?BOOM_DATABASE__PASSWORD must be set}
```
thomasculino, line 607: We are missing `BOOM_API__AUTH__SECRET_KEY` and `BOOM_API__AUTH__ADMIN_PASSWORD`.

src/bin/mpcorb_ingest.rs
```diff
@@ -0,0 +1,177 @@
...
+            match &staging {
+                Some(c) => {
+                    if let Err(e) = c.insert_many(std::mem::take(&mut batch)).await {
+                        error!("insert failed: {}", e);
+                        std::process::exit(1);
+                    }
+                }
+                None => batch.clear(),
+            }
+            if parsed % 200_000 == 0 {
```
thomasculino: This check is in the `batch.len() >= args.batch_size` if statement, which is true only for `batch_size` multiples. Therefore, if 200_000 is not a multiple of `batch_size`, this will never be true.

thomasculino (changes requested):

src/bin/mpcorb_ingest.rs
mcoughlin: fixed.

docker-compose.yaml
mcoughlin: fixed.

thomasculino: Did you push anything? I can't see any new commit

src/enrichment/ztf.rs
```diff
@@ -387,6 +391,18 @@ pub struct ZtfSsoAssociation {
     /// ZTF alert itself; an independent association computed by BOOM would
     /// identify itself differently here.
     pub source: Option<String>,
+    /// Sun-to-object distance at the alert epoch, au. Derived here from MPC
+    /// elements — the ZTF packet carries no state vectors — so it is `None`
+    /// whenever the object is missing from `MPC_orbits`. LSST populates the same
```
thomasculino: where is that done?

src/enrichment/ztf.rs
```diff
@@ -677,11 +779,21 @@ impl ZtfEnrichmentWorker {
         let ssmagnr = candidate.ssmagnr.unwrap_or(f32::INFINITY);
         let is_rock = ssdistnr >= 0.0 && ssdistnr < 12.0 && ssmagnr >= 0.0;
 
+        // Geometry is evaluated at the observation epoch, not at the elements'
+        // epoch, so a stale MPCORB shows up as a slowly growing error rather
+        // than as a wrong answer at a fixed date.
+        let elements = candidate
+            .ssnamenr
+            .as_deref()
+            .and_then(normalize_ztf_ssnamenr)
```
thomasculino: We normalize ztf ssnamenr on line 717 as well. Can we do that only once?

src/enrichment/ztf.rs
mcoughlin: We need to merge 578.

src/enrichment/ztf.rs
mcoughlin: fixed.

thomasculino (approved):
LGTM

thomasculino:
We need to resolve the conflicts before merging. I don't have permission to push to your branch so I couldn't do it.

PR #583, opened 2026-08-25
Adding missing docker configs for Milvus NRP pipeline

smaharjan-web:
- Copy `build.rs` and `proto/` into the Docker images so the Milvus gRPC bindings actually generate at build time
- Set `ORT_DYLIB_PATH` / `LD_LIBRARY_PATH` in the `dev` stage, so enrichment workers don't hang loading ONNX
- Fix dev compose: give `consumer-ztf` its network and service hostnames, pin Attu to amd64
- Rename the collection/table for embeddings as `boom_ztf_fusion_embeddings`
- Warn at startup if an existing Milvus collection is missing required fields that BOOM writes, instead of failing every upsert silently later

### DIFF
--- a/config/prod/caltech/config.yaml
+++ b/config/prod/caltech/config.yaml
@@ -51,7 +51,7 @@ milvus:
   database: "" # Set via BOOM_MILVUS__DATABASE environment variable
   timeout_seconds: 30
   collection:
-    name: ztf_fusion_embeddings
+    name: boom_ztf_fusion_embeddings
     # The CIDER fusion model (data/models/cider_fusion_plus_embedding.onnx)
     # emits 384 floats, L2-normalized -- so COSINE and IP are equivalent.
     dim: 384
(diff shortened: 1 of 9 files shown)


frenbox (approved):
These diff changes looks good to me :) .

config/prod/caltech/config.yaml
```diff
@@ -51,7 +51,7 @@ milvus:
   database: "" # Set via BOOM_MILVUS__DATABASE environment variable
   timeout_seconds: 30
   collection:
-    name: ztf_fusion_embeddings
+    name: boom_ztf_fusion_embeddings
```
frenbox, line 54: Might have to think how Caltech is going to deal with these changes in the Thursday meeting.

smaharjan-web: I will make changes to avoid duplicate Caltech inserts in a new PR, thank you for the reminder.

PR #585, opened 2026-08-25
use GPU when you have one and BOOM_GPU__ENABLED=true

thomasculino:
We need to apply the cuda override to use the GPU.


mcoughlin (approved):
LGTM

PR #586, opened 2026-08-25
refresh MPC_orbits automatically

mcoughlin:
This PR allows for MPC_orbits to refresh automatically.


src/utils/mpcorb.rs
```diff
@@ -204,6 +204,162 @@ pub fn parse_line(line: &str) -> Option<MpcorbEntry> {
...
+    let mut tmp = tempfile::NamedTempFile::new()?;
+    crate::utils::data::download_to_file(tmp.as_file_mut(), url, None, None, true)
+        .await
+        .map_err(|e| RefreshError::Download(e.to_string()))?;
+
+    let staging = match db {
+        Some(db) => {
+            let c = db.collection::<Document>(STAGING_COLLECTION);
+            // A previous run may have died between insert and rename.
+            let _ = c.drop().await;
```
thomasculino, line 274: MPC_orbits_staging is only dropped at the start of the next refresh_orbits() call. If this run fails after inserting some batches, the partial staging collection is left in the database until whenever the next attempt happens to run. We should drop it in the error path too, not just the next time.

src/utils/mpcorb.rs
```diff
@@ -204,6 +204,162 @@ pub fn parse_line(line: &str) -> Option<MpcorbEntry> {
...
+    db: Option<&mongodb::Database>,
+    url: &str,
+    batch_size: usize,
+    now: f64,
+) -> Result<RefreshReport, RefreshError> {
+    use std::io::{BufRead, BufReader};
+
+    tracing::info!("downloading MPCORB from {}", url);
+    let mut tmp = tempfile::NamedTempFile::new()?;
+    crate::utils::data::download_to_file(tmp.as_file_mut(), url, None, None, true)
```
thomasculino: The `true` parameter shows the progress bar. It's nice when running a binary but since it will now be called in the scheduler, it will surcharge the logs.

src/bin/scheduler.rs
```diff
@@ -29,6 +30,79 @@ use tokio::sync::oneshot;
...
+                "MPC_orbits is current"
+            );
+            continue;
+        }
+        match age {
+            Some(age) => warn!(age_hours = age / 3600.0, "MPC_orbits is stale, refreshing"),
+            None => warn!("MPC_orbits is missing, populating it"),
+        }
+
+        match mpcorb::refresh_orbits(Some(&db), mpcorb::DEFAULT_MPCORB_URL, 10_000, now).await {
```
thomasculino: We need a timeout in case something goes wrong with the download or it will hang forever.

src/utils/mpcorb.rs
mcoughlin: fixed

src/utils/mpcorb.rs

src/bin/scheduler.rs
mcoughlin: fixed

thomasculino (approved):

PR #587, opened 2026-08-26
Improve scheduler stability and logging

antoine-le-calloch:
- Wire `retry_transient` into the alert worker's Valkey calls (`rpoplpush`, `lpush`, `lrem`), so a transient Valkey blip no longer kills the worker thread on the output side
- Requeue alerts left in the alert temp queue at scheduler startup, before any alert worker spawns; nothing ever read that queue, so those alerts were stranded for good
- Log the rescued `alert aux already exists` duplicate key at debug instead of ERROR + WARN, since it is the normal concurrent-insert fallback; real Mongo errors still surface at ERROR through the callers
- Log a catalog entry with a missing or null `z` at trace instead of WARN, it is incomplete catalog data rather than an error
- Fix the test cleanup helper deleting a key literally named `{}_temp` instead of the survey's real temp queue


src/alert/base.rs
```diff
@@ -1327,16 +1333,70 @@ fn report_progress(start: &Instant, stream: &Survey, count: u64, message: &str)
     );
 }
 
+pub fn alert_input_queue_name(survey: &Survey) -> String {
+    format!("{}_alerts_packets_queue", survey)
+}
+
+pub fn alert_temp_queue_name(input_queue_name: &str) -> String {
+    format!("{}_temp", input_queue_name)
+}
```
petebachant: These strike me as better implemented in the alert worker trait as a required method rather than in the base module.

petebachant (approved):
One minor nit, but other than that LGTM. Seems like a nice improvement in robustness.

PR #588, opened 2026-08-26
derive SSO geometry on read for history that predates enrichment

mcoughlin:
This PR derives SSO geometry on read for history that predates enrichment.


src/utils/mpcorb.rs
```diff
@@ -234,6 +237,57 @@ pub fn elements_from_document(doc: &Document) -> Option<OrbitalElements> {
...
+/// on read rather than backfilled across the alert collection. A value already
+/// present is left alone: recomputing would re-derive the same number from the
+/// same elements.
+pub fn fill_geometry(
+    target: &mut Document,
+    designation: &str,
+    jd: f64,
+    elements: &HashMap<String, OrbitalElements>,
+) -> bool {
+    if target.get_f64("helio_dist").is_ok() {
```
thomasculino: We should check for all 3 fields: `helio_dist`, `topo_dist` and `phase_angle`.

src/api/routes/queries/find.rs
```diff
@@ -107,5 +115,125 @@ pub async fn post_find_query(
...
+            .get_document_mut("properties")
+            .and_then(|p| p.get_document_mut("sso"))
+        {
+            fill_geometry(sso, &designation, jd, &elements);
+        }
+    }
+}
+
+/// The designation of a document whose `properties.sso` lacks geometry.
+fn sso_designation_needing_geometry(doc: &Document) -> Option<String> {
```
thomasculino: This is run on already projected doc so if a user requests  {"properties.sso.designation":1, "candidate.jd":1}, this function will return true all the time and populate the response even when not asked.

src/filter/ztf.rs
```diff
@@ -821,6 +941,11 @@ impl FilterWorker for ZtfFilterWorker {
 
                 let now_ts = chrono::Utc::now().timestamp_millis() as f64;
 
+                // Before the annotations are serialized: most of the archive
+                // pre-dates enrichment writing geometry.
+                let mut out_documents = out_documents;
+                self.fill_sso_history_geometry(&mut out_documents).await;
```
thomasculino: `fill_sso_history_geometry` runs once per (programid, filter) inside the nested loop, so overlapping alerts can trigger repeated MPC_orbits queries in the same batch. We should cache fetched elements across the whole run() call to avoid re-querying the same designations.

src/utils/mpcorb.rs
```diff
@@ -234,6 +237,57 @@ pub fn elements_from_document(doc: &Document) -> Option<OrbitalElements> {
...
+        return false;
+    }
+    let Some(elements) = normalize_ztf_ssnamenr(designation).and_then(|k| elements.get(&k)) else {
```
thomasculino: `fill_geometry` re-normalizes the designation even though both call sites already normalize it to build the key set. Could pass the normalized key directly instead of the raw designation to skip the duplicate parse.

src/utils/mpcorb.rs

src/api/routes/queries/find.rs
mcoughlin: fixed

src/filter/ztf.rs
mcoughlin: fixed

src/utils/mpcorb.rs

src/api/routes/queries/find.rs
```diff
@@ -126,26 +127,47 @@ const ZTF_ALERTS: &str = "ZTF_alerts";
 ///
 /// The alternative to doing this on read is a backfill across the whole alert
 /// collection; geometry is a pure function of designation and epoch, so reading
-/// is enough. Callers that did not project `properties.sso` and `candidate.jd`
-/// are untouched.
-async fn fill_sso_geometry(db: &Database, collection_name: &str, docs: &mut [Document]) {
+/// is enough. Only fields the projection asks for are written, and a result
+/// without `properties.sso.designation` or `candidate.jd` is left alone.
+async fn fill_sso_geometry(
```
thomasculino, line 139: Requesting `properties.sso.helio_dist `alone silently returns nothing if `designation`/`jd` aren't also projected. They're needed to compute it but aren't forced into the query. We should inject them into the underlying find projection when any geometry field is requested, then stripping them back out of the response if the caller didn't ask for them.

src/filter/ztf.rs
```diff
@@ -944,7 +961,8 @@ impl FilterWorker for ZtfFilterWorker {
                 // Before the annotations are serialized: most of the archive
                 // pre-dates enrichment writing geometry.
                 let mut out_documents = out_documents;
-                self.fill_sso_history_geometry(&mut out_documents).await;
+                self.fill_sso_history_geometry(&mut orbit_cache, &mut out_documents)
```
thomasculino, line 968: This fills in geometry after run_filter's aggregation has already run, so it only helps for displaying values, not filtering on them. A filter's own `$match` on `helio_dist` still sees null for pre-enrichment points at decision time. Works fine for already-enriched alerts (geometry is already stored before the pipeline runs), just doesn't help the backfill case.

src/api/routes/queries/find.rs
mcoughlin: fixed.

src/filter/ztf.rs
mcoughlin: Yeah I am just going to document this as we aren't going to solve Kepler's equations in Mongo :)

thomasculino (approved):
LGTM

PR #589, opened 2026-08-26
Retry kafka metadata errors

antoine-le-calloch:


src/kafka/base.rs
```diff
@@ -40,6 +40,8 @@ static ALERT_PROCESSED: LazyLock<Counter<u64>> = LazyLock::new(|| {
 
 const MAX_RETRIES_PRODUCER: usize = 6;
 const KAFKA_TIMEOUT_SECS: std::time::Duration = std::time::Duration::from_secs(30);
+const METADATA_ATTEMPTS: usize = 4;
```
mcoughlin, line 43: Worth making configurable?

src/kafka/base.rs
```diff
@@ -171,7 +187,32 @@ pub fn count_messages(
...
+            return Err(error);
+        }
+        warn!(
+            ?topic_name,
+            %error,
+            "transient metadata error, retrying ({}/{})",
+            attempt,
+            METADATA_ATTEMPTS
+        );
+        std::thread::sleep(METADATA_RETRY_BACKOFF);
```
mcoughlin, line 206: Maybe we should do logarithmic back off?

mcoughlin (commented):
A few thoughts.

src/kafka/base.rs
antoine-le-calloch: This only affects the Kafka producer, which we only use for development and testing.
The change is there to avoid the flaky test that shows up sometimes when Kafka needs a moment
to settle after a topic is created. Not something worth polluting the config with I think.

mcoughlin (approved):
LGTM

src/kafka/base.rs
antoine-le-calloch: Same here, only use on the producer and If Kafka isn't coming up at all, I'd rather the tests fail fast than
sit there backing off.

PR #592, opened 2026-08-27
Enrich alerts with per-object detection_history for history-aware filters

mcoughlin:
This PR enriches alerts with per-object detection_history for history-aware filters.

### DIFF
--- a/src/enrichment/decam.rs
+++ b/src/enrichment/decam.rs
@@ -111,6 +111,7 @@ pub struct DecamAlertForEnrichment {
     #[serde(rename = "objectId")]
     pub object_id: String,
     pub candidate: DecamCandidate,
+    // Signed SNR is only needed here: detection history counts detections.
     pub prv_candidates: Vec<DecamPhotometry>,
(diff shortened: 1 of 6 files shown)


src/enrichment/decam.rs
```diff
@@ -83,7 +111,7 @@ pub struct DecamAlertForEnrichment {
     #[serde(rename = "objectId")]
     pub object_id: String,
     pub candidate: DecamCandidate,
-    pub prv_candidates: Vec<PhotometryMag>,
+    pub prv_candidates: Vec<DecamPhotometry>,
     pub fp_hists: Vec<PhotometryMag>,
```
thomasculino, line 116: Do we need to change `PhotometryMag` to `DecamPhotometry` for `fp_hists` as well?

thomasculino (commented):
One comment. The rest looks good

src/enrichment/decam.rs
mcoughlin: Nope since forced photometry isn't the same as detections

thomasculino (approved):
LGTM

PR #593, opened 2026-08-28
report chi2 and dof so a 2-point fit is testable

mcoughlin:
This PR reports chi2 and dof so a 2-point fit is testable.


frenbox (approved):
This is also a straightforward change.

frenbox:
If anything of this is going out in outgoing schema, might need to describe them in Babamul.

mcoughlin:
@frenbox How are we managing docs of that kind for the moment?

frenbox:
Something I am unaware of to be honest. Antoine/ Thomas might be the best people to ask. I have not even touched the front end.

PR #594, opened 2026-08-28
fit the recent light curve

mcoughlin:
This PR adds a "recent" block covering features in the last 30 days (came up in a discussion where folks were trying to make sense of the linear fits over much longer time scales).


frenbox (approved):
This looks okay to me. Good to confine within 30 days

PR #596, opened 2026-08-28
Add WISE as a survey

mcoughlin:
This PR adds WISE as a survey option.


antoine-le-calloch:
We should use watchlist here. I am creating the binary to ingest it.

PR #597, opened 2026-08-31
Improve test workflow

antoine-le-calloch:
Cancel superseded CI runs on pull request branches. Port the concurrency pattern skyportal uses: a new commit cancels the runs still going for the previous one, keyed on the SHA when on main so each main commit keeps its own run and the throughput baseline lookup still finds them. delete-superseded-runs.yaml, also taken from skyportal, deletes those cancelled runs so a red cross only ever means a real failure, going through workflow_run for a writable token on fork PRs and comparing run numbers to keep manual cancels.


mcoughlin (approved):
LGTM

PR #598, opened 2026-08-31
add lsdr10 and ps1_dr2 catalogs

thomasculino:
This pull request updates the crossmatch configuration to include two new catalogs, `PS1_DR2` and `lsdr10`, across multiple survey sections in both `config.yaml` and `config/prod/caltech/overrides.yaml`. These additions specify matching parameters and projection fields for each catalog, enabling their use in crossmatching operations.

**Catalog Integration Updates:**

* Added `PS1_DR2` and `lsdr10` as crossmatch catalogs with specified `radius`, `use_distance`, and `projection` fields to several survey sections in `config.yaml` (`crossmatch:`)
* Added `PS1_DR2` and `lsdr10` with matching parameters and projection fields to relevant survey sections in `config/prod/caltech/overrides.yaml` (`crossmatch:`)

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -283,14 +283,14 @@ crossmatch:
         z_phot_err: 1
         photo_z_type: 1
     - catalog: PS1_DR2
-      radius: 2.0 
+      radius: 2.0
       use_distance: false
       projection:
         _id: 1
         ra: 1
         dec: 1
-    - catalog: lsdr10 
-      radius: 30.0 
+    - catalog: lsdr10
+      radius: 30.0
       use_distance: false
       projection:
         _id: 1
--- a/config.yaml
+++ b/config.yaml
@@ -282,6 +282,20 @@ crossmatch:
         z_phot: 1
         z_phot_err: 1
         photo_z_type: 1
+    - catalog: PS1_DR2
+      radius: 2.0 
+      use_distance: false
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
+    - catalog: lsdr10 
+      radius: 30.0 
+      use_distance: false
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
   lsst:
     - catalog: LSPSC
       radius: 30.0 # 30 arcseconds
@@ -354,6 +368,20 @@ crossmatch:
         z_phot: 1
         z_phot_err: 1
         photo_z_type: 1
+    - catalog: PS1_DR2
+      radius: 2.0
+      use_distance: false
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
+    - catalog: lsdr10
+      radius: 30.0
+      use_distance: false
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
   decam: []
   winter:
     - catalog: Gaia_DR3
@@ -449,3 +477,17 @@ crossmatch:
         z_phot: 1
         z_phot_err: 1
         photo_z_type: 1
+    - catalog: PS1_DR2
+      radius: 2.0
+      use_distance: false
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
+    - catalog: lsdr10
+      radius: 30.0
+      use_distance: false
+      projection:
+        _id: 1
+        ra: 1
+        dec: 1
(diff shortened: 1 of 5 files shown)


mcoughlin (commented):
Claude had some comments. I think the overlap with the other catalog is worth thinking about. Also the storage, but maybe that isn't avoidable.

1. It duplicates a catalog that's already there, with a different photo-z parameterization.

LS_DR10_PHOTOZ (2.94B docs) is already crossmatched — same survey, Legacy Survey DR10. The PR adds lsdr10 (2.83B docs) alongside it. Every alert would carry DR10 photo-z twice under two names:


LS_DR10_PHOTOZ:  z_phot, z_phot_err, photo_z_type
lsdr10:          z_phot_mean, z_phot_median, z_phot_std, z_phot_l95, z_phot_u95
lsdr10 is a strict superset (adds objtype, ebv, z_spec, fluxes, shapes), so the sensible resolution is probably replacing LS_DR10_PHOTOZ rather than running both. As it stands, a filter author has two DR10 redshifts and no guidance on which is authoritative.

2. use_distance: false breaks the convention for photo-z catalogs, and that's what drives the cost.

Every existing 30″ catalog carrying a redshift uses use_distance: true — NED (300″), DESI_DR1 (30″), LS_DR10_PHOTOZ (30″). That applies a physical-separation cut in kpc rather than keeping every angular match. The one 30″ catalog with use_distance: false is LSPSC, a point-source classifier where you genuinely want all neighbours.

lsdr10 carries photo-z, so it patterns with LS_DR10_PHOTOZ — but it's configured like LSPSC. Measured on 25 real alert positions:


LS_DR10_PHOTOZ  30", use_distance: true    →  1 match in the sampled document
lsdr10          30", use_distance: false   →  median 16, max 37


3. The storage cost is large and lands on the aux collection.

ZTF_alerts_aux is 232M objects, 4.65 TB, averaging 19.5 KB/doc. A stored match runs ~32 bytes/field; lsdr10 projects 22 fields plus the two derived distance fields, so ~0.77 KB per match × 16 ≈ 12 KB per object — roughly +2.8 TB, a ~60% increase on aux. That's more than the entire ZTF_alerts collection (2.31 TB).

What isn't a problem: both collections exist and are 2dsphere-indexed, all projected fields are present in the data (including the strm_* value-added columns), and queries run in 1–2 ms. PS1_DR2 at 2″ gives median 0 / max 5 matches — negligible cost, and I'd merge that half as-is.

Suggested asks for Thomas: set use_distance: true on lsdr10 to match the other photo-z catalogs, and say explicitly whether it supersedes LS_DR10_PHOTOZ — if it does, drop the old one in the same PR so there's never a window with two conflicting DR10 redshifts.

One caveat on my numbers: match counts come from 25 recent ZTF alert positions, so they reflect that sky, not an all-sky average — Galactic-plane fields would be worse. The storage figure follows from those counts, so treat it as an order-of-magnitude estimate rather than a precise projection.

thomasculino:
@mcoughlin About the first finding,I think we should drop `LS_DR10_PHOTOZ`. We wanted to add missing fields to it but the new `lsdr10` includes everything. Should we remove the crossmatches with it, that I'm not sure. I think we should,  to avoid confusion. 

The second finding should be fixed for sure.

About storage, do we really need to reprocess the crossmatches with all alerts? The fields used are the ones Robert is using so we cannot cut anything there.

Also, we need to normalize naming across instances, it will make our lives easier.

mcoughlin:
@thomasculino https://github.com/boom-astro/boom-catalogs/pull/24 could you add these columns into your PR please?

mcoughlin (commented):
I know this came up yesterday, but for now, LS_DR10_PHOTOZ and lsdr10 will both be crossmatched. Is this what we want for the moment @thomasculino ?

thomasculino:
> I know this came up yesterday, but for now, LS_DR10_PHOTOZ and lsdr10 will both be crossmatched. Is this what we want for the moment @thomasculino ?

I'm waiting for an answer from the high z team. Based on their answer we can decide if we want to completely get rid of LS_DR10_PHOTOZ or add some of the fields in lsdr10. We can wait for their answer before merging this.

antoine-le-calloch (approved):
LGTM

PR #601, opened 2026-08-31
Read WINTER's fid as 1-indexed and refuse ids that are not a band

mcoughlin:
This PR reads WINTER's fid as 1-indexed and refuse ids that are not a band.


thomasculino (approved):
LGTM!

PR #605, opened 2026-09-01
score moving objects against their own recent photometry during enrichment

mcoughlin:
This PR scores moving objects against their own recent photometry during enrichment.

### DIFF
--- a/src/enrichment/ztf.rs
+++ b/src/enrichment/ztf.rs
@@ -712,9 +712,13 @@ impl EnrichmentWorker for ZtfEnrichmentWorker {
         let mut processed_alerts = Vec::new();
         let mut enriched_alerts: Vec<BabamulZtfAlert> = Vec::new();
 
-        let orbits = self.fetch_orbits(&alerts).await;
-        let sso_history = self.fetch_sso_history(&alerts).await;
-        let baselines = self.fetch_baselines(&alerts).await;
+        // Three independent reads of three collections; awaiting them in turn
+        // pays each round trip separately for no reason.
+        let (orbits, sso_history, baselines) = tokio::join!(
+            self.fetch_orbits(&alerts),
+            self.fetch_sso_history(&alerts),
+            self.fetch_baselines(&alerts),
+        );
 
         let batch_size = alerts.len();
         let mut skipped_empty_lightcurve = 0usize;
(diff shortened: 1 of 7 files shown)


src/utils/phase_curve.rs
```diff
@@ -0,0 +1,388 @@
...
+}
+
+/// Distance-corrected magnitude: what the object would show at 1 au from both
+/// the Sun and the observer, still carrying the phase angle's effect.
+fn reduced_magnitude(p: &Point) -> f64 {
+    p.mag - 5.0 * (p.rh * p.delta).log10()
+}
+
+/// Median of `values`, which must be non-empty. Sorts a copy.
+fn median(values: &[f64]) -> f64 {
```
thomasculino: we have the same function in `outburst.rs`, maybe worth keeping only one of them.

src/enrichment/ztf.rs
```diff
@@ -638,6 +713,8 @@ impl EnrichmentWorker for ZtfEnrichmentWorker {
         let mut enriched_alerts: Vec<BabamulZtfAlert> = Vec::new();
 
         let orbits = self.fetch_orbits(&alerts).await;
```
thomasculino: now that we fetch more things, maybe we should run them concurrently instead of sequentially.

src/utils/lightcurves.rs
```diff
@@ -241,6 +245,148 @@ pub struct ActivityMetrics {
...
+            let mut window: Vec<Point> = history
+                .iter()
+                .filter(|(jd, _)| *jd >= test_jd - days)
+                .map(|(_, p)| *p)
+                .collect();
+            if window.is_empty() {
+                return None;
+            }
+            window.push(test);
+            let (_, median) = outburst_statistic(&window, DEFAULT_G12, scatter).ok()?;
```
thomasculino: We shouldn't always use `DEFAULT_G12`. We should use the actual value and fallback to `DEFAULT_G12` when it's empty.

thomasculino (commented):
Minor comments

src/utils/lightcurves.rs
mcoughlin: fixed

src/enrichment/ztf.rs
mcoughlin: fixed

src/utils/phase_curve.rs
mcoughlin: fixed.

thomasculino (approved):

PR #615, opened 2026-09-02
update BTSbot to v2, which reads channels-first and emits logits

mcoughlin:
This PR updates BTSbot to v2, which reads channels-first and emits logits.


thomasculino (approved):
LGTM

PR #616, opened 2026-09-02
rank stars below real host candidates in distance-ordered crossmatches

mcoughlin:
This PR ranks stars below real host candidates in distance-ordered crossmatches.


thomasculino (approved):

petebachant:
FYI, I'm seeing a config diff on (#623) because this was submitted from a fork, which doesn't run the automatic config updating. I think I should enable config regen on pushes to main as well.

PR #617, opened 2026-09-02
Fix some clippy failed

antoine-le-calloch:
Cargo clippy failed on three deny-by-default lints: a `<= 0` comparison on an unsigned integer, a hand-written ln(10), and a loop that always returns on its first iteration. None changes behavior, and the ln(10) literal was truncated, so the constant is if anything more precise.


mcoughlin (approved):
LGTM

PR #619, opened 2026-09-03
Let a Kafka consumer connect to a broker that requires TLS.

antoine-le-calloch:
Its needed by the futur DZA ZTF Kafka alert cluster.


mcoughlin (approved):
LGTM

docs/kafka-auth.md
```diff
@@ -56,19 +58,23 @@ Admin user is a super user; ACLs not required.
...
-  .set("sasl.password", std::env::var("KAFKA_READONLY_PASSWORD").unwrap());
+fn main() {
+    let mut config = rdkafka::ClientConfig::new();
+    config
+        .set("bootstrap.servers", "broker.example.org:9093")
+        .set("security.protocol", "SASL_PLAINTEXT")
+        .set("sasl.mechanism", "SCRAM-SHA-512")
+        .set("sasl.username", "readonly")
+        .set("sasl.password", std::env::var("KAFKA_READONLY_PASSWORD").unwrap());
+}
```
petebachant: This is for SkyPortal-esque consumers. Why bundle into a function?

antoine-le-calloch: It was to avoid the parsing error of the md file. The rust code in .md file don't accept inline code.

petebachant (approved):

docs/kafka-auth.md
antoine-le-calloch: Fixed !

PR #620, opened 2026-09-03
Stop flagging the MPCORB header lines as dropped records

antoine-le-calloch:
`refresh_into_staging` samples any line that fails `parse_line` but is ≥ 103 characters, assuming a line that long is a record being dropped silently. MPCORB's column header and its dashed separator are both longer than that, so they are sampled on every refresh:

```
06:15:09 WARN rejected record-shaped line: Des'n     H     G   Epoch     M        Peri.      Node       Incl.       e            n           a        Reference #Ob
06:15:09 WARN rejected record-shaped line: --------------------------------------------------------------------------------------------------------
06:15:09 INFO MPC_orbits refreshed orbits=1562078 skipped=45
```

So `rejected_samples` is never empty, contrary to what it documents, and a genuinely dropped record would sit unnoticed between two constant false positives.

Every MPCORB record carries digits; the two header lines carry none. That is the discriminator:

```rust
fn is_record_shaped(line: &str) -> bool {
    line.len() >= 103 && line.contains(|c: char| c.is_ascii_digit())
}
```

### DIFF
--- a/src/utils/mpcorb.rs
+++ b/src/utils/mpcorb.rs
@@ -202,6 +202,11 @@ pub fn parse_line(line: &str) -> Option<MpcorbEntry> {
     })
 }
 
+/// A dropped record still carries digits; MPCORB's column header and rule do not.
+fn is_record_shaped(line: &str) -> bool {
+    line.len() >= 103 && line.contains(|c: char| c.is_ascii_digit())
+}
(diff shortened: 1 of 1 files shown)


mcoughlin (approved):
LGTM

PR #621, opened 2026-09-03
Let config.gpu.enabled alone decide GPU inference

antoine-le-calloch:
## Problem

The model loader read `BOOM_GPU__ENABLED` straight from the environment, bypassing the config layer, and defaulted to `true` when unset:

```rust
let use_gpu = env::var("BOOM_GPU__ENABLED")
    .map(|v| env_truthy(&v))
    .unwrap_or(true);
```

So GPU inference had two sources of truth that could disagree:

- `gpu.enabled: false` could not turn the GPU off. With the variable unset, ZTF took the per-worker CPU path while ORT still registered the CUDA execution provider on device 0.
- `gpu.enabled: true` with the variable set to `false` did the reverse: the scheduler built a `SharedModelPool` and a CUDA stream per device while the sessions were built on CPU.

`deploy.yaml` worked around this with ten lines of comment telling operators to keep the repo variable and the deployment YAML in sync by hand. `docs/deployment.md` repeated it.

## Fix

`device_id` already carries the answer:

```rust
if let Some(dev) = device_id {
```

`Some(id)` only originates from `SharedModelPool::load`, reached through `config.gpu.enabled`, or from `validate_gpu_inference`, gated the same way. `config.gpu.enabled` becomes the single source of truth, and `BOOM_GPU__ENABLED` still overrides it through the normal `BOOM_*` overlay, so "set false to force CPU" is unchanged.

It also makes `SharedModelPool::load`'s doc true: it says an empty `device_ids` loads a CPU model set, which the old env default contradicted. `ZtfEnrichmentWorker::gpu_enabled` now accounts for that case too.

On macOS this changes behaviour only when `gpu.enabled` is `false`, where CoreML was used regardless; `docs/gpu.md` already tells macOS users to set it to `true`.

## Also here

- `docker-compose.yaml`: `BOOM_GPU__DEVICE_IDS: ${...:-0}` becomes `${...:-}`, so a deployment config's `device_ids` can win under Docker — `ignore_empty(true)` exists for that form. `BOOM_GPU__ENABLED: ${...:-false}` is left alone: it also selects the CUDA overlay, so its default is an interlock.
- Dropped the now-false comments in `deploy.yaml`, `docs/deployment.md`, `config.yaml` and the generated prod configs. The generated configs carry the comment change only, no values.
- `env_truthy` and `device_id.unwrap_or(0)` are gone; `use std::env` is gated on Linux, its only remaining use.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -100,15 +100,8 @@ jobs:
       BOOM_API_RATE_LIMIT_AVERAGE: ${{ vars.BOOM_API_RATE_LIMIT_AVERAGE }}
       BOOM_API_RATE_LIMIT_BURST: ${{ vars.BOOM_API_RATE_LIMIT_BURST }}
       BOOM_API_RATE_LIMIT_PERIOD: ${{ vars.BOOM_API_RATE_LIMIT_PERIOD }}
-      # Set explicitly because the model loader (src/enrichment/models/base.rs)
-      # reads BOOM_GPU__ENABLED straight from the environment and DEFAULTS TO
-      # TRUE when unset; config.gpu.enabled does not gate it. Driven by the
-      # BOOM_GPU__ENABLED (true/false) and BOOM_GPU__DEVICE_IDS (e.g. "0,1")
-      # repository variables; no code change needed to flip it. Keep
-      # config/prod/<deployment>/config.yaml's gpu.enabled in sync so the
-      # scheduler's GPU-pool path agrees with the model loader.
-      # BOOM_GPU__ENABLED also drives the "Select GPU compose overlay" step
-      # below, which layers in docker-compose.cuda.yaml (CUDA image + device
+      # Overrides config.gpu.enabled, and drives the "Select GPU compose overlay"
+      # step below, which layers in docker-compose.cuda.yaml (CUDA image + device
       # passthrough); setting this var alone does nothing without that overlay.
       BOOM_GPU__ENABLED: ${{ vars.BOOM_GPU__ENABLED || 'false' }}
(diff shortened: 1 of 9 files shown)


mcoughlin (approved):
Nice.

.github/workflows/deploy.yaml
```diff
@@ -100,15 +100,8 @@ jobs:
...
-      # repository variables; no code change needed to flip it. Keep
-      # config/prod/<deployment>/config.yaml's gpu.enabled in sync so the
-      # scheduler's GPU-pool path agrees with the model loader.
-      # BOOM_GPU__ENABLED also drives the "Select GPU compose overlay" step
-      # below, which layers in docker-compose.cuda.yaml (CUDA image + device
+      # Overrides config.gpu.enabled, and drives the "Select GPU compose overlay"
+      # step below, which layers in docker-compose.cuda.yaml (CUDA image + device
       # passthrough); setting this var alone does nothing without that overlay.
       BOOM_GPU__ENABLED: ${{ vars.BOOM_GPU__ENABLED || 'false' }}
       BOOM_GPU__DEVICE_IDS: ${{ vars.BOOM_GPU__DEVICE_IDS || '0' }}
```
petebachant: Maybe we don't want defaults here so we can control directly in YAML instead of the GitHub environments?

antoine-le-calloch: Fixed for BOOM_GPU__DEVICE_IDS: || '0'.
Kept for BOOM_GPU__ENABLED: the overlay step keys off that var to add docker-compose.cuda.yaml, so unset would mean gpu.enabled: true on a CPU image with no devices.

petebachant: Gotcha. Just confirmed it is `true` in the GH env.

petebachant (approved):

PR #624, opened 2026-09-03
Regenerate deployment configs on pushes to main

petebachant:
**From Claude:**

## Why

`sync-configs` only runs on `pull_request` and `workflow_dispatch`, so a PR is only ever checked against the base it was last pushed against. A config source that lands *after* that check leaves main's generated configs stale, and nothing ever notices.

That is what happened to `config/prod/umn/config.yaml`:

| When | What |
| --- | --- |
| `2026-09-02T23:08Z` | #616's last `sync-configs` run — passed |
| `2026-09-02T23:14Z` | #600 merged, **creating** `config/prod/umn/config.yaml` |
| `2026-09-03T15:58Z` | #616 merged, never re-checked (no new pushes) |

When the check last looked at #616, `config/prod/umn/` did not exist, so there was nothing to regenerate — it passed honestly. The DESI stellar-type ranking then sat on main with UMN never regenerated against it. The gap is widest for fork PRs like #616, which cannot be fixed up in place and are only re-checked when the contributor pushes again.

The drift surfaced only when the next same-repo branch merged main in, at which point the bot regenerated and pushed the fix onto that unrelated PR (#623).

## What this does

Adds `push: branches: [main]`, so whatever a merge actually produced is regenerated against the tree it landed in.

The commit step needed the new event named explicitly: a `push` event has no `github.event.pull_request`, so it matched none of the existing conditions. Checkout and push already fall back to `github.repository` / `github.ref_name`, and the fork-verify step stays gated on `pull_request`.

Note the `concurrency:` block already special-cases main — "main keeps one run, and one status, per commit", with `cancel-in-progress` disabled there. The main trigger looks like it was designed and then dropped; this restores it.

The second commit fixes a related hole found in review: `git diff` only compares tracked paths, so a generated config that has never been committed read as "no changes" and was skipped by both the commit step and the fork-PR guard. That is the shape of a newly added deployment — a committed `overrides.yaml` whose generated `config.yaml` was not committed — i.e. exactly what #600 did. Staging as intent-to-add first makes new files visible to both.

## Verification

- `make check-configs` depends on `configs`, which regenerates *then* validates, so the main run writes files before the commit step. A genuinely invalid config still fails loudly rather than being committed.
- Running `make configs` locally against main reproduced the bot's UMN commit byte-for-byte (same blob `5fd17e10`).
- With an untracked `config/prod/*/config.yaml`, `git diff --quiet` exits 0 before `git add -AN` and 1 after; the clean-tree path still reports up to date.
- Pushes made with `GITHUB_TOKEN` do not trigger workflows, so committing back to main cannot retrigger this job.
- YAML parses; triggers are now `pull_request`, `push`, `workflow_dispatch`.

Once this lands, the stray config commit on #623 becomes redundant.

### DIFF
--- a/.github/workflows/sync-configs.yaml
+++ b/.github/workflows/sync-configs.yaml
@@ -3,12 +3,8 @@ name: Sync configs
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
-  # A PR is only ever checked against the base it was last pushed against, so a
-  # config source that lands after that check leaves main's generated configs
-  # stale with nothing to notice — worse for fork PRs, which can't be fixed up
-  # in place and are only re-checked when the contributor pushes again. Running
-  # on main closes that window: whatever the merge actually produced is
-  # regenerated against the tree it landed in.
+  # Run this on merges to main since it will not run against forks, so we may
+  # miss config updates on PRs from forks.
   push:
     branches: [main]
   workflow_dispatch:
--- a/.github/workflows/sync-configs.yaml
+++ b/.github/workflows/sync-configs.yaml
@@ -16,6 +16,7 @@ concurrency:
 
 permissions:
   contents: write
+  pull-requests: write
 
 jobs:
   sync-configs:
@@ -40,10 +41,8 @@ jobs:
       - name: Create and check prod configs
         run: make check-configs
 
-      # Pushes made with GITHUB_TOKEN don't trigger workflows, so committing
-      # back to main cannot retrigger this job.
-      - name: Commit generated configs
-        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
+      - name: Detect config drift
+        id: drift
         run: |
           # Stage new files as intent-to-add first: `git diff` only sees
           # tracked paths, so a generated config that has never been committed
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-configs.yaml
```diff
@@ -36,8 +44,10 @@ jobs:
       - name: Create and check prod configs
         run: make check-configs
 
+      # Pushes made with GITHUB_TOKEN don't trigger workflows, so committing
+      # back to main cannot retrigger this job.
       - name: Commit generated configs
-        if: github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
+        if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || github.event.pull_request.head.repo.full_name == github.repository
         run: |
           if git diff --quiet -- config/prod; then
```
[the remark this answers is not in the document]
petebachant: **From Claude:** (written by Claude, posted from Pete's account)

Good catch — confirmed and fixed in 47deffc3.

Verified locally: with an untracked `config/prod/_probe/config.yaml` present, `git diff --quiet -- config/prod` exits **0** ("no changes"), so both the commit step and the fork-PR guard silently passed.

Worth noting this isn't hypothetical. `make configs` writes `$dir/config.yaml` for every directory under `config/prod`, so a newly added deployment — a committed `overrides.yaml` whose generated `config.yaml` was never committed — arrives as an untracked file on its first run. That's precisely the shape of the UMN addition (#600) that caused the drift this PR exists to prevent.

Went with `git add -AN -- config/prod` (intent-to-add) rather than `git status --porcelain`, since it fixes detection in both steps *and* makes new files show up in the diff printed by the fork-PR error message, while staging nothing the existing `git add config/prod` wouldn't. After the change `git diff --quiet` exits 1 for the untracked case and the clean-tree path still reports up to date.

.github/workflows/sync-configs.yaml
```diff
@@ -3,6 +3,14 @@ name: Sync configs
 on:
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
+  # A PR is only ever checked against the base it was last pushed against, so a
```
mcoughlin: Maybe based on our discussion today, these comments can be tightened. But otherwise looks good.

mcoughlin (approved):
Might be good to tighten comments, but code is good!

antoine-le-calloch (approved):
LGTM. But I agree with Michael about removing some of the comments. Trying to keep them to one line maximum most of the time would be good.

.github/workflows/sync-configs.yaml
petebachant: Good idea. This might be worth explaining, but Claude is way too verbose. I actually need to check if this would even be able to push to `main` anyway since it's a protected branch.

@antoine-le-calloch could you put the comments policy you proposed into `AGENTS.md`? I have Claude read that so it works across all agents.

PR #626, opened 2026-09-03
Clean up files

antoine-le-calloch:
- `GpuConfig::is_active()` is now the single source for whether the GPU is in use: `enabled` with an empty `device_ids` means CPU for the scheduler, `enrich_reprocess` and the Linux startup validation too, not only for the ZTF enrichment worker.
- `ZtfEnrichmentWorker::models` drops its `Option`: it was always `Some`, so the `else` branch that skipped classification and wrote every alert of the batch without a score was unreachable.
- `get_triplet` and `get_triplet_indexed` share a `stack_triplets` helper and consume the `Vec`s returned by `prepare_triplet` instead of allocating each cutout twice.
- `classify_gpu_batch` carries each model-set position along with the selected index, replacing three `HashMap` builds and three `expect` calls per batch row, and stops cloning one `AllBandsProperties` per alert.
- `CatalogXmatchConfig::from_config` reads its keys through a `required()` closure and `map`/`transpose` instead of nine `match` blocks, and no longer clones every projection value.
- The scheduler loops over its three worker pools instead of repeating the metric call, formats the heartbeat through `pool_state`, and drops two `let mut x = x` rebinds.
- Clippy cleanups on the touched files: manual range comparisons become `contains`, `map_or` becomes `is_none_or`, plus redundant closures, single-character `push_str`, an argument-less `format!`, needless borrows and same-type casts.
- Redundant path prefixes are gone now that `Path`, `Database`, `Document`, `Collection`, `Array` and `Serialize` are imported.
- Multi-line explanatory comments are cut to one line and the paraphrases removed, keeping the ones that record a real breakage: the ORT arena OOM, the `ignore_empty` compose trap and the 12" SSO regression.


src/bin/scheduler.rs
```diff
@@ -197,11 +199,7 @@ struct Cli {
     deployment_env: String,
 }
 
-// `run` deliberately is NOT `#[instrument]`'d. The scheduler runs for the full
-// process lifetime; wrapping it in a single span would make every per-alert
-// span a descendant of the same root, producing a trace that grows unboundedly
-// until Tempo rejects it. The survey is already encoded in the OTel
-// `service.name` resource attribute, so a span field here is redundant.
+// Not `#[instrument]`'d: a process-lifetime root span grows until Tempo balks.
```
mcoughlin: Tempo is the only thing that can balk? ;)

mcoughlin (approved):
Nice cleanup.

Issue #518, opened 2026-06-29
LSST cutouts need a retention policy so we don't run out of storage


petebachant:
Can we offload to S3 or similar after 30 days? I believe AWS has an open data program we could apply for. We could archive all public alerts there, perhaps even have Babamul read old ones directly from S3.

mcoughlin:
https://application.opendata.aws/

so we can see how AWS evaluates applications.

PR #628, opened 2026-09-04
Add async task system for ingesting catalogs, reprocessing alerts, migrating schemas

### LINKED ISSUE #471: Add a way for admins to run database migrations or other similar one-off jobs in an async task queue
MVP could be an API that allows running certain binaries with certain args and logging status in the database, which user did it, etc. This way we can reduce the need to SSH into production.

### LINKED ISSUE #625: Ensure the same catalog has the same name across all BOOM instances
From @thomasculino's comment today.

One way to solve this is to put catalog definitions in the BOOM source code, and use BOOM itself to ingest catalogs so they are named by BOOM, not by the dev/script ingesting it.

In any case, it is BOOM's job to manage its catalogs, and shouldn't allow outsiders to mutate its database without guardrails.

Related to #471

petebachant:
Resolves #471, resolves #625.

Any Python-specific logic from the `boom-catalogs` repo has been migrated in here into a `boompy` Python package, which the Rust code will call via subprocess. Downloading and ingesting are chunked so they can be resumed on error or reboot.

BOOM and Babamul API users share an authorization scheme now. It's always been a little awkward trying to separate the two, and would probably be better thinking of BOOM as a single system, Babamul as the name of an instance, where it has some public streams and API routes along with some private internal ones.

## TODO

- [x] ~~Always ensure catalogs in config exist on startup? I'm not sure about the two-step modify config then kick off ingestion from web app process.~~ Desired catalogs are declared through cross-match config. Ingest kicked off on admin page.
- [ ] Allow for scheduled tasks to suit things like #518.
- [ ] Archive `boom-catalogs` repo? This makes it obsolete, since catalogs are downloaded and ingested here.
- [x] Define admin user emails through env var.
- [ ] Admins should be able to be defined through the admin panel as well? Probably a follow-on PR.
- [ ] Split this into stacked PRs to make it easier to review and merge in cohesive chunks.
- [ ] Need some way to re-ingest catalogs
- [ ] Want to detect missing cross-matching for old alerts when a new catalog is added or the cross-match config has changed, invalidating old cross-matches

## Screenshots

image


petebachant:
From Claude: Closing this in favor of a six-PR stack that splits it into layers that can be reviewed one at a time:

1. #656 Build repo images rather than pulling, and publish the frontend image
2. #657 Add the task system: queue, worker, ledger and admin API
3. #658 Ingest archival catalogs as a task, sourced through boompy

The top of the stack has the same tree this branch ended at, merged with current `main`. It also fixes `tests/api/test_babamul.rs`, which stopped compiling here once `is_admin` was added to `BabamulUser`. The `async-task` branch is left as it is.

Where this PR's TODO list ended up: splitting into stacked PRs is done. Scheduled tasks for #518 are written up in `docs/task-system.md` but not built. Archiving `boom-catalogs` can happen once the catalog ingest layer merges. Defining admins from the admin panel is left for a follow-on.

PR #629, opened 2026-09-05
fixing GPU issues

frenbox:
# Fix GPU memory growth and CUDA stream crashes in ZTF enrichment

Two independent GPU defects found while benchmarking ZTF enrichment on 8× A40 (MSI). One caused unbounded VRAM growth at large batch sizes; the other killed enrichment workers whenever they outnumbered the GPUs. Together they unblock multi-worker, multi-GPU enrichment: 8 workers across 4 GPUs now run clean at ~850–910 alerts/s, a configuration that previously could not complete.

## 1. Unbounded CUDA arena growth

**Changes**

- Pin ORT's CUDA arena strategy to `SameAsRequested` for all sessions.
- Cap the enrichment `RPOP` at the configured `batch_size` instead of a hardcoded `1000`, in both `run_enrichment_worker` and `enrich_reprocess`.

**Reason**

ORT's default `kNextPowerOfTwo` never reuses the freed activation chunk at large batches. Measured on an A40 at batch 900: 34 extensions of 1 GiB each, arena reaching 35 GiB, the freed 872 MiB chunk never reused. Batch 750 reuses it (2 extensions, 3.1 GiB). Capping with `gpu_mem_limit` does not help — a constant ~895 MiB remainder against a 969 MiB request, at caps of 4, 6 and 12 GiB.

`SameAsRequested` extends by the exact request, which fits the next batch because enrichment pads every batch to one shape.

This reverses the note previously at that call site, which said `SameAsRequested` had been "tried and reverted" because it "kills BFC arena reuse and OOMs." Measurement with ORT's own `Total allocated bytes` (log level Verbose plus verbosity 1, since the arena logs through VLOG) shows the opposite: it is `kNextPowerOfTwo` that fails to reuse, and `SameAsRequested` that holds. The old comment's reasoning was right about dynamic shapes, but enrichment is not dynamic — `classify` zero-pads to a fixed shape.

The `RPOP` mismatch meant a configured batch smaller than 1000 was split into a full chunk plus a mostly-padded one, paying a second full-cost GPU pass for the remainder.

## 2. CUDA stream crashes when workers outnumber GPUs

**Changes**

- Add `SharedModels::bind_device()`, called before each GPU inference.

**Reason**

Workers round-robin over one model set per GPU, so surplus workers share a set and its single CUDA stream. Nothing ever called `cudaSetDevice`, which is per-thread state defaulting to device 0, so an ORT run using a device-4 stream could execute on a thread bound to device 0. That surfaces as:

```
Non-zero status code returned while running Transpose node.
Name:'model/conv_conv_1/separable_conv2d/depthwise__198'
Status Message: CUDA error cudaErrorInvalidResourceHandle: invalid resource handle
```

At 4 workers / 2 GPUs: 24 errors, 12 worker failures, and 2 slots dead after exhausting their restart budget, leaving the scheduler at `enrichment=2/4` indefinitely. 

**Need to check if all enrichment workers are surviving in Caltech or not**

The bind is per-inference rather than per-worker-startup because `run_enrichment_worker` is `#[tokio::main]`, a multi-thread runtime whose tasks migrate between OS threads — which is also why the crash was intermittent. Villar-PSO needed no change: `GpuBatchData::new` and `batch_pso` already call `set_device` themselves, consistent with the failure always naming an ONNX node.

The model-set load log now also reports `n_gpus` and `sessions_per_device`, so the worker-to-set ratio is visible in any run.

## Notes for reviewers

- Surveys other than ZTF now take the `RPOP` cap from the config default (750) rather than the old hardcoded 1000. No-op for the committed configs, since none of them set `batch_size`.
- `SameAsRequested` is unconditional rather than configurable: every in-tree caller of `load_model_on_device` pads to a fixed shape. A dynamic-shape model added later would need to revisit it; the call site says so.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -145,9 +145,8 @@ workers:
     enrichment:
       n_workers: 3
       # Alerts per enrichment batch: the queue RPOP cap AND the fixed ONNX
-      # inference shape. Partial batches are zero-padded so ORT sees one shape
-      # and the GPU memory arena stays stable. 750 is stable for both VRAM => 12GB.
-      # Caltech and MSI can handle 1000.
+      # inference shape. One shape alone does not bound the GPU arena; the CUDA
+      # sessions also pin SameAsRequested. Use 750 on cards under 12GB.
       batch_size: 1000
     filter:
       n_workers: 3
--- a/config.yaml
+++ b/config.yaml
@@ -144,10 +144,8 @@ workers:
       n_workers: 3
     enrichment:
       n_workers: 3
-      # Alerts per enrichment batch: the queue RPOP cap AND the fixed ONNX
-      # inference shape. One shape alone does not bound the GPU arena; the CUDA
-      # sessions also pin SameAsRequested. 1000 is sized for ~12GB cards; the
-      # startup check only guarantees 10GiB free, so use 750 on smaller cards.
+      # Alerts per batch: the RPOP cap and the fixed ONNX input shape (CUDA pins SameAsRequested).
+      # 1000 is sized for ~12GB cards; the 10GiB startup check is only a floor — use 750 below that.
       batch_size: 1000
     filter:
       n_workers: 3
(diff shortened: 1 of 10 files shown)


frenbox:
I tested it locally on my computer. Both classification models and villar fit were working fine.

antoine-le-calloch (approved):
BIG LGTM. Very nice work here. Fable and I are very happy.

PR #630, opened 2026-09-06
ingest comet orbits and propagate non-elliptical conics

mcoughlin:
This PR allows to ingest comet orbits and propagate non-elliptical conics.

### DIFF
--- a/src/enrichment/ztf.rs
+++ b/src/enrichment/ztf.rs
@@ -475,6 +475,13 @@ pub struct ZtfSsoAssociation {
     /// Sun-object-observer angle at the alert epoch, degrees. Same provenance.
     #[serde(default)]
     pub phase_angle: Option<f32>,
+    /// Angle from perihelion at the alert epoch, degrees, negative inbound.
+    #[serde(default)]
+    pub true_anomaly: Option<f32>,
+    /// Time of perihelion passage, JD. Stored per detection because a refreshed
+    /// orbit moves it, so one value per object would go stale.
+    #[serde(default)]
+    pub perihelion_time: Option<f64>,
 }
 
 impl ZtfSsoAssociation {
@@ -496,6 +503,8 @@ impl ZtfSsoAssociation {
             helio_dist: None,
             topo_dist: None,
             phase_angle: None,
+            true_anomaly: None,
+            perihelion_time: None,
         }
     }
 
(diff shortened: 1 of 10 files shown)


src/utils/mpcorb.rs
```diff
@@ -393,9 +408,73 @@ async fn refresh_into_staging(
...
+                "h": entry.h,
+                "g": entry.g,
+                "updated_at": now,
+            });
+        }
+    }
+
+    let parsed = documents.len() as u64;
+    if let (Some(c), false) = (staging, documents.is_empty()) {
+        c.insert_many(documents).await?;
```
thomasculino, line 486: One error stops the whole refresh because of the `?`. An `Ok/Err` would avoid dropping the whole batch because of one error.

src/utils/mpcorb.rs
```diff
@@ -555,9 +640,14 @@ pub fn normalize_ztf_ssnamenr(ssnamenr: &str) -> Option<String> {
         return Some(s.to_string());
     }
 
+    // Comets key on IPAC's own form, which is how the ingest stores them.
+    if s.contains('/') || s.ends_with(|c: char| "PCDXI".contains(c)) {
+        return Some(s.to_string());
```
thomasculino: If the designation ends with PCDXI, it never reaches the space insertion logic.

Also, is it not possible for a comet to have another format? Where is the IPAC form you are mentioning?

src/enrichment/ztf.rs
```diff
@@ -496,6 +502,8 @@ impl ZtfSsoAssociation {
             helio_dist: None,
             topo_dist: None,
             phase_angle: None,
+            true_anomaly: None,
+            perihelion_time: None,
```
thomasculino, line 506: these field need to be added to `GEOMETRY_FIELDS` and `fill_geometry` in `mpcorb.rs`

src/utils/mpcorb.rs
```diff
@@ -393,9 +408,73 @@ async fn refresh_into_staging(
...
+            });
+        }
+    }
+
+    let parsed = documents.len() as u64;
+    if let (Some(c), false) = (staging, documents.is_empty()) {
+        c.insert_many(documents).await?;
+    }
+    tracing::info!("parsed {} comet orbits", parsed);
+    Ok(parsed)
```
thomasculino, line 489: `Ok(parsed) `returns even when `parsed == 0`. Since the final rename fully replaces MPC_orbits, this silently wipes all previously-known comet orbits in this specific case.

src/utils/mpcorb.rs
mcoughlin: fixed

src/utils/mpcorb.rs

src/enrichment/ztf.rs
mcoughlin: fixed

src/utils/mpcorb.rs
mcoughlin: fixed

thomasculino (approved):

PR #631, opened 2026-09-06
Improve binaries

antoine-le-calloch:
- Hoist the range-sharding machinery (`shard_field`, `range_shards`, `merge_filters`, `join_tasks`) from `reprocess_crossmatch` into `utils::db`, with new shared helpers `exact_count`, `collection_exists` and `check_shard_coverage`.
- `range_shards` now assigns documents missing the shard field to the first shard, and samples its cut points from the caller's filter subset (keeping `$sample` first to preserve its random-cursor plan).
- A worker that panics now fails the run instead of being silently ignored, which previously reported a partial pass as a success.
- Parallelize `prepare_catalog`'s scan with `--processes`, sharded on an indexed insertion-order field.
- `prepare_catalog` no longer collection-scans the catalog just to count: exact count via an `_id`-hinted `COUNT_SCAN` under `--force`, collection estimate otherwise.
- Fail `prepare_catalog` when the shards did not cover every document (after the 2dsphere index is created, so the catalog stays usable), and only warn when the shortfall can just be concurrent modification.
- Unit-test the shard cut arithmetic, filter merging and task joining.

### DIFF
--- a/src/bin/prepare_catalog.rs
+++ b/src/bin/prepare_catalog.rs
@@ -3,17 +3,21 @@ use boom::{
     conf::{load_dotenv, AppConfig},
     utils::{
         data::{make_progress_bar, spawn_progress_logger},
-        db::create_index,
+        db::{
+            check_shard_coverage, collection_exists, create_index, exact_count, join_tasks,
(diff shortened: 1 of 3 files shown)


mcoughlin (approved):
LGTM!

PR #632, opened 2026-09-06
Add binary that repair photometry ordering

antoine-le-calloch:
- Add `repair_photometry_ordering`, a binary that scans `<survey>_alerts_aux` for timeseries arrays that are not strictly increasing by `jd` (out of order, duplicate, non-finite or non-numeric `jd`) and rewrites them with the same server-side expression ingestion uses (`update_timeseries_op`).
- The scan is sharded with `--processes` and repairs are batched into unordered bulk writes.
- `--dry-run` reports the counts without writing anything, including how many points a real run would delete, split by invalid vs duplicate `jd` and by field.
- Exit nonzero when the target collection does not exist or the shards did not cover the whole collection.
- A test locks `timeseries_fields` to the `AlertAuxForUpdate` structs so a new timeseries field cannot be silently ignored.
- Ship the binary in both Docker images.


mcoughlin (approved):
LGTM

PR #633, opened 2026-09-07
Fix rustfs issue

antoine-le-calloch:
- Pin `rustfs/rustfs` to `1.0.0-rc.5` in the s3 overlays: the `latest` image repushed on Sep 7 breaks every s3 throughput run at container startup.
- Drop `RUSTFS_SERVER_DOMAINS` from both overlays: recent rustfs builds refuse it at startup (`[FATAL] Some subdomains overlap with each other`) and boom forces path-style addressing, so the variable was unused.


thomasculino (approved):

mcoughlin (approved):
LGTM

PR #634, opened 2026-09-07
Add FlareCard component for FLARE classifier output

asasli:
image

Adds the `FlareCard` React component that renders the output of the FLARE photometric classifier for an object: the conformal prediction set with its coverage level, credibility and confidence, the anomaly verdict, and a folded table of class probabilities and p-values.

The card reads the alert's `flare` field and renders nothing when the field is absent, so it is inert until the enrichment side and the wiring in `section-cards.tsx` land in a follow-up PR.


asasli:
@antoine-le-calloch could you take a look when you have a moment? This is the frontend card for the FLARE classifier.

antoine-le-calloch (changes requested):
I see that the style of this component is very AI-generated. Since this component will live within an existing frontend application with its own design and styling, it would be better to align the component's style with the existing app.

asasli:
Screenshot 2026-09-07 at 10 52 45 AM

I reworked the card to reuse the app's own patterns instead of the previous styling. Let me know what you think.

asasli:
Screenshot 2026-09-07 at 11 26 42 AM

This can be the report for normal classes.

antoine-le-calloch (approved):
LGTM

asasli:
@antoine-le-calloch I added some more code here. I added the part that makes the card and the ML Scores tile actually show up on the object page: `section-cards.tsx` renders `FlareCard` below ML Scores, and `ClassificationsV2.tsx` adds a `FLARE` tile with the probability of the predicted class.

Two notes, and I can change these if you prefer:
- Flare does not return per-alert FLARE history yet, so the tile currently has no temporal view. I can modify this later and add it as well.
- Missing fields currently fall back to 0 or an empty set, similarly to the ACAI tiles. I’m not sure whether this is really necessary, though.

PR #635, opened 2026-09-07
Add a guide for adding a model to the enrichment worker

asasli:
Adds `docs/adding-a-model.md`, a short guide on what needs to change in BOOM to run a new classifier on ZTF alerts, and links it from the README.

I think @mcoughlin was thinking about something like this. It is based on my experience with the FLARE integration. Let me know if you have more thoughts and I can work on them.

Who wants to review it? @mcoughlin @antoine-le-calloch or @frenbox?

### DIFF
--- a/docs/adding-a-model.md
+++ b/docs/adding-a-model.md
@@ -1,0 +1,52 @@
+# Adding a model to the enrichment worker
+
+This page lists what needs to change in BOOM to run a new classifier on ZTF alerts. It uses FLARE as the example, which reads the light curve and the crossmatches and writes a structured result. Models that only need the cutouts should implement the `Model` trait in `src/enrichment/models/` instead, like ACAI and BTSbot.
+
+## Where the model runs
+
+The enrichment worker (`src/enrichment/ztf.rs`) reads alerts from the database in batches, computes properties and ML scores, and writes them back with a `$set` update. The alert is fetched with the aggregation pipeline in `create_ztf_alert_pipeline`, which joins the aux document. Add any field you need to the projection there (FLARE added `cross_matches`).
+
+What is available per alert:
+
+- `candidate`: the current detection
+- `prv_candidates`: previous alert detections
+- `fp_hists`: forced photometry
+- `cross_matches`: catalog name to list of matches, each with `distance_arcsec`
+- properties BOOM already computed, e.g. the peak magnitude
+
+Which catalogs exist depends on the deployment. The base `config.yaml` has Gaia DR3, milliquas, NED, DESI DR1 and the Legacy Surveys photo-z. Each site adds its own in `config/prod/<site>/overrides.yaml`. Treat a missing catalog as NaN rather than failing.
+
+## Files to change
+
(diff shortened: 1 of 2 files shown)


docs/adding-a-model.md
```diff
@@ -0,0 +1,52 @@
...
+- `prv_candidates`: previous alert detections
+- `fp_hists`: forced photometry
+- `cross_matches`: catalog name to list of matches, each with `distance_arcsec`
+- properties BOOM already computed, e.g. the peak magnitude
+
+Which catalogs exist depends on the deployment. The base `config.yaml` has Gaia DR3, milliquas, NED, DESI DR1 and the Legacy Surveys photo-z. Each site adds its own in `config/prod/<site>/overrides.yaml`. Treat a missing catalog as NaN rather than failing.
+
+## Files to change
+
+- `Cargo.toml`: the model crate as an optional dependency and a feature that enables it, so the default build does not change.
```
petebachant, line 21: I wonder if we really want to go down the road of adding many different feature options versus simply saying all BOOM instances have a fixed set of features/models, since they're in the repo.

antoine-le-calloch (changes requested):
A few things to fix in the text:

- The sentence about catalogs is wrong on two points: the base ZTF list also has `PS1_DR2`, and sites do not "add" their own. `make configs` uses a yq `*` merge which replaces the list, so each `config/prod/*/overrides.yaml` repeats the full list (and umn even drops `PS1_DR2`).
- `Dockerfile.gpu` already builds with `--features gpu`, and it lists its `--bin` targets, so a new binary must be added there too.
- `create_ztf_alert_pipeline` is shared with the filter worker (`src/filter/ztf.rs:849`), so adding a field to the projection also makes every filter read bigger. Worth saying.
- `flare = "0.2.0"` is already a dependency (`flare::Time` is used everywhere), so a `flare` module and a `flare` feature would be confusing.
- `src/enrichment/mod.rs` declares workers as private `mod ztf;` with a re-export, so use `mod` + `pub use` instead of `pub mod`.
- `cross_matches` lives on the aux document, so the projection line is `"cross_matches": "$aux.cross_matches"`, not `1`.

docs/adding-a-model.md
```diff
@@ -0,0 +1,52 @@
+# Adding a model to the enrichment worker
+
+This page lists what needs to change in BOOM to run a new classifier on ZTF alerts. It uses FLARE as the example, which reads the light curve and the crossmatches and writes a structured result. Models that only need the cutouts should implement the `Model` trait in `src/enrichment/models/` instead, like ACAI and BTSbot.
```
mcoughlin, line 3: Since FLARE hasn't landed yet, I wonder if we posit this as a theoretical addition or how to make this not confusing?

mcoughlin (commented):
I think it looks really good, just tricky as Flare isn't merged yet, so not yet applicable as writtent.

docs/adding-a-model.md
asasli: I agree, but we can also wait a little bit for the flare to be fully integrated (maybe?). Otherwise, we can remove it.

PR #636, opened 2026-09-08
Add WTP (WISE Transient Pipeline) watchlist configuration

antoine-le-calloch:
The test data have already been added to both instances (Caltech and UMN).


mcoughlin (approved):
LGTM

PR #638, opened 2026-09-09
keep one document per comet when MPC publishes a second solution

mcoughlin:
The comet ingest failed with a duplicate `_id` after #637:

```
E11000 duplicate key error collection: boom.MPC_orbits_staging
index: _id_ dup key: { _id: "C/2023RS61" }
```

CometEls.txt carries two rows for `C/2023 RS61 (PANSTARRS)` — the same packed
designation with near-identical elements, one referencing an MPEC and one still
pending:

```
CK23R61S  ... q=7.992085  H=6.8   MPEC 2026-RA
CK23R61S  ... q=7.992060  H=9.1   MPC xxxxx
```

Before #637 these landed under different keys, because each row fused its own
trailing reference into the designation. With that fixed they collide, and
`insert_many` aborts the whole refresh.

Collecting into a map keyed by designation makes the later row win. Against
today's file: 959 records, 958 unique.

### DIFF
--- a/src/utils/mpcorb.rs
+++ b/src/utils/mpcorb.rs
@@ -453,28 +453,34 @@ async fn refresh_comets_into_staging(
     }
 
     let file = std::fs::File::open(tmp.path())?;
-    let mut documents: Vec<Document> = Vec::new();
+    // MPC publishes a second solution for a comet while one is pending, so the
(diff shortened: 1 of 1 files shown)


mcoughlin:
Trivial fix to the comet ingest. Self-merging.

PR #640, opened 2026-09-10
Linking

mcoughlin:
This PR explores linking alerts together (cross-surveys) to identify potential asteroids.

### DIFF
--- a/src/bin/find_tracklets.rs
+++ b/src/bin/find_tracklets.rs
@@ -6,6 +6,7 @@
 //! where anything new would be.
 
 use boom::conf::{load_dotenv, AppConfig};
+use boom::utils::heliolinc::{link_tracklets, main_belt_hypotheses, LinkConfig};
 use boom::utils::linking::{find_tracklets, Detection, Tracklet, TrackletConfig};
 use clap::Parser;
 use futures::StreamExt;
@@ -48,6 +49,73 @@ struct Cli {
     /// Read detections from a JSONL dump instead of the database.
     #[arg(long, value_name = "FILE")]
     input: Option<String>,
+
+    /// Find tracklets per night, then link them across nights.
+    #[arg(long, default_value_t = false)]
+    link: bool,
+
+    /// Position agreement required to cluster propagated states, au.
+    #[arg(long, default_value_t = 0.002)]
+    position_tol: f64,
+
+    /// Velocity agreement required to cluster propagated states, au/day.
+    #[arg(long, default_value_t = 0.0004)]
+    velocity_tol: f64,
+}
+
+/// Tracklets found independently in each night the detections span.
+fn tracklets_per_night(detections: &[Detection], cfg: &TrackletConfig) -> Vec<Tracklet> {
+    let mut by_night: HashMap<i64, Vec<Detection>> = HashMap::new();
+    for d in detections {
+        by_night
+            .entry((d.jd - 0.5).floor() as i64)
+            .or_default()
+            .push(*d);
+    }
+    let mut nights: Vec<_> = by_night.into_iter().collect();
+    nights.sort_by_key(|(n, _)| *n);
+    let mut out = Vec::new();
+    for (night, dets) in nights {
+        let found = find_tracklets(&dets, cfg);
+        info!(
+            "night {}: {} detections -> {} tracklets",
+            night,
+            dets.len(),
+            found.len()
+        );
+        out.extend(found);
+    }
+    out
+}
+
+/// How well tracks reproduce the labels: pure, mixed, and objects recovered.
+fn score_tracks(
+    tracks: &[boom::utils::heliolinc::Track],
+    tracklets: &[Tracklet],
+    labels: &HashMap<i64, String>,
+) -> (usize, usize, usize) {
+    let name_of = |t: &Tracklet| -> Option<&str> {
+        t.ids
+            .iter()
+            .find_map(|id| labels.get(id).map(|s| s.as_str()))
+    };
+    let mut pure = 0;
+    let mut mixed = 0;
+    let mut recovered: std::collections::HashSet<&str> = std::collections::HashSet::new();
+    for track in tracks {
+        let names: std::collections::HashSet<&str> = track
+            .members
+            .iter()
+            .filter_map(|&m| name_of(&tracklets[m]))
+            .collect();
+        if names.len() == 1 {
+            pure += 1;
+            recovered.extend(names);
+        } else if names.len() > 1 {
+            mixed += 1;
+        }
+    }
+    (pure, mixed, recovered.len())
 }
 
 /// One line of a dump: the fields `load` would have projected.
@@ -230,7 +298,11 @@ async fn main() {
         ..TrackletConfig::default()
     };
     let started = std::time::Instant::now();
-    let tracklets = find_tracklets(&detections, &cfg);
+    let tracklets = if args.link {
+        tracklets_per_night(&detections, &cfg)
+    } else {
+        find_tracklets(&detections, &cfg)
+    };
     info!(
         "{} tracklets from {} detections in {:.1}s",
         tracklets.len(),
@@ -258,6 +330,53 @@ async fn main() {
             linked.len(),
             distinct.len()
         );
+    }
+
+    if args.link {
+        let jds: Vec<f64> = tracklets.iter().map(|t| t.jd_ref).collect();
+        let reference_jd = (jds.iter().cloned().fold(f64::MAX, f64::min)
+            + jds.iter().cloned().fold(f64::MIN, f64::max))
+            / 2.0;
+        let link_cfg = LinkConfig {
+            hypotheses: main_belt_hypotheses(),
+            reference_jd,
+            position_tol_au: args.position_tol,
+            velocity_tol_au_per_day: args.velocity_tol,
+            min_nights: 2,
+        };
+        let started = std::time::Instant::now();
+        let tracks = link_tracklets(&tracklets, &link_cfg);
+        info!(
+            "{} tracks from {} tracklets over {} hypotheses in {:.1}s",
+            tracks.len(),
+            tracklets.len(),
+            link_cfg.hypotheses.len(),
+            started.elapsed().as_secs_f64()
+        );
+        if !labels.is_empty() {
+            let (pure, mixed, recovered) = score_tracks(&tracks, &tracklets, &labels);
+            info!(
+                "tracks: {} pure, {} mixed, {} distinct objects recovered",
+                pure, mixed, recovered
+            );
+        }
+        for track in tracks.iter().take(args.show) {
+            let name = track
+                .members
+                .iter()
+                .find_map(|&m| tracklets[m].ids.iter().find_map(|id| labels.get(id)))
+                .map(|s| s.as_str())
+                .unwrap_or("-");
+            info!(
+                "track n={} nights={} r={:.2} au rdot={:+.5} label={}",
+                track.members.len(),
+                track.nights,
+                track.hypothesis.r_au,
+                track.hypothesis.rdot_au_per_day,
+                name
+            );
+        }
+        return;
     }
 
     for t in tracklets.iter().take(args.show) {
--- a/src/bin/find_tracklets.rs
+++ b/src/bin/find_tracklets.rs
@@ -1,0 +1,282 @@
+//! Run intra-night tracklet finding over one night of ZTF alerts.
+//!
+//! Two modes. `--known` links the detections IPAC already matched to a solar
+//! system object and scores the result against those labels, which is how the
+//! thresholds get calibrated. The default links the unassociated detections,
+//! where anything new would be.
+
+use boom::conf::{load_dotenv, AppConfig};
+use boom::utils::linking::{find_tracklets, Detection, Tracklet, TrackletConfig};
+use clap::Parser;
+use futures::StreamExt;
+use mongodb::bson::{doc, Document};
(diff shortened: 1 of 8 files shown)


src/utils/thor.rs
```diff
@@ -0,0 +1,448 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct Projected {
+    pub id: i64,
+    pub jd: f64,
+    pub x: f64,
+    pub y: f64,
+}
+
+/// Gnomonic projection of `(ra, dec)` about `(ra0, dec0)`, degrees.
+fn tangent_plane(ra: f64, dec: f64, ra0: f64, dec0: f64) -> Option<(f64, f64)> {
```
thomasculino: Same `tangent_plane()` as in `linking.rs`.  We should keep only one.

src/utils/identify.rs
```diff
@@ -0,0 +1,232 @@
...
+                    jd: d.jd,
+                });
+            }
+        }
+    }
+    matches
+}
+
+/// Median of a set of separations, arcseconds.
+pub fn median(values: &mut [f64]) -> Option<f64> {
```
thomasculino: Duplicates `outburst::median()` but rounds even-length inputs differently. Use `outburst::median()` instead of keeping a second version.

src/utils/orbit_fit.rs
```diff
@@ -0,0 +1,366 @@
...
+        let mut atb = vec![0.0; 6];
+        for obs in observations {
+            let (r_ra, r_dec) = residual(&state, epoch_jd, obs)?;
+            let mut jac_ra = [0.0; 6];
+            let mut jac_dec = [0.0; 6];
+            for (k, step) in steps.iter().enumerate() {
+                let up = perturb(&state, k, *step);
+                let down = perturb(&state, k, -*step);
+                let (ra_up, dec_up) = residual(&up, epoch_jd, obs)?;
+                let (ra_down, dec_down) = residual(&down, epoch_jd, obs)?;
```
thomasculino: This will abort the whole fit if a perturbed state goes out of bounds (can happen near escape speed) — even if the fit was already converging nicely. The candidate-acceptance check below (lines 188–199) handles this same kind of failure gracefully, by just increasing damping and retrying instead of giving up. Might be worth doing the same here instead of propagating the error with ?.

src/bin/find_tracklets.rs
```diff
@@ -0,0 +1,524 @@
...
+    use std::io::Write;
+    let by_id: HashMap<i64, &Detection> = detections.iter().map(|d| (d.id, d)).collect();
+    let mut file = std::io::BufWriter::new(std::fs::File::create(path)?);
+
+    for (i, track) in tracks.iter().enumerate() {
+        let mut epochs: Vec<serde_json::Value> = Vec::new();
+        for &m in &track.members {
+            for id in &tracklets[m].ids {
+                let Some(d) = by_id.get(id) else { continue };
+                epochs.push(serde_json::json!({
```
thomasculino, line 364: Detections shared between a track's tracklets get written twice here. `ades.rs::track_to_psv` handles the same case with a seen set in this diff — might be worth the same guard here.

src/utils/thor.rs
```diff
@@ -0,0 +1,448 @@
...
+                .collect();
+            let mut grid: std::collections::HashMap<(i64, i64), Vec<usize>> =
+                std::collections::HashMap::new();
+            for (k, (x, y)) in shifted.iter().enumerate() {
+                grid.entry(((x / cell).floor() as i64, (y / cell).floor() as i64))
+                    .or_default()
+                    .push(k);
+            }
+
+            let mut groups: Vec<Vec<usize>> = Vec::new();
```
thomasculino, line 179: `cluster()` only checks each point's immediate neighbors, so a chain like A–B–C where A and C aren't directly close can split into two groups instead of staying one. `heliolinc::connected_group()` handles this with a BFS — might be worth the same approach here.

src/utils/linking.rs
```diff
@@ -0,0 +1,467 @@
...
+                .partial_cmp(&b.rms_arcsec)
+                .unwrap_or(std::cmp::Ordering::Equal),
+        )
+    });
+
+    let mut kept: Vec<Tracklet> = Vec::new();
+    for c in candidates {
+        let covered = kept
+            .iter()
+            .any(|k| c.ids.iter().all(|id| k.ids.contains(id)));
```
thomasculino: This dedup loop uses `Vec::contains` (O(n)) instead of a HashSet, unlike `heliolinc::deduplicate()` for the same kind of check. Not a problem at today's scale, but worth the cheaper lookup if candidate counts grow.

thomasculino (commented):
Is the fact that `thor.rs` and `identify.rs` are both not connected to anything normal? Do we have another PR that connects stuff together or is that our next step? 

You also mention cross-surveys in the original comment, is this PR supposed to support that or lay the groundwork for it?

mcoughlin:
@thomasculino yes thor.rs and identify.rs will be wired in next.

src/utils/orbit_fit.rs
mcoughlin: fixed

src/utils/identify.rs
mcoughlin: fixed

src/utils/thor.rs
mcoughlin: fixed.

src/bin/find_tracklets.rs
mcoughlin: fixed

src/utils/thor.rs
mcoughlin: fixed

src/utils/linking.rs
mcoughlin: fixed

antoine-le-calloch (commented):
- `src/utils/heliolinc.rs:528` — night counting uses `jd_ref.floor()` (noon UTC boundary) while the rest of the code uses `(jd - 0.5).floor()`. A ZTF night running past 12:00 UTC counts as two nights, so single-night clusters pass `min_nights = 2`. Should be `(jd_ref - 0.5).floor()`.
- `src/bin/find_tracklets.rs:423` — the RA patch scale uses each detection's own declination instead of the band centre. At dec 60-62 the same RA lands up to 5 patches apart, splitting one object's detections across patches. Use the cosine of the `dy` band centre.
- `src/bin/find_tracklets.rs:416-426` — patches do not overlap (`patch_deg == cfg.max_offset_deg`) and the RA 0/360 wrap is not handled, so an object straddling a boundary loses half its detections and falls below `min_detections` / `min_nights`. The comment claims the opposite; either add overlap or fix the comment.
- `src/utils/linking.rs:271` — the growth window is `|jd - jd_ref| <= max_span_days`, so a tracklet can span twice `max_span_days`. Enforce the total span in `fit_tracklet`.
- `src/utils/linking.rs:270` — every seed pair rescans all detections, O(pairs x N). On a real night (~1e5 detections) this dominates everything else. Restrict the scan to the already sorted declination band, and skip pairs already covered by a kept tracklet.
- `src/utils/linking.rs:300-308` — the dedup loop flagged earlier in this review is marked fixed but is unchanged: still `Vec::contains` inside a nested scan over `kept`.
- `src/utils/orbit_fit.rs:38` — `predict_radec` only needs the position, but `propagate` also computes the velocity by central difference, so 3x the work in the Jacobian loop (12 calls per observation per iteration). Add a position-only propagation path.
- `src/utils/identify.rs:109` — the whole MPCORB catalogue is propagated sequentially per night although rayon is now a dependency; iterating the `HashMap` also makes the order of `matches` nondeterministic.
- No light-time or aberration correction in the RA/Dec predictions (`src/utils/orbit_fit.rs:38`, `src/utils/identify.rs:69`): roughly 10-20 arcsec at 2 au. That accounts for a good share of the 120 arcsec default match radius and biases the 2 arcsec residual gate. The Horizons validation tests only check distances and phase angle, never a sky position.
- `src/utils/ades.rs:56` — `trkSub` is `candid % 36^7`, so two tracklets in one submission can collide and the MPC would merge two unrelated objects. Use a per-submission counter.

src/bin/find_tracklets.rs
```diff
@@ -0,0 +1,929 @@
...
+        patches.len(),
+        patch_deg,
+        args.thor_distances.len()
+    );
+
+    let started = std::time::Instant::now();
+    let clusters: Vec<(thor::Cluster, boom::utils::heliolinc::State)> = patches
+        .par_iter()
+        .flat_map(|patch| {
+            let ra0 = patch.iter().map(|d| d.ra).sum::<f64>() / patch.len() as f64;
```
thomasculino, line 464: We need a circular mean instead of an arithmetic one here otherwise  a detection at 359.9º and another one at 0.1º get averaged at 180º.

src/bin/find_tracklets.rs
```diff
@@ -0,0 +1,929 @@
...
+
+    let started = std::time::Instant::now();
+    let mut cursor = db
+        .collection::<Document>(ORBITS_COLLECTION)
+        .find(doc! {})
+        .await
+        .expect("failed to read MPC_orbits");
+    let mut orbits: Vec<OrbitEntry> = Vec::new();
+    let mut epochs: Vec<f64> = Vec::new();
+    while let Some(d) = cursor.try_next().await.expect("orbit cursor failed") {
```
thomasculino: The `.expect("orbit cursor failed")` panics the whole run on cursor error.

src/utils/ades.rs
```diff
@@ -0,0 +1,446 @@
...
+/// Render tracklets as an ADES PSV submission.
+///
+/// `detections` supplies the astrometry the tracklets refer to by id; a
+/// tracklet whose detections are all missing is skipped rather than reported
+/// with holes in it.
+pub fn to_psv(
+    tracklets: &[Tracklet],
+    detections: &[Detection],
+    header: &SubmissionHeader,
+    uncertainty: Uncertainty,
```
thomasculino, line 138: We use the same `Uncertainty` for all rows. Should we use `rms_arcsec` associated with a tracklet rather than using the same value for the whole submission?

src/utils/heliolinc.rs
```diff
@@ -0,0 +1,1124 @@
...
+        t.dec_ref + t.dec_rate_deg_per_day * step / 2.0,
+    );
+    [
+        (b[0] - a[0]) / step,
+        (b[1] - a[1]) / step,
+        (b[2] - a[2]) / step,
+    ]
+}
+
+fn dot(a: &[f64; 3], b: &[f64; 3]) -> f64 {
```
thomasculino: We already have this in `sso_geometry` (L300). We should remove duplication.

src/bin/find_tracklets.rs
```diff
@@ -0,0 +1,929 @@
...
+
+    /// Longest a tracklet may span, days. Separate from `--span`, which is how
+    /// much data to read: widening it widens the pair search radius. Pass
+    /// 0.0625 to match heliolinx's 1.5 hour default when comparing against it.
+    #[arg(long, default_value_t = 3.0 / 24.0)]
+    max_tracklet_span: f64,
+
+    /// Distinct nights a track must appear on.
+    #[arg(long, default_value_t = 2)]
+    min_nights: usize,
```
thomasculino, line 73: This overwrites the default config from thor.rs which is 3. What default value do we want to use?

src/utils/orbit_fit.rs
```diff
@@ -0,0 +1,402 @@
...
+const POS_STEP_AU: f64 = 1e-6;
+const VEL_STEP_AU_PER_DAY: f64 = 1e-8;
+
+/// Where a state puts the object on the sky at `jd`, degrees.
+///
+/// Light-time corrected: the object is seen where it was when the light left
+/// it, which at 2 au is about 17 minutes and so tens of arcseconds of motion.
+/// One correction leaves well under a milliarcsecond, since the geocentric
+/// distance barely changes over the light time itself.
+pub fn predict_radec(state: &State, epoch_jd: f64, jd: f64) -> Option<(f64, f64)> {
```
thomasculino: The new correction accounts for the observer standing on Earth surface, not its center. `identify.rs` is the only one to use it, should we use the same correction everywhere?

src/bin/find_tracklets.rs
mcoughlin: fixed

src/bin/find_tracklets.rs
mcoughlin: fixed

src/utils/ades.rs
mcoughlin: fixed

src/utils/heliolinc.rs
mcoughlin: fixed

src/bin/find_tracklets.rs
mcoughlin: now its 2

src/utils/orbit_fit.rs
mcoughlin: fixd

src/bin/find_tracklets.rs
thomasculino: The problem remains, I don't think it was fixed

thomasculino (commented):
The arithmetic mean is still here which can can issues. Other than that it looks good

mcoughlin:
@thomasculino sorry about that, there were a couple of spots. I pushed a fix.

antoine-le-calloch (approved):
LGTM

thomasculino (approved):
LGTM

PR #641, opened 2026-09-11
keep the ZTF ToO exposure flag instead of discarding it

mcoughlin:
This PR keeps the ZTF ToO exposure flag instead of discarding it.


thomasculino (approved):
LGTM!

PR #642, opened 2026-09-11
Throughput gate

mcoughlin:
This PR stops the throughput baseline oscillation (and try to widen it past runner noise).


antoine-le-calloch (approved):
I was waiting for that one to come. Thanks for this! Can you clean up the comments a bit so they explain the code rather than the changes? Other than that its a LGTM.

petebachant (approved):
Thanks for the fix. This has been irritating for a while now. The prod system hasn't even come close to saturating, though we haven't had true LSST volume yet.

PR #645, opened 2026-09-13
Keep user email addresses out of PostHog

antoine-le-calloch:
This need to be merge before https://github.com/boom-astro/boom/pull/623.

Data minimization: PostHog is a third-party processor, and the user id is enough for every funnel we track, so the address has no reason to be there. It stays in Mongo, keyed by that same id, so nothing is lost.

PostHog only gets the user id. No email address is sent: not as a person property, not as an event property, not as a `distinct_id`.

- `posthog.identify` and the 15 signup / login / error events no longer send `{ email }`. Their helpers take no properties now, so it cannot come back by accident.
- If `/profile` fails after a login, we identify nobody instead of falling back to the address as `distinct_id`. The session stays anonymous until the next identify.
- Activation and reset links carry `email`, `token` and `activation_code` in the URL, and PostHog sends `$current_url` with every event, so these three are masked in `main.tsx`.
- Cleanup

### DIFF
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -50,6 +50,10 @@ profiles.
 or as a `distinct_id`: the id keys everything to one person already, and the
 id-to-address join lives in Mongo, where the account does. A login whose
 `/profile` call fails identifies nobody rather than falling back to the address.
+Activation and password-reset links do carry the address in their query string,
+so the client masks `email`, `token` and `activation_code` out of the
+`$current_url` every event reports (`mask_personal_data_properties` in
+[`frontend/src/main.tsx`](../frontend/src/main.tsx)).
 
 ## PostHog events
 
--- a/docs/analytics.md
+++ b/docs/analytics.md
@@ -46,6 +46,11 @@ endpoints) are reported against a fixed `babamul-anonymous` id and carry
 `$process_person_profile: false`, so they're counted without creating person
 profiles.
 
+**Email addresses are never sent**, as a person property, as an event property,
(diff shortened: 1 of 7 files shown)


petebachant (approved):
LGTM.

It would be nice to state reasoning for the changes in the PR body, which I assume is GDPR or similar. We could alternatively show a consent banner. It might be nice to at least know the email domain of the user if our goal is to serve as many external ones as possible. I don't know where that falls in the priority list though.

petebachant (approved):
LGTM.

It would be nice to state reasoning for the changes in the PR body, which I assume is GDPR or similar. We could alternatively show a consent banner. It might be nice to at least know the email domain of the user if our goal is to serve as many external ones as possible. I don't know where that falls in the priority list though.

antoine-le-calloch:
> LGTM.
> 
> It would be nice to state reasoning for the changes in the PR body, which I assume is GDPR or similar. We could alternatively show a consent banner. It might be nice to at least know the email domain of the user if our goal is to serve as many external ones as possible. I don't know where that falls in the priority list though.

@petebachant the email is already in the database and link to the id use in posthog so we don't lose any information

petebachant:
> @petebachant the email is already in the database and link to the id use in posthog so we don't lose any information

That's true, but the data won't exist in the system where analysis happens (PostHog), unless we set up some way to push or pull the data in there. Though I don't know how much user analytics are influencing Babamul, and how high a priority growth is.

PR #646, opened 2026-09-13
recurring transients

mcoughlin:
This PR summarizes detection episodes so filters can select recurrence.


src/enrichment/ztf.rs
```diff
@@ -1252,6 +1256,14 @@ impl ZtfEnrichmentWorker {
...
             candidate.jd,
         );
+        let episode_history = EpisodeHistory::from_points(
+            alert
+                .prv_candidates
+                .iter()
+                .map(|p| (p.jd, p.flux.filter(|f| !f.is_nan()).map(|f| f < 0.0))),
+            candidate.jd,
+            EPISODE_GAP_DAYS,
+        );
```
thomasculino: Can we create a function to iterate only once over `alert.prv_candidates`? The same pattern appears in all surveys as well.

thomasculino (commented):
Just one comment to remove iterations that can be done once instead of twice.

src/enrichment/ztf.rs
mcoughlin: done

thomasculino (approved):
LGTM

PR #647, opened 2026-09-13
Add the FLARE ONNX models

asasli:
Adds the three ONNX files of the FLARE light-curve classifier under `data/models/applecider_flare/`, tracked with Git LFS like the other models (3.3 MB in total).

They are exported from the `applecider-ml/flare` repository (tag v2.0.0) and carry their input column order and calibration in the ONNX metadata, so nothing else needs to ship with them. The enrichment step that loads them follows in a separate PR; until then they are not used.


asasli:
@antoine-le-calloch @mcoughlin or @frenbox, could one of you take a look at this?

frenbox:
These are only onnx models. Would be better to see the implementation.

asasli:
Sure, the implementation will follow in this PR. I opened it with the models first to check the packaging before the code lands: a directory for the model (three ONNX files that belong together, with column order and calibration in the ONNX metadata), tracked with LFS like the other models. Is that layout fine with you, or would you rather have them flat like `btsbot-v2.0.0.onnx`?

asasli:
Notes on `src/enrichment/models/flare_classifier.rs`:

Unlike BTSbot and ACAI, the feature extraction lives in the applecider-ml/flare. The 190 features have to match the training pipeline exactly, so the code is kept in one place and tested there, the same way villar-pso is pulled in as a dependency. The crate is imported as `applecider_flare` because BOOM already depends on an unrelated crate called `flare`.

This file is not wired in yet: it needs the dependency in `Cargo.toml`, a config section, the module declaration and the call in the ZTF enrichment worker. I have all of that ready and tested locally. Let me know if you want it in one go or step by step.

PR #649, opened 2026-09-14
quiet the routine MPC_orbits logs

antoine-le-calloch:
- daily refresh logs at info, warn kept for a missing catalogue
- a batch whose objects have no orbit logs at debug, warn only when MPC_orbits is empty


mcoughlin (approved):
LGTM

PR #651, opened 2026-09-14
match a filter pipeline against a MOC region

mcoughlin:
This PR matches a filter pipeline against a MOC region.


src/utils/moc.rs
```diff
@@ -469,6 +473,102 @@ const MIN_COVERING_DEPTH: u8 = 3;
...
+///
+/// A MOC is a set of ranges in the nested ordering, so with the index stored at
+/// [`HPX_DEPTH`] this tests membership rather than approximating it: no cell is
+/// circumscribed and nothing outside the region is returned.
+///
+/// Requires `coordinates.hpx`, which is written on ingest and backfilled across
+/// the archive. An alert without it cannot match, and that is indistinguishable
+/// from a position outside the region, so the backfill has to be complete
+/// before this is trusted over the whole baseline.
+pub fn moc_hpx_stage(moc: &HpxMoc) -> Result<mongodb::bson::Document, String> {
```
thomasculino, line 527: We should add a limit so the query doesn't blow up.

src/utils/moc.rs
```diff
@@ -469,6 +473,102 @@ const MIN_COVERING_DEPTH: u8 = 3;
...
+///
+/// Each becomes an `$or` branch, so this bounds the query mongo has to plan.
+pub const MOC_MATCH_MAX_CONES: usize = 500;
+
+/// Parse a MOC from its IVOA ASCII serialization, e.g. `"5/1-3 8 9 11/1234"`.
+///
+/// A MOC is a sorted cell list, and ASCII carries it directly. FITS pads to
+/// 2880-byte blocks and then needs base64 on top, so it costs several times the
+/// bytes and makes the sender round-trip through a file to produce it.
+pub fn moc_from_ascii(input: &str) -> Result<HpxMoc, String> {
```
thomasculino, line 499: We should cap the input size like we do for `obj_ids` & `candids`

src/api/routes/filters.rs
```diff
@@ -1022,7 +1028,19 @@ pub async fn post_filter_test(
...
+    if let Some(moc_ascii) = body.moc_ascii {
+        let stage = match moc_from_ascii(&moc_ascii).and_then(|moc| moc_hpx_stage(&moc)) {
+            Ok(stage) => stage,
+            Err(e) => return response::bad_request(&e),
+        };
+        let stage = match serde_json::to_value(&stage) {
+            Ok(v) => v,
+            Err(e) => return response::internal_error(&format!("failed to encode moc stage: {e}")),
+        };
+        pipeline.insert(0, stage);
```
thomasculino: Inserting the stage here puts the stage after the $match + $project prefix. We should merge the match stages.

thomasculino (commented):
We need to run the backfill binary once this is merged to query by `coordinates.hpx`

src/utils/moc.rs
mcoughlin: fixed.

src/utils/moc.rs
mcoughlin: fixed

src/api/routes/filters.rs
mcoughlin: fixed.

thomasculino (approved):
LGTM

PR #655, opened 2026-09-15
apply the MOC region to a filter test count

mcoughlin:
This PR applies the MOC region to a filter test count.


thomasculino (approved):
LGTM

PR #657, opened 2026-09-15
Add the task system: queue, worker, ledger and admin API (2/6)

petebachant:
From Claude: Layer 2 of 6, stacked on #656. This adds the task system itself — the queue, the worker, the ledger and the admin API — with no task types registered yet, so the mechanics can be reviewed without any particular job in the way. The task types arrive in the layers above. Part of #471.

## What this does

- `task_runs` in MongoDB is both the queue and the record. A worker claims a run with an atomic `find_one_and_update` that moves it from `queued` to `running` and takes a lease; a heartbeat renews the lease and carries cancellation back.
- A run whose lease lapses is requeued, but only if its task type declares itself `idempotent`. Anything else, including a type this build does not recognise, is failed with an explanation rather than silently run twice.
- `task_worker` is its own binary and compose service (prod and dev), separate from the API so a deploy does not kill a run that has been going for hours. On SIGTERM it stops the task at its next safe point and hands the run back to the queue.
- `data_mutations` is an append-only ledger: who changed what, with which parameters, under which release. The commit is compiled in from a `BOOM_GIT_SHA` build arg and is recorded as absent, not invented, when unset.
- Parameters may carry connection URIs, so every readback — API responses and the ledger — redacts credentials.
- Admin API: `GET /tasks/types`, `POST /tasks`, `GET /tasks`, `GET /tasks/{id}`, `GET /tasks/{id}/logs`, `POST /tasks/{id}/cancel`, `GET /data/mutations`.
- One admin check accepts both BOOM users and Babamul users. Babamul admins are declared in `babamul.admin_emails` (`BOOM_BABAMUL__ADMIN_EMAILS`, comma-separated) and reconciled onto accounts at API startup in both directions, so removing an email revokes access on the next restart.
- `Trigger::Schedule` and `Actor::system()` are on the run record from the start, so scheduled runs (#518) can be added without migrating history. `docs/task-system.md` records what that still needs, including firing exactly once per tick across several workers.

## Also fixes

`tests/api/test_babamul.rs` stopped compiling once `is_admin` was added to `BabamulUser`: seven struct literals were missing the field. #628 had the same problem. It only shows when the integration test targets are built, which I had not been doing locally.

## Testing

All test targets build. `cargo test --lib`: 373 passed. `cargo test --test test_api`: 67 passed. Compose validates for base, override and CUDA.

## Follow-ups

Defining admins from the admin panel, from #628's TODO list, is left for a later PR.

## Stack

1. #656 Build repo images rather than pulling
2. #657 Add the task system: queue, worker, ledger and admin API  **← this PR**
3. #658 Ingest archival catalogs as a task, sourced through boompy

**Merging:** merge bottom-up. This repo does not delete branches on merge, so after merging a layer, delete its branch (GitHub then retargets the next PR to `main`) or retarget the next PR by hand. Prefer a merge commit or a rebase merge over squash: a squashed layer gets a new commit, and the next PR would show that layer's changes again until it is rebased. `Resolves` links only close their issues once the PR carrying them merges into `main`.


config/prod/caltech/config.yaml
```diff
@@ -60,6 +60,10 @@ api:
     admin_password: "" # Set via BOOM_API__AUTH__ADMIN_PASSWORD environment variable
     admin_email: admin@example.com
 babamul:
+  # Emails of the accounts that may run data-mutating tasks from the admin page.
+  # Reconciled onto every account at API startup, so removing someone here
+  # revokes their access on the next restart.
```
petebachant, line 65: This should probably work differently. Admins should be able to add other admins on the web UI. Removal I'm not so sure. We can probably push this admin management thing off until later and keep this system for now.

src/api/routes/babamul/mod.rs
```diff
@@ -205,6 +213,11 @@ pub struct BabamulUserPublic {
     pub orcid_id: Option<String>,
     /// Full name the user chose to display, if any
     pub name: Option<String>,
+    /// Whether this account may run data-mutating tasks. Read-only here: it is
+    /// reconciled from `babamul.admin_emails` at startup, never set through the
+    /// API. Exposed so the web app knows whether to offer the admin page --
+    /// authorization is still enforced server-side on every admin route.
```
petebachant, line 219: This is explained in the other model. We can probably make shits shorter, or remove comment. Admin is an easily understandable concept. Someday Admins will be able to do user management too.

src/api/admin.rs
```diff
@@ -0,0 +1,256 @@
+//! Authorization for the admin surface.
+//!
+//! BOOM has two login realms -- the original API's `users` and Babamul's
+//! `babamul_users` -- and the admin page is reached from the web app, which
+//! authenticates as the latter. Rather than merge the logins, both realms carry
+//! an `is_admin` flag and both are accepted here, so there is one authorization
+//! check and one shape of actor recorded on a run.
```
petebachant, line 7: A cleaner solution in the future would probably be one users collection with more specific permissions, e.g., internal, external, admin.

src/api/admin.rs
```diff
@@ -0,0 +1,256 @@
...
+use crate::api::routes::{babamul::BabamulUser, users::User};
+use crate::tasks::Actor;
+use actix_web::{web, HttpResponse};
+
+/// An authenticated admin, whichever realm they came from.
+#[derive(Debug)]
+pub struct AdminActor {
+    /// `boom` or `babamul`. Recorded so a run can be traced back to an account
+    /// in the right collection -- the two id spaces are unrelated.
+    pub realm: &'static str,
```
petebachant, line 19: Realm is a funny word choice for this, but it works.

docs/task-system.md
```diff
@@ -0,0 +1,266 @@
+# Task system
+
+Operating a BOOM system typically involves adding new catalogs,
+changing the schema of alerts and object already in the database,
+and reprocessing alerts already saved in the database, e.g., when a new
+catalog or enrichment step, e.g., ML model classifier, is added.
```
petebachant:
```suggestion
catalog or enrichment step like an ML model classifier is added.
```

docs/task-system.md
```diff
@@ -0,0 +1,266 @@
...
+It is important for us to be able to track what mutations were done to the
+data when.
```
petebachant:
```suggestion
data, at what time, and with what version of the code.
```

docs/task-system.md
```diff
@@ -0,0 +1,266 @@
...
+  shell access to the production host.
+- **It has to be cancellable** — cleanly, at a point the task chooses, not by
+  killing a process partway through a batch.
+- **It has to leave a record.** Who ran what, with which parameters, against
+  which release. BOOM's scientific artifacts are a function of the current state
+  of the data, and that state is the raw stream plus a sequence of mutations. If
+  those mutations are only in someone's shell history, the artifacts derived
+  from them can't be reasoned about or reproduced.
+
+Access control is the fourth reason and the least interesting one: mutating
```
petebachant: Beyond this, the BOOM application should own its database and not allow access from others--humans, scripts. apps, etc. The purpose of an API is to define that kinds of access and actions are acceptable from whom, i.e., authentication and authorization. We don't want to use MongoDB for auth.

docs/task-system.md
petebachant: We may want to restructure the docs for the reader persona. Users of BOOM are typically SkyPortal integrators or public Babamul users. Then there are BOOM devs and BOOM admins. The task system, for example, is irrelevant to Babamul users and maybe to SkyPortal integrators. SkyPortal gets a single user account, so I suppose we could perhaps give it admin access, but maybe a better approach is for BOOM to have a "log in with SkyPortal (or Fritz?)" option so there is some relationship between user accounts? Otherwise, an admin task could only be attributed to SkyPortal, not the individual. We don't want to couple the systems too tightly, however.

docs/task-system.md
```diff
@@ -0,0 +1,266 @@
+# Task system
```
petebachant: A fundamental principle followed here is that BOOM is responsible for its own data. Others can tell it a desired state, but should not reach in and mutate it on their own.

src/bin/task_worker.rs
```diff
@@ -0,0 +1,287 @@
...
+//!
+//! Claims one run at a time from `task_runs`, holds a lease on it, renews that
+//! lease with a heartbeat, and streams progress and logs back to Mongo while
+//! the task runs. See [`docs/task-system.md`](../../docs/task-system.md).
+//!
+//! Deliberately a separate service from the API: these jobs run for hours and
+//! are memory-hungry, and an API restart or deploy must not kill one. When this
+//! process does go away mid-run, the run's lease lapses and the next worker to
+//! start picks it back up -- which is why every task body is written to be
+//! resumable.
```
petebachant: Is it true that they all are, or just most?

src/api/routes/babamul/mod.rs
```diff
@@ -188,6 +188,14 @@ pub struct BabamulUser {
     /// `username` it is free text, optional, and not unique.
     #[serde(default)]
     pub name: Option<String>,
+    /// Whether this account may run data-mutating tasks from the admin page.
```
petebachant: The scope is broader. Is this an admin user that can perform actions that require elevated privileges, e.g...

src/api/routes/babamul/oauth.rs
```diff
@@ -634,6 +634,9 @@ fn new_social_user(
         // Seeded from the provider so the profile isn't blank on day one; the
         // user can change or clear it via PATCH /babamul/profile.
         name,
+        // Granted only by reconciling against babamul.admin_emails at
+        // startup, never at sign-up.
```
petebachant: Probably don't need this comment. Of course we don't let people sign up as admins.

PR #658, opened 2026-09-15
Ingest archival catalogs as a task, sourced through boompy (3/6)

### LINKED ISSUE #625: Ensure the same catalog has the same name across all BOOM instances

petebachant:
From Claude: Layer 3 of 6, stacked on #657. Catalog definitions now live in the BOOM codebase and BOOM ingests them itself, so a catalog has the same name on every deployment. This replaces the `boom-catalogs` repo. Resolves #625.

## What this does

- `CATALOGS` in `src/catalogs/mod.rs` defines 11 catalogs: slug, collection name, reader, and where the data comes from.
- `catalog_ingest` is a task. Each chunk is downloaded, ingested, deleted, then recorded as done, so peak disk is one chunk rather than one catalog and an interrupted run resumes. `drop_existing` starts over; `max_chunks` smoke-tests a catalog end to end.
- `boompy` is an internal Python package, not a client library, run with `uv run`. It lists and fetches chunks, because the archives are best reached from Python (LSDB, astroquery, astropy). Everything after a file lands on disk is Rust, reading parquet through Arrow or delimited text. It has its own workflow and tests.
- A crossmatch entry naming a catalog this release has no definition for now fails config load. `LSPSC` (the Legacy Survey Point Source Catalog of [Liu et al. 2025](https://arxiv.org/abs/2505.17174)), `TNS` and `LSDR10` are recognised as collections built outside BOOM.
- A configured crossmatch catalog that is missing or empty does not error in MongoDB; every alert just gets zero matches, which looks exactly like a genuine non-match. Alert workers now warn about each such catalog at startup.
- `GET /catalogs/status` reports every catalog the release defines, whether crossmatch config uses it, and whether it is present, partial or missing. The intended order is: add the definition, ingest it from the admin page, then add it to crossmatch config.
- `.gitignore` excludes bulk data by extension, with explicit negations for the fixtures the repo does track.

## Worth a look

- `LSDR10` (added in #598) is not the `ls-dr10-photoz` dataset this ingests. Its projection asks for `z_phot_mean`, fluxes and shape parameters that the record type does not have, so pointing the definition at it would build a collection whose distance crossmatch silently matches nothing. It is declared as externally built instead.
- `PS1_DR2` is now the Pan-STARRS collection, with `PS1_DR1` kept as an alias. This ingest does not carry the PS1-STRM `strm_*` columns that crossmatch config projects.
- Re-running an ingest fills in what is missing but never refreshes: a document whose `_id` already exists is kept and the incoming one discarded. Refreshing needs `drop_existing`, which empties the collection for the duration of the run. `docs/catalogs.md` describes the safer route of ingesting a new release under a versioned collection name.
- Crossmatches against an existing catalog are not invalidated when its config or data changes. Adding a new catalog is handled selectively (`reprocess_crossmatch` with `skip_existing`); the rest is written up in `docs/catalogs.md` and not built.

## Testing

All test targets build. `cargo test --lib`: 422 passed, plus `the_shipped_prod_configs_pass_validation`, which needs the config secrets set. `boompy` pytest: 37 passed. `make check-configs` passes and compose validates. The boompy, config and compose checks ran before this stack was rebased onto #623, which touched none of those files.

## After merging

The `boom-catalogs` repo can be archived.

## Stack

1. #656 Build repo images rather than pulling
2. #657 Add the task system: queue, worker, ledger and admin API
3. #658 Ingest archival catalogs as a task, sourced through boompy  **← this PR**

**Merging:** merge bottom-up. This repo does not delete branches on merge, so after merging a layer, delete its branch (GitHub then retargets the next PR to `main`) or retarget the next PR by hand. Prefer a merge commit or a rebase merge over squash: a squashed layer gets a new commit, and the next PR would show that layer's changes again until it is rebased. `Resolves` links only close their issues once the PR carrying them merges into `main`.


petebachant:
@frenbox this is basically Claude turning your Python scripts into a Python package/CLI for BOOM to call from inside the Rust task worker.

petebachant:
From Claude: Pete found the reference for LSPSC, and it turned out to be in the repo already — `frontend/src/components/kafka/KafkaTopicRules.tsx` cites it on the Kafka topic rules page. It is the Legacy Survey Point Source Catalog of Chang Liu et al. 2025 ([arXiv:2505.17174](https://arxiv.org/abs/2505.17174), PASP): a morphological resolved/unresolved classifier that scores ~3×10⁹ Legacy Survey sources, now running in LS4's real-time pipeline. That is what the crossmatch projection reads as `score` and `mag_white`, and what `enrichment/lsst.rs` uses to call an LSST object stellar or hosted.

The `WITHOUT_DEFINITIONS` entry and `docs/catalogs.md` now say that. They previously said its data source "is not recorded", which was wrong — it was recorded, just nowhere near the catalog code. It still has no ingest definition, because the paper does not say where the catalog itself is published, so there is nothing for BOOM to fetch.

While there I also fixed that doc list: it said two crossmatch targets have no definition, when `LSDR10` had made it three.

petebachant:
From Claude: Correction to my comment above. I had read only the arXiv abstract, which says nothing about a data release; the full text does: "The final catalog containing our star–galaxy scores of the entire LS DR10 is available at https://ls-xgboost.lbl.gov/", with the model code at https://github.com/slowdivePTG/LS-PSC. Pete was right that we had found the dataset.

It is published as a **query service** rather than an archival download: cone search at `/getsources`, radius up to 300″, with an optional `mag_limit` on `white_mag` — which is the `mag_white` our crossmatch projection already reads, so that corroborates the identification.

So `LSPSC` stays in `WITHOUT_DEFINITIONS`, but now for the right reason. It is not that the source is unknown; it is that there is no bulk file set to list and fetch, and BOOM's ingest is built on chunked download. Querying ~3×10⁹ rows through a cone-search endpoint is not an ingest path. If a bulk export appears, this becomes an ordinary `CatalogDef`. The entry and `docs/catalogs.md` now say exactly that, including the endpoint and the code link.

frenbox:
LSDR10 Photo z catalog sits in a unique position where most of the data is downloaded in two parts (tractor and photo-z) and which are minifed and then combined into one file to be ingested. We have now stepped out of it and are using LSDB for LSDR10 which has photo z mean and median values. May be we remove it from both boom-catalogs and here?

frenbox:
There seems to be a serial chunk download and ingestion. For a large catalog (like PANSTARRS) do you think parallel download and ingestion (spawning limited tokio threads) would be a good idea?

frenbox:
I have found some catalogs where ra, dec value exceeded the usual [0,360] , [-90, 90] range. This fails the geojson conversion and ultimately terminates the process. Might need to work around that in both this and boom-catalogs.

config.yaml
```diff
@@ -196,6 +196,20 @@ workers:
...
+# Not needed to make a catalog ingestable: the admin page lists every catalog
+# this release defines, whether or not anything names it. The catalogs under
+# `crossmatch` below are the ones the alert pipeline queries, and those are what
+# the admin page flags when missing. A slug here that the release has no
+# definition for is reported on the admin page rather than ignored.
+#
+# Declaring one does not ingest it. Ingesting is an explicit task started from
+# the admin page: auto-converging on deploy would let a typo rebuild a catalog,
+# and an ingest is hours to days of work.
+catalogs: [] # BOOM_CATALOGS, comma-separated
```
petebachant: Do we actually need this, or is the set in cross-matches enough to know which catalogs are available? Or maybe this is not defined in the config, but instead a property of the instance that is mutated after the fact, i.e., by catalog ingests from the admin page. There is a fixed set of catalogs BOOM knows about living in the codebase.

docker-compose.override.yaml
```diff
@@ -102,6 +102,11 @@ services:
       KAFKA_SOCKET_CONNECTION_SETUP_TIMEOUT_MS: 30000
       KAFKA_SOCKET_CONNECTION_SETUP_TIMEOUT_MAX_MS: 60000
   # Runs queued tasks in dev, the same way the prod task-worker service does.
+  #
+  # `uv sync` before cargo-watch so boompy's environment exists before the first
+  # ingest asks for it; the venv lives in its own volume so a container restart
+  # does not re-resolve it. `--frozen` makes a lockfile that drifted fail here
+  # rather than silently resolving to versions nothing was tested against.
```
petebachant: Comment is too verbose

docker-compose.override.yaml
```diff
@@ -116,29 +121,38 @@ services:
...
     working_dir: /app
+    environment:
+      BOOM_CATALOG_DATA_PATH: /app/data/catalogs
+      BOOM_BOOMPY_PATH: /app/boompy
+      UV_PROJECT_ENVIRONMENT: /app/boompy/.venv
+    entrypoint:
+      # `bash -c`, not `-lc`: a login shell sources /etc/profile and replaces
+      # the PATH the rust image sets, which drops /usr/local/cargo/bin and makes
+      # cargo vanish. A shell is needed at all only to run `uv sync` before
+      # handing off to cargo-watch.
```
petebachant: Slim this down

docker-compose.yaml
```diff
@@ -3,6 +3,10 @@ name: boom
 volumes:
   mongodb:
   valkey:
+  # Staging area for catalog chunks. A named volume rather than the container's
+  # writable layer because a chunk can be a few hundred megabytes, and because
+  # a half-downloaded chunk should survive the task worker being replaced.
+  catalog_data:
```
petebachant:
```suggestion
  # Staging area for catalog chunks
  catalog_data:
```

Dockerfile
```diff
@@ -16,6 +21,9 @@ RUN apt-get update && \
     ln -s /opt/kafka_${SCALA_VERSION}-${KAFKA_VERSION} /opt/kafka && \
     rm -f /tmp/kafka.tgz
 
+RUN curl -LsSf https://astral.sh/uv/${UV_VERSION}/install.sh | \
+    env UV_INSTALL_DIR=/usr/local/bin UV_UNMANAGED_INSTALL=1 sh
```
petebachant: Isn't there a COPY command to install uv directly from a Docker image layer?

src/catalogs/arrow.rs
```diff
@@ -0,0 +1,365 @@
+//! Columnar catalogs, read a record batch at a time.
```
petebachant: This is a poor description of the module, which is named `arrow`. What does it contain?

petebachant: For example, "Functionality for working with catalog data in Parquet format, using Arrow."

PR #663, opened 2026-09-15
Improve config

antoine-le-calloch:
Refactor crossmatch configuration to use maps instead of lists for better config and catalogs management.


mcoughlin (approved):
LGTM!

src/conf.rs
```diff
@@ -308,16 +308,20 @@ impl CatalogXmatchConfig {
...
 
     #[instrument(skip_all, err)]
-    fn from_config(config_value: Value) -> Result<CatalogXmatchConfig, BoomConfigError> {
-        let hashmap_xmatch = config_value.into_table()?;
+    fn from_config(
+        catalog: &str,
+        config_value: Value,
+    ) -> Result<CatalogXmatchConfig, BoomConfigError> {
+        let mut hashmap_xmatch = config_value.into_table()?;
+        // An explicit null unsets a field inherited from the base config.
```
thomasculino: Unsetting a field by nulling it only works at the top level. Nulling a field in a projection of a catalog would not work for example. Not sure if that's intended or not.

thomasculino (commented):
One comment to address, otherwise looks good.

thomasculino (approved):
LGTM! We just need to keep in mind that we can end up with incompatible keys (e.g. `use_distance` and `max_results`) if we don't pay attention, but that's something we need to be careful of.

PR #666, opened 2026-09-16
Install libcfitsio in the GPU runtime image

antoine-le-calloch:
`Dockerfile.gpu` installs `libcfitsio-dev` in the build stage but nothing in the runtime stage, so `scheduler` fails at startup on a missing `libcfitsio.so.10`. The CPU `Dockerfile` already installs `libcfitsio10t64` in its runtime stage.


thomasculino (approved):
LGTM

mcoughlin (approved):
LGTM

PR #667, opened 2026-09-17
2 days for WINTER

mcoughlin:
This PR extends the ingest for WINTER (which publishes everything after the night is over) to 2 days.


thomasculino (approved):
LGTM

PR #668, opened 2026-09-17
Fixing villar un-normalized quantity.

frenbox:
There was a bug in villar-pso package where `beta` parameter was not properly unnormalized. 
During this merge its good to remove all the villar fit parameters from the database.


antoine-le-calloch (approved):
LGTM

PR #669, opened 2026-09-18
Add a dashboard refresh button

antoine-le-calloch:
- `POST /babamul/stats/refresh` drops the cached nightly counts for the given range, plus the collection and Kafka topic caches, so the next read recomputes them.
- Range capped at 6 months, one refresh every 5 minutes globally (429 with `Retry-After`).
- The dashboard gets a Refresh button: it calls the endpoint for the displayed range and refetches the charts, collections and topics.

Refresh button on the dashboard


mcoughlin (approved):
LGTM

PR #670, opened 2026-09-18
Post throughput test results in PR body instead of as new comments

petebachant:
The notifications are kind of annoying and obscure genuine PR discussion.


petebachant:
Or maybe we should just remove the commenting altogether? If it fails, devs can simply look in the Actions logs to see.

PR #671, opened 2026-09-19
log the self-healing timeseries fallback as a warning

mcoughlin:
This PR logs the self-healing timeseries fallback as a warning.


antoine-le-calloch (approved):
LGTM

PR #672, opened 2026-09-19
first detection time from forced photometry

mcoughlin:
This PR adds first detection time from forced photometry (useful for the GRB/GW filters).


src/bin/backfill_detection_span.rs
```diff
@@ -0,0 +1,360 @@
...
+    match doc.get(key) {
+        Some(mongodb::bson::Bson::Double(v)) => Some(*v),
+        Some(mongodb::bson::Bson::Int32(v)) => Some(*v as f64),
+        Some(mongodb::bson::Bson::Int64(v)) => Some(*v as f64),
+        _ => None,
+    }
+}
+
+/// Read one object's arrays, keeping only what the summary needs.
+fn history_from_aux(aux: &Document) -> ObjectHistory {
```
thomasculino: We use `psfFlux` and `snr_psf` in this function but these are ZTF/LSST only fields. Do we not want to include WINTER in this?

